PR #26, opened 2016-06-03
Fix travis badge

xerebz:
Was pointing to the master branch instead of develop


xerebz:
Better solution to this is to fix master's builds :dancer:

PR #27, opened 2016-06-03
Update from District Data Labs

ansate:


bbengfort:
@ansate so we have  develop/release cycle where we add development code to the develop branch and only merge to master when we're ready to release a version of the library. This ensures that master remains stable. Please check out develop to get the latest code. 

You can also read about the branching model here:

- http://nvie.com/posts/a-successful-git-branching-model/
- https://barro.github.io/2016/02/a-succesful-git-branching-model-considered-harmful/

@pvomelveny just mentioned that you told him to tell us to ignore this; so I guess I'll close the PR then!

Issue #141, opened 2017-03-04
ClassificationReport fails with multi-class classifiers


NealHumphrey:
When the ClassificationReport is given a multi-class classifier, the `draw()` method produces an `IndexError` on row 160:

```
C:\Users\humph\Documents\Github\yellowbrick\yellowbrick\classifier.py in draw(self, y, y_pred)
    158         for column in range(len(self.matrix)+1):
    159             for row in range(len(self.classes_)):
--> 160                 self.ax.text(column,row,self.matrix[row][column],va='center',ha='center')
    161 
    162         fig = plt.imshow(self.matrix, interpolation='nearest', cmap=self.cmap, vmin=0, vmax=1)

IndexError: list index out of range
```

It appears that at fault is row 158, which uses `range(len(self.matrix)+1)` instead of `3` (3 is the number of scores for columns of self.matrix, i.e. precision, recall, f1). Changing this to `range(len(self.matrix[0]))` bypasses the IndexError, but still produces an incorrect report - i.e. one that has too many columns, though the data labels and shading are only included in the first 3 columns. The column logic must also be happening elsewhere (maybe in `.imshow`?). 

See Neal's Jupyter Notebook example (pull request coming shortly).

bbengfort:
I experienced a similar issue; here are my notes (thanks for opening the bug report!) 

Attempting to get a classification report for a multi-class classifier:

```
logit_balance = ClassificationReport(logit, classes=set(labels_test))
logit_balance.score(docs_test, labels_test)
logit_balance.poof()
``` 

In the same way as before led to the following exception:

```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-27-7b911cc79c37> in <module>()
      1 logit_balance = ClassificationReport(logit, classes=set(labels_test))
----> 2 logit_balance.score(docs_test, labels_test)
      3 logit_balance.poof()

/Users/benjamin/Repos/tmp/yellowbrick/yellowbrick/classifier.py in score(self, X, y, **kwargs)
    133         self.scores = map(lambda s: dict(zip(self.classes_, s)), self.scores[0:3])
    134         self.scores = dict(zip(keys, self.scores))
--> 135         return self.draw(y, y_pred)
    136 
    137     def draw(self, y, y_pred):

/Users/benjamin/Repos/tmp/yellowbrick/yellowbrick/classifier.py in draw(self, y, y_pred)
    158         for column in range(len(self.matrix)+1):
[...]

Issue #139, opened 2017-02-28
Confusion Matrix for multi-class classifiers


NealHumphrey:
Create a ScoreVisualizer that creates a heatmap color coded confusion matrix showing true label vs. predicted label for multi-class classifiers. Goal of heatmap is to better identify frequently-confused categories. 

- Raw number version (heatmap body shows counts)
- percent-of-true version (i.e. percent of row or percent of column, depending on orientation of predicted vs. true)
- could implement smart color coding, i.e. more than just a single-color heatmap where 100% accurate and other relevant threshold levels get different color treatment. 
- could implement precision / recall scores at the ends of columns/rows, similar to a joint plot. Maybe use a small pie chart  of the score (since heatmap is already used for the number of items in each block).

NealHumphrey:
@rebeccabilbro another option for our Visualizer this sprint.

rebeccabilbro:
@NealHumphrey this sounds a bit like the one we have [here](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier.py#L58). Would this be a subclass of `ClassificationReport`?

Re: for our Visualizer for this sprint, I'd like us to try to stick to user-driven design. We want to look for blog posts, articles, etc where machine learning practitioners are using visualizations to guide their model selection/evaluation process. Then we can think about how to implement those visualizations in the Yellowbrick API. The advantage of doing it this way is that we can address a specific use case, which is helpful for scoping purposes. Whatever preliminary implementation we build can be generalized in later sprints. 

Let me know if you find any interesting posts or articles that you think might work!

bbengfort:
Just wanted to add a couple of things to this issue, but don't want to get in the way of your conversation regarding this Sprint (so feel free to ignore the following):

- [sklearn metrics confusion matrix heatmap](http://scikit-learn.org/stable/auto_examples/model_selection/plot_confusion_matrix.html#sphx-glr-auto-examples-model-selection-plot-confusion-matrix-py)
- [multi-class confusion matrix for text mining](http://text-analytics101.rxnlp.com/2014/10/computing-precision-and-recall-for.html)

Question: how do we add k-fold cross-validation or something like `cross_val_predict` to the mix?

rebeccabilbro:
A few related posts I found:

 - http://notmatthancock.github.io/2015/10/28/confusion-matrix.html
[...]

Issue #121, opened 2017-02-23
User Study: Neal and Rebecca


bbengfort:
- [x] @rebeccabilbro user study
- [ ] @NealHumphrey user study

http://www.scikit-yb.org/en/latest/evaluation.html

NealHumphrey:
Just testing if I have permission to write on a card!

rebeccabilbro:
Added my user study [here](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/examples/rebeccabilbro/testing.ipynb)

NealHumphrey:
@rebeccabilbro fyi I completed most of mine on the plane on Friday, but still need to do a bit more wrap up. Might have some time this week while I'm out here, but will likely be on the plane on the way back on Fri. If we decide on a visualizer I can also do some work on that then, otherwise I have time to work on it this weekend (Sat/Sun).

rebeccabilbro:
@NealHumphrey ok, sounds good -- I've posted responses to the rest of your comments within the issues. Let me know if you find any good article/blog post candidates for our #125 card. I thought [this one](https://delta1epsilon.github.io/2017/When-Scatter-Plot-doesn't-work/) was kind of interesting, but I'll keep looking too.

PR #142, opened 2017-03-04
WIP - Neal's user testing example

NealHumphrey:
Not really presentation ready - demonstrates a couple usability issues with large multi-class classifiers (images get cramped with too many classifiers), as well as a bug in the ClassificationReport

Better not to merge this in yet, but doing PR to share w/ @rebeccabilbro and others. In particular, note the use of Seaborn confusion matrix (so we can improve on it in #139 ) and the Yellowbrick bug identified in #141


NealHumphrey:
Also, I seemed to need to add the line to __init__ to get the ClassBalance visualizer to import - this should be replicated/verified by someone else. I'll open a bug report for that too.

bbengfort:
See #121

yellowbrick/__init__.py
```diff
@@ -31,6 +31,7 @@
 
 # Import yellowbrick functionality to the top level
 from .anscombe import anscombe
+from .classifier import ROCAUC, ClassBalance, ClassificationScoreVisualizer
```
bbengfort: Ok, I think I see what you mean now; @rebeccabilbro and I were debating whether or not to hoist everything into the top level like NLTK does or to keep things in their packages. To import these tools you'd do:

```python
from yellowbrick.classifier import ROCAUC, ClassBalance, ClassificationScoreVisualizer 
```

My guess is that you're doing something similar to:

```python
import yellowbrick as yb 

yb.ClassBalance()
```

Is that right? 

Frankly, I'm torn about whether or not to hoist these to the top level; I'd say things need to be hoisted to the second level (e.g. text, features, classifier, or regressor). Since we're going to be creating directory models for these soon, my sense is that we should leave them as is.

NealHumphrey: I actually used it like this: 
`yb.classifier.ClassBalance(model, classes=city_options)`

I can't remember if I also tried importing i.e. from yellowbrick.classifier import ClassBalance (maybe tried it and abandoned it?)

Agree we'll hoist to second level. My opinion would be that scikit-learn has too many levels and could benefit from more hoisting, but that separate groups for features, classifiers, and regressors is a good package level to bring each visualizer to - the level defines the meaning of what it would be used for, and then you can put all the visualizer options into the root of that usage level.

bbengfort:
@NealHumphrey bump

NealHumphrey:
@bbengfort what do you want me to do on this one? This notebook demonstrates a few bugs/issues I found, so at first I didn't suggest merging this since it But per our discussion yesterday sounds like we want to see the bugs? I think it can be merged as is. I need an approved review to do this though. 

Small side note - I see you updated this branch via the 'Update branch' button on Github. I suggest that only the branch author be in charge of this, so they can do it via command line or via Github. Not a big problem, but I had a few merge conflicts between my find_text_color and my confusion_matrix branch and it would have been slightly cleaner/easier to resolve them if the most recent develop branch hadn't been merged in already.

bbengfort:
Review approved - go for it.

PR #160, opened 2017-03-08
User testing notebook - fun_3000 research lab

lcombs:
User testing example using https://archive.ics.uci.edu/ml/datasets/Abalone dataset. Part of @ayota @lauralorenz @lempira research lab.


NealHumphrey:
@lcombs or @bbengfort you can change the base branch to merge into by clicking the 'Edit' button at the top right of this pull request page (beside the title). Then look for the drop down menu just under the title. Any of the maintainers can make this change. 

@bbengfort it might be a good idea to change the default branch in the repo settings to be `develop` instead of `master`, if you do that it is the automatically chosen branch for any new pull requests. Same place that you set up the branches to require review.

bbengfort:
@NealHumphrey changed default branch to develop for now just to see if that works better. Master is the release/stable branch -- so we do want to make sure that's what folks are _using_, but I guess most folks will pip install yellowbrick, so we can manage the release life cycle.

Issue #58, opened 2016-10-05
Add InformativeFeatures class


rebeccabilbro:
Need to add a class to enable the user to evaluate the features that were most informative for a fitted model. This class will inherit from `ScoreVisualizer`

rebeccabilbro:
http://scikit-learn.org/stable/auto_examples/ensemble/plot_forest_importances.html

rebeccabilbro:
In Scikit-Learn, it looks like most supervised methods have a `coef_` or `feature_importances_` method to determine the most important features. 

Examples that use `feature_importances_`:
 - http://scikit-learn.org/stable/modules/generated/sklearn.svm.LinearSVC.html
 - http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.GradientBoostingRegressor.html
 - http://scikit-learn.org/stable/modules/generated/sklearn.tree.DecisionTreeClassifier.html
 - http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html
 - http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestRegressor.html
 - http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.AdaBoostClassifier.html
 

Examples that use `coef_`:
 - http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html
 - http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html
 - http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.Lasso.html#sklearn.linear_model.Lasso
 - http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.ElasticNet.html#sklearn.linear_model.ElasticNet

bbengfort:
@rebeccabilbro adding the comments I wrote on this from the paper. Also; I'd say this is a feature, not technical debt? 

Generalized linear models compute a predicted independent variable via the linear combination of an array of coefficients with an array of dependent variables. GLMs are fit by modifying the coefficients so as to minimize error and regularization techniques specify how the model modifies coefficients in relation to each other. As a result, an opportunity presents itself: larger coefficients are necessarily "more informative" because they contribute a greater weight to the final prediction in most cases. Additionally we may say that instance features may also be more or less 
"informative" depending on the product of the instance feature value with the feature coefficient. This creates two possibilities:

[...]

PR #164, opened 2017-03-09
WIP top coefficients visualizer for text classifier

ayota:
Looking at way to visualize top coefficients for multiple classes (not really sure if this makes sense, but this is the rabbit hole I've headed down).

This is contrib from the fun_3000 folks; @lempira, @lauralorenz, @lcombs 

**sample of single class visualized**
sports

**sample of all classes visualized -- wip with multiclass visualizer**
index


bbengfort:
@ayota would you mind adding an example of the visualization to the thread here; you can just drag and drop images into the comment box to upload them.

bbengfort:
@ayota are you ready for me to do a code review on this? I have some suggestions about where we can take your prototype and put it into the main body of the code.

ayota:
@bbengfort not ready for review yet 😄 . I want to add docstrings and rework it back to how it should be in light of work on the multiplot class, which would solve the hack-y way I made it possible to display all the class graphs at once. definitely should have this wrapped by wednesday. thanks!

bbengfort:
@ayota ok, perfect - I just wanted to make sure I wasn't the hold up! Let me know when you're ready.

rebeccabilbro:
@ayota bump! 

What's the status of this feature? Any chance you're thinking of wrapping it up in the next two weeks? After 5/3 we're switching into version bump mode for the 0.4.0 release in advance of PyCon.

ayota:
This could eventually help with #58 if made more general to handle estimators with feature_importances_ instead of coef_

yellowbrick/classifier/top_features.py
```diff
@@ -0,0 +1,253 @@
...
+-   Implement one-sided version that takes absolute value of coefs/features
+"""
+
+##########################################################################
+## Imports
+##########################################################################
+
+import numpy as np
+import warnings
+import matplotlib.pyplot as plt
```
bbengfort: This import is unused (and also makes me nervous since all operations should happen on the ax object). Would you mind removing it?

yellowbrick/classifier/top_features.py
```diff
@@ -0,0 +1,253 @@
...
+
+from .base import ClassificationScoreVisualizer
+from ..exceptions import YellowbrickValueError
+from ..style import color_palette
+
+##########################################################################
+## Top Features Visualizer
+##########################################################################
+
+class TopFeaturesVisualizer(ClassificationScoreVisualizer):
```
bbengfort: I suggest a rename to `TopNFeatures`; this might make it a bit easier to type (you can also alias the class to `TopFeaturesVisualizer` if you'd like).

yellowbrick/classifier/top_features.py
```diff
@@ -0,0 +1,253 @@
...
+    target = corpus.target
+    vectorizer = CountVectorizer(stop_words='english')
+    vectorizer.fit(data)
+    X_train = vectorizer.transform(data)
+    features = vectorizer.get_feature_names()
+
+    """
+
+    def __init__(self, model, ax=None, N=20, features=None, **kwargs):
+        # TODO: Make sure model is a classifier; otherwise throw a warning.
```
bbengfort: The `ClassificationScoreVisualizer` will actually do this check for you and will throw an exception if it's not. So by calling super you're ensuring it's a classifier.

yellowbrick/classifier/top_features.py
```diff
@@ -0,0 +1,253 @@
...
+        # Check to see if there are N features; otherwise replace with count of features
```
bbengfort: Not sure what you mean here; are you going to do a check of N to features? I think that's actually a good idea, what happens if `N != len(features)`?

yellowbrick/classifier/top_features.py
```diff
@@ -0,0 +1,253 @@
...
+
+        super(TopFeaturesVisualizer, self).__init__(model=model, ax=ax, **kwargs)
+
+        self.N = N
+        self.features = features
+
+
+    def fit(self, X_train, y, **kwargs):
```
bbengfort: Everything that you're doing here is done by the super class, so you don't need to specify a `fit()` method unless you're going to do something extra. (Also make sure to update your branch to take advantage of the super class features; let me know if you'd like me to do it)

yellowbrick/classifier/top_features.py
```diff
@@ -0,0 +1,253 @@
...
+        Call draw().
+        """
+
+        # raise error is no class label provided
+        if class_label is None:
+            raise YellowbrickValueError(
+                "Please provide a class label."
+                )
+
+        self.class_label = class_label
```
bbengfort: If you're storing this here, could the user also pass in the `class_label` at `__init__`? Or is it just too complicated?

yellowbrick/classifier/top_features.py
```diff
@@ -0,0 +1,253 @@
...
+        """
+
+        self.estimator.fit(X_train, y, **kwargs)
+        self.classes_ = self.estimator.classes_
+
+        return self
+
+    def score(self, class_label=None, **kwargs):
+        """
+        Extract top N features for desired class label.
```
bbengfort: Need parameters and user-oriented docstring

yellowbrick/classifier/top_features.py
```diff
@@ -0,0 +1,253 @@
...
+        if 2 * self.N > len(self.top_coefficients):
+            self.N = int(len(self.top_coefficients)/2)
+            warnings.warn(
+                'Value of N exceeds number of features. Plotting all features instead.'
+            )
+
+        palette = color_palette(n_colors=2)
+        colors = [palette[0] if c < 0 else palette[1] for c in self.top_coefficients]
+        self.ax.bar(np.arange(2 * self.N), self.top_coefficients, color=colors)
+        self.ax.set_xticks(np.arange(2 * self.N))
```
bbengfort: Very nice draw method!

yellowbrick/classifier/top_features.py
```diff
@@ -0,0 +1,253 @@
...
+        self.top_coefficients = coefs[top_index]
+
+        # get pretty names if they exist
+        try:
+            self.top_features = [features[i] for i in top_index]
+        except TypeError:
+            self.top_features = top_index
+            warnings.warn(
+                'No feature names specified. Index values will be used.'
+            )
```
bbengfort: I think this is a pretty solid start!

tests/test_classifier/test_top_features.py
```diff
@@ -0,0 +1,121 @@
...
+# Copyright (C) 2017 District Data Labs
+# For license information, see LICENSE.txt
+#
+# ID: test_top_features.py [] $
+
+"""
+Testing for the top N features visualizer
+
+TODO:
+-   Cover all estimators outlined in issue #58
```
bbengfort: I think this is a pretty important todo.

tests/test_classifier/test_top_features.py
```diff
@@ -0,0 +1,121 @@
...
+        try:
+            model = LinearSVC()
+            visualizer = TopFeaturesVisualizer(model, N=10)
+            visualizer.fit(X, y)
+            visualizer.score(class_label=0)
+            visualizer.poof()
+        except Exception as e:
+            self.fail('Test of N > number of features failed.')
+
+    # def test_warn_features(self):
```
bbengfort: Rather than commenting, you can add a skip, which will remind us to implement it:

```python
@unittest.skip("not implemented yet")
def test_warn_features(self):
    ... 
```

tests/test_classifier/test_top_features.py
```diff
@@ -0,0 +1,121 @@
...
+            visualizer = TopFeaturesVisualizer(model, N=5, features=self.features)
+            visualizer.fit(self.X_train, self.target)
+            visualizer.score(class_label='sports')
+            visualizer.poof()
+        except Exception as e:
+            self.fail('Basic visualizer test failed.')
+
+    def test_too_many_features(self):
+        """
+        Make sure replaces N with number of features in data
```
bbengfort: Make sure that the docstring comment for a test is one line -- this string is what goes into the test report.

tests/test_classifier/test_top_features.py
```diff
@@ -0,0 +1,121 @@
...
+        when N is greater than number of features.
+        """
+        try:
+            model = LinearSVC()
+            visualizer = TopFeaturesVisualizer(model, N=10)
+            visualizer.fit(X, y)
+            visualizer.score(class_label=0)
+            visualizer.poof()
+        except Exception as e:
+            self.fail('Test of N > number of features failed.')
```
bbengfort: So you're expecting no exceptions when `N > len(features)`?

tests/test_classifier/test_top_features.py
```diff
@@ -0,0 +1,121 @@
...
+    #     """
+    #     pass
+    #
+    # def test_error_nolabel(self):
+    #     """
+    #     Raise error is no class label passed.
+    #     """
+    #     pass
+
+    # def test_param_coef(self):
```
bbengfort: I think this needs to be done before merge.

tests/test_classifier/test_top_features.py
```diff
@@ -0,0 +1,121 @@
...
+    #     """
+    #     Make sure estimators with coef params work properly.
+    #     """
+    #     pass
+    #
+    # def test_param_feature_importance(self):
```
bbengfort: I think this needs to be done before merge.

tests/test_classifier/test_top_features.py
```diff
@@ -0,0 +1,121 @@
...
+    #     """
+    #     Make sure estimators with feature importance params raise errors.
+    #     """
+    #     pass
+    #
+    # def test_param_none(self):
```
bbengfort: I think this needs to be done before merge.

yellowbrick/classifier/top_features.py
ayota: My idea was to make this able to plot multiple class labels at this point instead of having to do a new visualizer, but my kicking the tires here haven't worked out, so I will move up to `__init__` and add support for when `coef_` doesn't yield the coefficients split out by classes.

yellowbrick/classifier/top_features.py
ayota: Yea, so basically the idea (not well implemented by my test lolol) is that we need to make sure there actually **are** N features to visualize (e.g., on the iris dataset there are 4 features. If I ran this visualizer asking for the top 10, it just shows an empty plot and an axes error. This would catch it and just show all four features and a warning.)

So to be overly specific:
**1. `N <= len(features)`**: The use case for this is typically text data with eleventy bajillion features. This would just take the top N of them, or all of them if `N==len(features)`.

**2. `N > len(features)`**: This causes an error. So, until I have time to really smooth this out, I wanted to just:
1. Replace the N param with `len(features)` 
2. Throw a warning that it did this

In most cases I figure this would just be counting error, and should still show a graph. But if the graph is wonky, then there's a way to know what possibly went wrong.

**3. `N==None`**: User don't care; this would just show all the features.

Any ideas of how to do this more gracefully would be much appreciated; or of any other scenarios I am not covering!

bbengfort: This is a good strategy, and well thought out - if you could document it in the docstring or in comments that would be great; but I don't think there is a more graceful solution. Nice one!

yellowbrick/classifier/top_features.py
bbengfort: I guess I'm still not sure what the plan is, but sounds like you have one -- so go for it!

bbengfort:
@ayota hopefully my comments helped; ping me when you're ready for a merge review.

bbengfort:
@ayota how can we take this across the finish line? I'm happy to help however I can!

bbengfort:
@ayota closing this in favor of #317 which does something similar; potentially in the future, we can add a horizontal argument to make it look like this (added in #322) especially when #335 is complete.

PR #169, opened 2017-03-09
Add 2D ScatterViz  with tests (Feature Visualizer)

ndanielsen:
Adds a basic 2D scatter plot as a feature visualizer.

(It's a great building block for more complex visualizations)

scatterplot example


bbengfort:
@ndanielsen awesome, this looks great; I just did a quick check through; are you ready for me to do a code review so that you can do a PR?

Also, @NealHumphrey -- I really appreciate your participation, would you also be willing to do a review for htis PR?

Also note that I updated your branch to the latest, I don't think it should change anything since your PR created new files.

yellowbrick/features/scatter.py
```diff
@@ -0,0 +1,244 @@
...
+                if is_dataframe(X) and X.columns == 2:
+                    self.features_ = X.columns
+
+                # Otherwise create numeric labels for each column.
+                elif ncols == 2:
+                    self.features_ = [
+                        str(cdx) for cdx in range(ncols)
+                    ]
+                else:
+                    raise YellowbrickValueError('ScatterVisualizer only accepts two features, please explicted set the features key argument or only pass a matrix/ dataframe with only two columns')
```
NealHumphrey: Your error message says " please [explicitly] set the features key argument or only pass a matrix/ dataframe with only two columns...." In combo with row 162, this seems to imply you meant for the 'features' argument to be able to slice you dataframe using column names? But, this doesn't happen anywhere. It looks like you need to add the `else` statement to correspond to row 162 that decides how to use `features_` to slice the `X` dataframe to match your `__init__` docstring. In addition, change your `fit` docstring describing what X is. 

OR, abandon this bonus functionality and require an m-width X and edit all documentation to match. It's a nice-to-have feature but seems like it might be something we want to only implement if it is done consistently across other feature visualizers.

tests/test_features/test_scatter.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        Assert that only two features are allowed for this visualizer if not
+        set in the init
+        """
+        visualizer = ScatterViz()
+        with self.assertRaises(YellowbrickValueError) as context:
+            visualizer.fit_transform(self.X, self.y)
+            self.assertTrue('only accepts two features' in str(context.exception))
+
+
```
NealHumphrey: If implementing the Dataframe slicing functionality described in my comment below, should add a test for it.

NealHumphrey:
@bbengfort don't mind at all, I just added my review!

NealHumphrey:
Note @ndanielsen I added this as an issue rather than adding this to the pull request review since it shouldn't block merging this in. I tried extending your  scatter plot by pulling it into the MultiVisualizer I worked on this sprint, as an example. I wanted to be able to slice the data to only plot one class at a time, but I got an index error. 
https://github.com/DistrictDataLabs/yellowbrick/issues/175

If you're interested, here's the branch I was working on for the multivisualizer mashup. Right now it plots your plot 3 times instead of 3 plots with one class each.
https://github.com/NealHumphrey/yellowbrick/tree/multi-visualizer-with-scatter/examples/nealhumphrey

yellowbrick/features/scatter.py
```diff
@@ -0,0 +1,244 @@
...
+        """
+        # Get the shape of the data
+        nrows, ncols = X.shape
+
+        # Create the axes if they don't exist
+        if self.ax is None:
+                self.ax = plt.gca(xlim=[-1,1], ylim=[-1,1])
+
+        color_values = get_color_cycle()
+        colors = dict(zip(self.classes_, color_values))
```
ndanielsen: add if not colors

ndanielsen:
@balavenkatesan

tests/test_features/test_scatter.py
ndanielsen: This is a bit hard to test as we don't have pandas as a testing dependency.

bbengfort: @ndanielsen for now, feel free to add the tests, just do what @rebeccabilbro did in the pos visualizer tests: [test_postag.py#L60](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_text/test_postag.py#L60) -- use a `skipIf` if there is an `ImportError` for Pandas. At the very least we can then test locally when pandas is installed, but skip tests on continuous integration where it's harder to get these things installed.

ndanielsen: okay will do!

ndanielsen: very clever way of testing this

ndanielsen:
@NealHumphrey thanks for the great review on this before. I made the requested changes and added the feature that i misworded at first but then realized from your review that it was a great idea.

@bbengfort FYI

bbengfort:
@ndanielsen meaning you're ready for me to review it?

ndanielsen:
@bbengfort please go ahead and give it a review

yellowbrick/features/scatter.py
```diff
@@ -0,0 +1,278 @@
...
+        The axes to plot the figure on.
+
+    features : list of strings
+        The names of two features or columns.
+        More than that will raise an error.
+
+    classes : list of strings
+        The names of the classes in the target
+
+    color : list or tuple of colors
```
bbengfort: This is the correct docstring style.

yellowbrick/features/scatter.py
```diff
@@ -0,0 +1,278 @@
...
+        Specify the colors for each individual class
+
+    colormap : string or matplotlib cmap
+        Sequential colormap for continuous target
+
+    :param markers: iterable of strings
```
bbengfort: This is not the correct docstring style.

yellowbrick/features/scatter.py
```diff
@@ -0,0 +1,278 @@
...
+                 markers=None,
+                 **kwargs):
+        """
+        Initialize the base scatter with many of the options required in order
+        to make the visualization work.
+
+        Parameters
+        ----------
+
+        :param ax: the axis to plot the figure on.
```
bbengfort: I know the docstring thing has been confusing at best; but we're slowly but surely getting all together on the docstring format. Would you mind shifting these docstrings to the new format?

yellowbrick/features/scatter.py
```diff
@@ -0,0 +1,278 @@
...
+
+        :param features: a list of feature names to use
```
bbengfort: Is there a way to control which feature is on the x axis and which is on the y?

yellowbrick/features/scatter.py
```diff
@@ -0,0 +1,278 @@
...
+        if ncols == 2:
+            X_two_cols = X
+            if self.features_ is None:
+                self.features_ = [0, 1]
+
+        # Handle the feature names if they're None.
+        elif self.features_ is not None and is_dataframe(X):
+            X_two_cols = X[self.features_].as_matrix()
+        else:
+            raise YellowbrickValueError("""
```
bbengfort: What happens if I pass a matrix with > 2 columns, but then supply features as indices, e.g. features=[2,8]? Also, as it turns out, numpy does provide [named arrays](https://docs.scipy.org/doc/numpy-1.10.1/user/basics.rec.html).

yellowbrick/features/scatter.py
```diff
@@ -0,0 +1,278 @@
...
+                ScatterVisualizer only accepts two features, please
+                explicitly set these two features in the init kwargs or
+                pass a matrix/ dataframe in with only two columns.""")
+
+        # Store the classes for the legend if they're None.
+        if self.classes_ is None:
+            # TODO: Is this the most efficient method?
+            self.classes_ = [str(label) for label in set(y)]
```
bbengfort: This is perfectly fine, but I wanted to point out `np.unique`, which I've been considering  using instead of the python `set` -- I don't know if it's faster or not, but if it is, it might be better. If you have any thoughts, I'd be happy to hear them.

bbengfort:
@NealHumphrey -- were your changes sufficiently addressed?

yellowbrick/features/scatter.py
ndanielsen: I think that the best way to handle this is via the order of the features. First feature is X, second is Y.

yellowbrick/features/scatter.py
ndanielsen: Updated these.

yellowbrick/features/scatter.py
ndanielsen: np.unique seems much faster.

screen shot 2017-05-01 at 8 51 22 pm

yellowbrick/features/scatter.py
ndanielsen: I'll add the grabbing features by via indexes of narrays without names.

I added in `is_structured_array` similar to is_dataframe, and came across an interesting internal mplt error. 

`TypeError: ufunc 'invert' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''`

Take another crack at this tomorrow.

yellowbrick/features/scatter.py
bbengfort: Ok, then we should probably specify that in the docstring. 

In the future, would it be ok if I (or someone else) also added optional `x` and `y` keyword arguments in addition to the `features` argument such that either `x,y` must be specified or `features` may be specified, but not both? It would only change the signature of the function, not any other functionality.

yellowbrick/utils.py
```diff
@@ -184,6 +184,28 @@ def is_dataframe(obj):
 isdataframe = is_dataframe
 
 
+def is_structured_array(obj):
```
bbengfort: Nice - potentially in the future we could create a single function, `has_named_columns` that does `is_dataframe() or is_structured_array()` since this seems like it will be the common use-case.

yellowbrick/features/scatter.py
```diff
@@ -0,0 +1,286 @@
...
+        if ncols == 2:
+            X_two_cols = X
+            if self.features_ is None:
+                self.features_ = ["Feature One", "Feature Two"]
+
+        # Handle the feature names if they're None.
+        elif self.features_ is not None and is_dataframe(X):
+            X_two_cols = X[self.features_].as_matrix()
+
+        # TODO handle numpy named arrays
```
bbengfort: Awesome, thanks for adding this TODO in.

tests/test_features/test_scatter.py
```diff
@@ -0,0 +1,188 @@
...
+        X.columns = [
+            "temperature", "relative_humidity", "light", "C02", "humidity"
+        ]
+
+        # Test the visualizer
+        features = ["temperature", "relative_humidity"]
+        visualizer = ScatterViz(features=features)
+        visualizer.fit_transform_poof(X, y)
+
+    # def test_integrated_scatter_numpy_named_arrays(self):
```
bbengfort: Not a big deal, but you could also decorate with @unittest.skip("not implemented yet") instead of commenting out the function. The benefit of the skip is that the tests will show the skipped test but not fail, alerting us that we have to implement the test sometime in the future.

yellowbrick/features/scatter.py
```diff
@@ -0,0 +1,286 @@
...
+        else:
+            raise YellowbrickValueError("""
+                ScatterVisualizer only accepts two features, please
+                explicitly set these two features in the init kwargs or
+                pass a matrix/ dataframe in with only two columns.""")
+
+        # Store the classes for the legend if they're None.
+        if self.classes_ is None:
+            # TODO: Is this the most efficient method?
+            self.classes_ = [str(label) for label in np.unique(y)]
```
bbengfort: Ok, cool - glad to know that's faster; I might run some benchmarks to see this in action; I'll let you know when I do them.

NealHumphrey:
All set from my end

yellowbrick/features/scatter.py
ndanielsen: sounds good, make the change and pushing it up soon

yellowbrick/utils.py
ndanielsen: just got it working as my brain isn't mush today. pushing this up

yellowbrick/features/scatter.py
ndanielsen: Got the named arrays working, just needs to transform the array a bit. Also added in the selection by indices.

ndanielsen:
@bbengfort added the remaining items on this (support for named arrays, selection by array indices, and the x,y in the init methods). I'll be able to port this stuff over nicely to the other visualizer that I'm finishing up. Thanks again for the great code review!

bbengfort:
@ndanielsen ok, update with the base branch and merge when ready; see you in a bit!

PR #172, opened 2017-03-16
Feature 154 light vs dark text

### LINKED ISSUE #154: Create utility function for determining light vs. dark text when given a specific palette color.
This might go in .util, or .style. Should be callable from any class, so that proper text default choice can be used. 

Proposed approach:
```
def textColor(baseColor, textColors=("black","white")):
    # baseColor: the color of the background, in any valid color specification format
    # textColors: a tuple of (dark, light) text format. 
    if baseColor is dark:
        return textColors[0]
    elif baseColor is light:
        return textcolors[1]
```
This lets the user pass in specific light and dark colors to get back if they don't want white vs. black, but allows them to use it in the text function call directly as it actually returns the color (instead of just returning a boolean or something that must be interpreted). 

This should especially be used in the Heatmap classifier visualizers, e.g. ClassificationReport and ConfusionMatrix

NealHumphrey:
Resolves #154 . 
Added a utility function that determines if a background color is light or dark, and returns the appropriate text color (black or white) as needed.

Added this to the confusionMatrix visualizer; will add to ClassificationReport as part of refactor.

Note this is built off of #144, so that pull request should be accepted before this one to make sure not to get unexpected behavior. 

Here's a demo:
image


NealHumphrey:
@bbengfort or @rebeccabilbro this is ready for review!

yellowbrick/utils.py
```diff
@@ -155,6 +157,45 @@ def is_dataframe(obj):
...
+    Ufunc-extension that returns 0 instead of nan when dividing numpy arrays
+    example: >> div0( [-1, 0, 1], 0 ) -> [0, 0, 0]
+    """
+    with np.errstate(divide='ignore', invalid='ignore'):
+        c = np.true_divide( a, b )
+        c[ ~ np.isfinite( c )] = 0  # -inf inf NaN
+    return c
+
+
+def find_text_color(base_color, text_colors=("black","white")):
```
bbengfort: So I think this would be better in the style module rather than utils. Perhaps create a utils.py in style and import this function to the root? What do you think?

bbengfort: In terms of the signature of the function, why have you chosen to use a tuple of text_colors? Wouldn't the below function signature be just as simple, but perhaps a bit more explanatory (for example, if I wanted to make the light color blue but keep the dark color the default):

```
def find_text_color(base_color, dark_color="black", light_color="white"):
    pass
```

While we're at it, there is a semantic repetition of the `_color` that could probably be omitted since it's part of the function signature, but that's more stylistic than anything else. 

```
def find_text_color(background, dark_text="black", light_text="white"):
    pass
```

yellowbrick/utils.py
```diff
@@ -155,6 +157,45 @@ def is_dataframe(obj):
...
+    Users can specify the dark and light text color, or accept the defaults of 'black' and 'white'
+
+    # baseColor: the color of the background, in any valid color specification format
+    # textColors: a tuple of (dark, light) text format. 
+    '''
+
+    #matplotlib cmap exports rgba colors scaled 0-1 instead of standard 255
+    rgb=[]
+    for idx,val in enumerate(base_color):
+        rgb.append(val*255)
```
bbengfort: I think this would be more understandable as:

```
# multiply the components of the cmap rgb ([0,1]) by 255 to get standard scale. 
rbg = tuple(map(lambda c: c * 255, base_color))
```

Potentially you could also simply divide the constants starting on line 187 by 255.0 and not worry about performing this mapping.

bbengfort: And also, it might be helpful to have a utility function that does this for us if we're going to routinely scale.

yellowbrick/utils.py
```diff
@@ -155,6 +157,45 @@ def is_dataframe(obj):
...
+    
+    #Calculate brightness relative to the color type
+    base_brightness = math.sqrt( .241*rgb[0]*rgb[0]+
```
bbengfort: So perhaps you mentioned this in the related issue, but I'm not sure where the coefficients used in this luminescence algorithm  come from. For example, I found the following: `sqrt(0.299 * R^2 + 0.587 * G^2 + 0.114 * B^2)` from http://stackoverflow.com/questions/596216/formula-to-determine-brightness-of-rgb-color -- but there are clearly other options, including for gamma-correction, etc. 

One way to deal with this would be to put the coefs in a variable, then we could change them with different arguments in the function signature:

```python
coefs = np.array((.241, .691, .068))
rgb = np.array(rgb)
brightness = np.sqrt(np.dot(coefs, rgb**2))
```

yellowbrick/utils.py
```diff
@@ -155,6 +157,45 @@ def is_dataframe(obj):
...
+                                 .691*rgb[1]*rgb[1]+ 
+                                 .068*rgb[2]*rgb[2] 
+                                ) 
+
+    if base_brightness > 130:
```
bbengfort: And as a follow up, where does this threshold come from?

yellowbrick/utils.py
```diff
@@ -155,6 +157,45 @@ def is_dataframe(obj):
...
+        #base is light
+        return text_colors[0]
+    else:
```
bbengfort: point of minor annoyance that I think would be best practice to help us out in open source land, convert:

```python 
if bool:
    return a 
else:
    return b 
```

to 

```python

if bool:
    return a 
return b 
```

The fear is that the else or conditional gets modified in such a way that return None becomes possible, which can break things in such a way so as to be very difficult to solve! 

If you don't want to change it, I won't fight you, but thought I'd comment since this relates to friendliness rather than style.

yellowbrick/utils.py
bbengfort: also, if you're going to go with the numpy array as below, then this is just 

```
rgb = np.array(background) * 255 
```

bbengfort:
@NealHumphrey awesome work! This is going to come in really helpful in the future! I added some comments in the code review; only one that has to be addressed, related to the one from the previous commit (just the examples folder management). Very cool stuff, glad to have these features in!

yellowbrick/utils.py
NealHumphrey: Good point on breaking out the tuple. It was because I was at first imagining that the function might be designed to take more than 2 text colors, but it seemed like a 'middle' color would probably never really be used so didn't implement it. 

I'd vote for keeping the `_color` part of the variable definition to help keep it clear what type of variable type is expected. `dark` and `light` can be any valid matplotlib color definition (so `"black"` works but so does a hex code or rgb).

NealHumphrey:
@bbengfort I made most all of the changes you suggested. I added a link to the source I used for the coefficients and threshold - it wasn't very scientific, but it was the first one I found that had both coefficients and a threshold, though the author decided the threshold by trial and error. I tested it and it seemed to work to my eye, so left it. 

I added in another coefficient option and now demo both in the ipynb, so if you disagree we can switch. Results are only different for one base color. 

I dug around a lot trying to find how to incorporate the 'A' of RGBA into this calculation but every example I found only dealt with RGB. It would be good to incorporate this in the future, though I don't think any of our current palettes use transparency. I clarified this limitation in the docstring, as well as the limitation of RGBA needed for base_color (as opposed to dark_color and light_color that can be any valid matplotlib color)

bbengfort:
@NealHumphrey so I don't think you need to include the alpha in the computation in principle. The problem with including alpha is that you don't know the background color behind the alpha (though it's likely white). If someone does specify an alpha, my reaction is that they would want the entire cell at that alpha, so it should just be carried over and appended to either the light or dark color, whichever one is selected. 

I've approved the review, as soon as the checks are complete you may merge when ready.  I think you want to chose the "rebase and merge" option.

PR #176, opened 2017-03-22
WIP: Basic version of the multiplevisualizer base class

NealHumphrey:
Work in progress - just adding this as a pull request so we can discuss at the sprint planning meeting whether this conceptual approach fits with how we might make various types of multi-plot visualizers in the future (small multiples or dashboards). 

Issues to resolve:
- Should this be a mixin?
- Can we expect to be able to set the ax of a visualizer using vizualizer.ax after __init__ in all cases?

Future work:
Need to extend various features so it performs as expected. Examples: mXn subplots not handled correctly, fewer visualizers in self.visualizers than subplot spaces available (e.g. 2x2 grid with only 3 visualizers), smart handling of labels based on sharex/sharey or not.


bbengfort:
@NealHumphrey thoughts with what we should do with this PR?

NealHumphrey:
Good question. It's not super stale obviously. As I recall I think you had a different idea about a competing approach to this type of problem? It's been a while. 

It might be most efficient to talk through it with you and @rebeccabilbro and @ndanielsen - I'm 9 months out of touch with the current development goals. If it's generally of interest I could do a small bit of extra work to get it usable enough to be in the codebase as a starting point, otherwise we can ditch it.

NealHumphrey:
So if I remember correctly your competing idea was visual pipelines. Was any progress ever made on that?

This proposal had two potential uses:
- creating any sort of visualizer that needs multiple subplots (e.g. small multiples). For that purpose I think it's better used as a design principle/example for people creating those types of visualizations to show a simple use of subplots, which they can use as a template for their code - while it could be a mixin or integrated into a base class I think that approach will add unnecessary complexity that will be hard for occasional contributors to use - better to keep each visualizer a bit more contained. 
- A quick and dirty visual pipeline - create an instance of the MultipleVisualizers, and then you can use it with fit/scale/poof or whatnot to produce a set of visualizations at once. At the bottom of the jupyter notebook is an example of this combining radviz and parallel coordinates as a set of feature selection visualizers. For this purpose the code is OK to use as is, though some enhanced features would make it more useful. Top two enhanced features would likely be a) checking to see if the visualizers are of the same type (feature visualizers vs. model visualizers, to make sure they are run in the same pattern i.e. `.score()` used or not). and b) ability to add visualizers to the MultipleVisualizers list at a stage later than initialization (currently you ahve to pass all visualizers into `__init__`). Not a hard fix.

NealHumphrey:
Here's the code in that Jupyter notebook showing the second use case:

```
from yellowbrick.features.radviz import RadViz
from yellowbrick.features.pcoords import ParallelCoordinates
from yellowbrick.multivisualizer.base import MultipleVisualizer

visualizers = [RadViz(classes=classes, features=features),
              ParallelCoordinates(classes=classes, features=features)]

mv = MultipleVisualizer(visualizers)
mv.fit(X,y)
mv.poof()
```
Produces a single plot with two subplots, one for each visualizer

NealHumphrey:
Actually, come to think of it - for the first use case (demo of using subplots in a visualizer), this could be turned into a blog post for potential contributors called 'How to build a visualizer with subplots'. 

One thing that's been on my 'interested in building' list is a port of the "ManyModels" class and the [accompanying bar chart](https://github.com/orgs/DistrictDataLabs/projects/2#card-7610035) of precision/recall that I built for the capstone. This is one potential answer to the [Rapid Experimentation...](https://github.com/orgs/DistrictDataLabs/projects/2#card-7610035) card that @bbengfort put into the research lab idea list. I could build this precision/recall bar chart for a manymodels class and write up a blog post about building a visualizer with subplots using this as the example.

bbengfort:
@NealHumphrey so I do still like the idea, it feels like this is our version of a visualizer-specific `sns.FacetGrid`. I can think of a lot of different uses for this, though like the `sns.FacetGrid`, I assume it will be very difficult for end-users but better for internal multi-visualizers. 

As to your question of visual pipelines, some progress has been made on that, but it's still in a feature branch of mine. It is my priority to get that done on or around PyCon. However, this idea could be a good complement to pipelines, e.g. there could be a "flip book" mode of the pipeline, where each visualizer is displayed in a different figure and a "grid mode" that uses this. In fact, we should probably implement both these modes as MultipleVisualizers. 

Personally, I think it would be good to move forward on this, get an implementation together and merged in, and then I can use it as a base for visual pipelines.

NealHumphrey:
Ok, if that's the case then I'll pick this back up and get it merge-ready. Todo:
- [ ] merge in current dev
- [ ] enhance the strawman version to be more robust, most notably some better methods of appending visualizers after initialization.
- [ ] Add tests
- [ ] Add a bit of inline documentation on usage, but for the purpose of getting it into the codebase for next stage develepment this will mostly be a 'hidden' feature. 

I'll push that all to this branch so this will be a ready to go PR. Then, for follow on work:
- [ ] Implement a version of the 'manymodels' class and a bar chart classification report, modeled on the example from my capstone cited above. 
- [ ] modify the base MultipleVisualizers as needed so that it works as a mixin or base class, whatever's most appropriate per this example.
- [ ] draft a quick blog post with a target audience of YB contributors that shows how to use this base class for making your own multi-visualizer.

bbengfort:
Perfect, thanks @NealHumphrey!

bbengfort:
@NealHumphrey noticed some commits -- let me know when you're ready for a review.

NealHumphrey:
@bbengfort OK I'm ready for a review now. 

You can check out my Jupyter Notebook (`examples/NealHumphrey/MultipleVisualizerExample`) for demo usage. Currently I am just demoing this by putting some existing visualizers onto subplots. However, I expect that for other uses (like the bar chart mentioned above) tailor-made visualizers that are subplot-ready may be needed. 

For the purpose of creating set of visualizers that can be plotted as a group, here's demo usage and output (copied from the notebook):
```
from sklearn.linear_model import LogisticRegression
from yellowbrick.classifier import ConfusionMatrix
from yellowbrick.classifier import ClassBalance

model = LogisticRegression()

visualizers = [ClassBalance(model),ConfusionMatrix(model)]
mv = MultipleVisualizer(visualizers, ncols=2)
mv.fit(X_train, y_train)
mv.score(X_test, y_test)
mv.poof()
```
image

yellowbrick/base.py
```diff
@@ -407,3 +408,119 @@ def predict(self, X, y):
...
+    >>> model = LogisticRegression()
+    >>> visualizers = [ClassBalance(model),ConfusionMatrix(model)]
+    >>> mv = MultipleVisualizer(visualizers, ncols=2)
+    >>> mv.fit(X_train, y_train)
+    >>> mv.score(X_test, y_test)
+    >>> mv.poof()
+    """
+    def __init__(self, visualizers = [], nrows = None, ncols = None, axarr = None, **kwargs):
+        self.visualizers = visualizers
+        self.plotcount = len(visualizers)
```
bbengfort: When we do get around implementing the flipbook visualizer, I intend to make a superclass for both this class and that one that treats the visualizers as an immutable collection, subclassing [`collections.Sequence`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence) -- which requires implementing `__getitem__` and `__len__` on the class; which are pretty straightforward since they simply reference the underlying visualizers list. 

This has two important points for this code:

1. We should store the visualizers as `self._visualizers` then create a visualizers property with no setter (e.g. a clear signal you can't change the visualizers after instantiation). 
2. We should not store `self.plotcount`; as it's only used here, I'd recommend just using a local variable, `plotcount` 

If you'd like to implement this as a sequence straight away, please feel free!

yellowbrick/base.py
```diff
@@ -407,3 +408,119 @@ def predict(self, X, y):
...
+        if nrows == None and ncols == None:
+            self.ncols = 1
+            self.nrows = self.plotcount
```
bbengfort: Perhaps we could check to see if visualizers is a two-dimensional array in case the user specifies a grid?

yellowbrick/base.py
```diff
@@ -407,3 +408,119 @@ def predict(self, X, y):
...
+
+        idx = 0
+        for row in range(self.nrows):
+            for col in range(self.ncols):
+                try:
+                    self.visualizers[idx].ax = self.axarr[row, col]
+                #If len(visualizers) isn't evenly divisibly by rows/columns, 
+                #we want to create the illusion of empty space by hiding the axis
+                except IndexError as e:
+                    self.axarr[row,col].axis('off')
```
bbengfort: Very nice!

yellowbrick/base.py
```diff
@@ -407,3 +408,119 @@ def predict(self, X, y):
...
+
+                idx += 1
+
+        self.kwargs = kwargs
+
+    def fit(self,X,y):
```
bbengfort: We should probably add `**kwargs` to the `fit` method signature just to ensure we've got our basis covered.

yellowbrick/base.py
```diff
@@ -407,3 +408,119 @@ def predict(self, X, y):
...
+
+        for idx in range(len(self.visualizers)):
```
bbengfort: Just out of curiosity, is there a reason not to just loop over `self.visualizers`?

yellowbrick/base.py
```diff
@@ -407,3 +408,119 @@ def predict(self, X, y):
...
+        if self.axarr is None: return
+
+        #Finalize all visualizers
+        for idx in range(len(self.visualizers)):
+            self.visualizers[idx].finalize()
+
+        #Choose a reasonable default size if the user has not manually specified one
+        # self.size() uses pixels rather than matplotlib's default of inches
+        if not hasattr(self, "_size") or self._size is None:
+            self._width = 400 * self.ncols
```
bbengfort: Could we make this `400` a static class var?, e.g. `self.SUBPLOT_DEFAULT_PIXELS` or something like that? (that's a terrible name). Magic numbers like these are likely to trip us up in the future.

yellowbrick/base.py
```diff
@@ -407,3 +408,119 @@ def predict(self, X, y):
...
+        if axarr == None:
+            fig, axarr = plt.subplots(self.nrows, self.ncols, squeeze = False)
+        
+        self.axarr = axarr
+
+        idx = 0
+        for row in range(self.nrows):
+            for col in range(self.ncols):
+                try:
+                    self.visualizers[idx].ax = self.axarr[row, col]
```
bbengfort: I very much like assigning the visualizer ax at init. 

If you wanted (and this is not necessary) you could do `self.visualizers[idx]._ax = self.axarr[row, col]` to bypass the `ax` setter; right now the setter does nothing but set `_ax`, so it would not be functionally different but does give readers of the code a sense of the authority of this class.

yellowbrick/base.py
```diff
@@ -407,3 +408,119 @@ def predict(self, X, y):
...
+            self.nrows = math.ceil(self.plotcount / self.ncols)
+        else:
+            raise YellowbrickValueError("You can only specify either nrows or ncols, \
+                the other will be calculated based on the length of the list of visualizers.")
+        
+
+        if axarr == None:
+            fig, axarr = plt.subplots(self.nrows, self.ncols, squeeze = False)
+        
+        self.axarr = axarr
```
bbengfort: What happens when a user accesses `MultipleVisualzer.ax`? Right now I believe what would happen is that the current axes in the `axarr` would be fetched and stored as `_ax`; however, I believe correct behavior would be akin to `get_current_axes`. I recommend adding the following:

```python 
@property 
def ax(self):
     """
     Override Visualizer.ax to return the current axis 
     """
     return plt.gca() 

@ax.setter
def ax(self, ax):
     raise YellowbrickTypeError("cannot set new axes objects on multiple visualizers")
```

yellowbrick/base.py
```diff
@@ -407,3 +408,119 @@ def predict(self, X, y):
         """
         for model in self.models:
             yield cvp(model, X, y, cv=12)
+
+
+
+#TODO it appears that the previous class (MultiModelMixin) is not actually used
+# anywhere and would be redundant to this new class. Remove that one?
```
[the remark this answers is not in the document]
NealHumphrey: Got it. The first multivisualizer I was planning to make (mentioned above) is in fact (by chance) also comparing multiple estimators. I'll probably implement that one independent of this mixin and then we can look back at it and see what makes sense to hoist up. I'll remove this note.

yellowbrick/base.py
```diff
@@ -407,3 +408,119 @@ def predict(self, X, y):
...
+class MultipleVisualizer(Visualizer):
```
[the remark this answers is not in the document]
NealHumphrey: Definitely open to change. `VisualizerGrid` is a clearer name I think, but you're right it conflicts with gridsearch. We could also directly state what it implements with `VisualizerSubplots`. Other ideas?

tests/test_base.py
```diff
@@ -81,3 +83,46 @@ def test_size_property(self):
...
+
+    def test_draw_with_cols(self):
+        #A simple multiple visualizer that puts two RadViz on two subplots
+        mv = MultipleVisualizer(self.visualizers, ncols=2)
+        mv.fit(self.X,self.y)
+        mv.poof()
+
+    def test_cant_define_both_rows_cols(self):
+        with pytest.raises(YellowbrickValueError) as e:
+            mv = MultipleVisualizer(self.visualizers, ncols=2, nrows=2)
```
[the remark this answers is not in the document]
NealHumphrey: so would that line just be `MultipleVisualizer(self.visualizers, ncols=2, nrows=2)`?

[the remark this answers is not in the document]
bbengfort: It would

yellowbrick/base.py
NealHumphrey: I think it's that when I first wrote this a year ago I wasn't 100% certain that a copy of the visualizer would be created with `for viz in self.visualizers` - this one is clearly operating on the element in place. Rather than double check I just wrote it that way. I can switch to `self.visualizers`

NealHumphrey:
Cool, I can make these changes. Let me know thoughts on a rename. 

Is there any way to run pyflakes locally? I ran the tests with pytest and they passed. Do we have instructions written up for our current testing procedure anywhere?

yellowbrick/base.py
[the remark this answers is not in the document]
NealHumphrey: Of the ideas so far I'd put `VisualizerFrames` at the top though nothing is clearly best pick - grid would be if it weren't for grid search. I think there's still a chance we'll use this directly for external users so maybe not subplots.

bbengfort:
That's strange - if you run `pytest` it should also run pyflakes and coverage; that's defined in the `setup.cfg` file. Did you run the tests in the root directory? Also, make sure the new tests/requirements.txt are installed in your environment; possibly you're running a pytest that's outside the virtualenv? Alternatively, you could try `make test` which runs `python setup.py test` -- which includes a build phase.

yellowbrick/base.py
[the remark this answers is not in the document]
bbengfort: Let's go with `VisualizerGrid` - honestly, I don't think grid search when I hear grid, just when I hear grid search, so I think it'll be fine.

NealHumphrey:
I ran directly on the tests folder so I think I bypassed setup.cfg. Will try it at root level next time

yellowbrick/base.py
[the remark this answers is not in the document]
NealHumphrey: Done

tests/test_base.py
```diff
@@ -81,3 +83,46 @@ def test_size_property(self):
...
+        mv.poof()
+
+    def test_draw_with_cols(self):
+        #A simple multiple visualizer that puts two RadViz on two subplots
+        mv = MultipleVisualizer(self.visualizers, ncols=2)
+        mv.fit(self.X,self.y)
+        mv.poof()
+
+    def test_cant_define_both_rows_cols(self):
+        with pytest.raises(YellowbrickValueError) as e:
```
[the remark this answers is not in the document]
NealHumphrey: done

yellowbrick/base.py
NealHumphrey: I've switched this to an @property. Instead of not creating a setter, I created one that just raises an AttributeError - not sure if that's best practice but seemed even more clear by raising the error if the user tries to edit the visualizers list.

yellowbrick/base.py
NealHumphrey: That would be a good enhancement. For now I've just put a TODO in the code as a reminder and we can see how we end up using this class.

yellowbrick/base.py
NealHumphrey: done

yellowbrick/base.py
NealHumphrey: changed

yellowbrick/base.py
NealHumphrey: done. I used the name.

yellowbrick/base.py
NealHumphrey: I'll skip for now - I'd be worried about future updates to the base .ax property wanting to do more than set _ax.

yellowbrick/base.py
NealHumphrey: Done - I didn't test previous behavior, but adding this seems like it would override any such behavior and does what we want it to whether or not previous behavior was problematic.

NealHumphrey:
@bbengfort I've addressed all your comments. I fixed the two items you mentioned that caused failing CI tests; but, I still couldn't run the tests locally. I summarized the issue with test running in the overhaul unittest issue #318. Hopefully those two fixes will let the CI tests pass this time - I ran just the test files (i.e. without pyflakes or coverage, which was not working for me) locally and those worked.

yellowbrick/base.py
bbengfort: If you leave the setter off and try to assign a value to the property an `AttributeError: can't set attribute` is raised. Your exception is equivalent and more informative, so I'm happy to go with that, though I think it's more typical to simply leave the setter off. 

```python
class Thing(object):

    def __init__(self, x):
        self._x = x 

    @property
    def x(self):
        return self._x

>>> t = Thing(42)
>>> t.x = 'bob'
Traceback (most recent call last):
  File "t.py", line 13, in <module>
    t.x = 'bob'
AttributeError: can't set attribute
```

bbengfort:
@NealHumphrey I just merged from develop to see if some of the fixes we made to the tests help resolve the tests for this PR.

bbengfort:
@NealHumphrey I'm going to go ahead and merge this and see if I can fix the tests.

PR #185, opened 2017-03-29
added jointplot visualizer

pdamodaran:


rebeccabilbro:
@pdamodaran as discussed, let's add a Matplotlib version check at the beginning to let the user know that this visualizer requires 2.0.0

PR #192, opened 2017-04-10
Added tests for JointPlotVisualizer

pdamodaran:


tests/test_features/test_jointplot.py
```diff
@@ -0,0 +1,65 @@
+# tests.test_features.test_jointplot
+# Test the JointPlotVisualizer
+#
+# Author:   Prema Damodaran Roman
+# Created:
```
bbengfort: For this line, you can go ahead and put in a  timestamp formatted like as follows: Mon Apr 10 08:19:54 2017 -0400.

tests/test_features/test_jointplot.py
```diff
@@ -0,0 +1,65 @@
+# tests.test_features.test_jointplot
+# Test the JointPlotVisualizer
+#
+# Author:   Prema Damodaran Roman
+# Created:  
+#
+# Copyright (C) 2017 District Data Labs
+# For license information, see LICENSE.txt
+#
+# ID: test_jointplot.py
```
bbengfort: The ID line is also special and is automatically filled in with the revision when we merge to master. It looks something like this:

```
# ID: setup.py [] benjamin@bengfort.com $
```

You can just put your GitHub username instead of your email if you'd prefer; but technically that's the email associated with the commit that created the document.

tests/test_features/test_jointplot.py
```diff
@@ -0,0 +1,65 @@
...
+from tests.dataset import DatasetMixin
+from yellowbrick.features.jointplot import *
+
+##########################################################################
+## JointPlotVisualizer Tests
+##########################################################################
+
+class JointPlotTests(unittest.TestCase, DatasetMixin):
+
+	X = np.array([1, 2, 3, 5, 8, 10])
```
bbengfort: Potentially you could update this to a 2 dimensional array and then test all the X[i], y combinations since this is closer to what a user would have to do with this plot. You could also stress the joint plot a bit by using np.rand and having slightly larger arrays -- this might catch edge cases and errors that you didn't expect.

tests/test_features/test_jointplot.py
```diff
@@ -0,0 +1,65 @@
...
+	def tearDown(self):
+		self.concrete = None
+
+	def test_jointplot(self):
+		"""
+		Assert no errors occur during jointplot visualizer integration
+		"""
+		
+		visualizer = JointPlotVisualizer()
+		visualizer.fit(self.X, self.y)
```
bbengfort: I would also call `visualizer.poof()` under this line like you're doing in the text test.

pdamodaran:
I have updated the test code per the review.  The only modification remaining is to attempt to run the code on multi-dimensional array testing all X[i] cases.

bbengfort:
@pdamodaran great! You can merge when ready by hitting the squash and merge button. Thanks for all your hard work on this visualizer!

Issue #194, opened 2017-04-11
Add FeatureImportanceVisualizer for Treebased-Models


tsterbak:
When you use tree-based models, a common approach is to investigate the feature importance induced by the splits of the trees. 

* Make a barplot to show the feature importance provided by sklearn.
* Option for normalization of values
* Featurename parameter

I will start implementing this today and PR. Any opinions about this?

bbengfort:
@tsterbak we have some folks working on feature importance with linear models using coefficients right now, but I'm not sure how far along they are, so my opinion is that you should go for it, and we'd be happy to have a PR. Looking forward to seeing what you put together. 

Let me know if you have any questions. We're in the process of writing a better contributor guide, check out what we have so far in the develop branch.

rebeccabilbro:
@tsterbak - sounds great! Here's [some info](http://www.scikit-yb.org/en/develop/contributing.html#visualizer-api) we put together on rolling your own yb visualizer, in case it's helpful!

tsterbak:
I opened a pull request and added a basic implementation of what I was thinking of.
Maybe we should abstract this more and build a baseImportanceVisualizer and subVisualizers for different types of models, or make visualization for a bigger class of models possible with the ImportanceVisualizer.

bbengfort:
Looks good so far. 

In terms of the hierarchy, I'd say that this was a `ModelVisualizer` potentially we could create an extension of that called `ImportanceVisualizer`, and then have your model subclass that. 

In terms of checking if the model is a forest model, we've certainly been using `is_classifier()` and the like in the utils.py module, since Scikit-Learn also does stuff like that. See how we use that in the `ClassifierScoreVisualizer` base class: [classifier.py#L49](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier.py#L49). Not sure how you would come up with an is_forest method though.

bbengfort:
See #58

bbengfort:
Copied from @ianozsvald on #195  for reference as we develop this further. 

At the risk of adding noise - I'll note that commonly I create the following code to look for the most important features in a RandomForest:
![image](https://user-images.githubusercontent.com/273210/27447491-0a15df9c-5779-11e7-95fa-03677f614e97.png)

```
df_feature_importances = pd.DataFrame(list(zip(clf.feature_importances_, X.columns)), columns=['importance', 'feature'])
[...]

PR #195, opened 2017-04-11
added ImportanceVisualizer for tree-based models

tsterbak:
This PR follows #194 and adds:
* basic implementation of the tree-based feature importance as shown in sklearn examples

TODO
- [ ] create an example and visualize the output
- [ ] make sure only used for tree-based models (RandomForest, ExtraTrees, GradientBoosting, DecisionTree, ...)
- [ ] unit tests
- [x] is the Visualizer in the right place?
- [ ] all module rules satisfied
- [ ] proper inheritance


bbengfort:
@tsterbak would you mind posting an example and the resulting visualization in the comment thread? 

As for the visualizer being in the right place, it absolutely is. However, we're planning on refactoring the classifier module into a package (similar to regressor and cluster); so it would actually make it easier if you could put your visualizer in a file called tree.py that is a sibling to classifier.py, and then I'll move it into the classifier directory when we refactor (which will probably occur right after we merge your PR). That way you can also write your tests in test_tree.py. 

I'll address the class hierarchy in the other issue since you brought it up there. 

Looks good so far!

ianozsvald:
At the risk of adding noise - I'll note that commonly I create the following code to look for the most important features in a RandomForest:
image

```
df_feature_importances = pd.DataFrame(list(zip(clf.feature_importances_, X.columns)), columns=['importance', 'feature'])
df_feature_importances = df_feature_importances.sort_values('importance', ascending=False).set_index('feature')
nbr_features = 20

fig, ax = plt.subplots(figsize=(8,6)); 
df_feature_importances[:nbr_features].plot(kind="barh", title="Feature importances", legend=None, ax=ax)
plt.gca().invert_yaxis()
plt.xlabel("Importance")

# remove right/top border to make things slightly neater
ax.spines['right'].set_color('none')
ax.spines['top'].set_color('none')

# visual tidy-up to make left axis small values slightly easier to read
# offset left and bottom axis
ax.spines['bottom'].set_position(('axes', -0.05))
ax.yaxis.set_ticks_position('left')
ax.spines['left'].set_position(('axes', -0.05))
```

In #194 the sklearn visualisation was noted http://scikit-learn.org/stable/auto_examples/ensemble/plot_forest_importances.html and this adds a standard deviation to each bar. A measure of spread sounds very useful - I'm not sure if assuming a _Gaussian_ is right (I only have a weak opinion on this). Can we guarantee a Gaussian distribution for feature importances? Might it be skewed? Maybe a boxplot per feature is actually a better thing (as it is non-parametric and really quick)? I'm wondering if occasional outliers might skew the distribution and make it non-Normal?

Looking at a problem I'm working on right now, with a RandomForest with 58 estimators, extracting the most important feature (X314 in the picture above) and plotting each tree's feature importance for this feature I get:
image

but if I take the second most important feature then the mean and standard distribution would paint the wrong summarised picture (as a set of my estimators think this feature has very low value indeed):
image

```
vals = pd.Series([tree.feature_importances_[288] for tree in clf.estimators_])
vals.plot(kind="hist", title="Mean importance for feature 288 'X315' == {:0.2f}".format(vals.mean()))
```

I also wonder if some additional tree information (e.g. n_estimators, max_depth) might be a useful addition to the plot, perhaps in the title?

bbengfort:
@tsterbak I'd like to merge this as an undocumented feature, then open up an issue to take this to an end, let me know if you have any thoughts.

bbengfort:
See #58

tsterbak:
@bbengfort I guess this makes sense. I would like to finish it since it is laying around for nearly a year now... I'm not sure what needs to be done. Maybe an issue would be the right fit.

bbengfort:
@tsterbak I've actually quickly put this together with tests and documentation and fitting in with the new structure; see #317 -- this is not a tree specific decorator but could use tree specific things like the standard deviation or percentiles that @ianozsvald mentioned and things like depth or n_estimators. I'll post this in the #194 issue that you started and perhaps you could subclass this and take it to the end.

PR #196, opened 2017-04-12
Decision Boundary Visualizer

ndanielsen:
Update: 

The DecisionBoundaryVisualizer visualizes the model decisions given a two feature space. Next iteration on this will include ability to select one class along with score metrics with selected class.

This visualizer supports numpy structured arrays, dataframes and normal ndarrays. 

test_real_data_set_viz


bbengfort:
@ndanielsen excellent! Let me know when you would like me to take a look.

ndanielsen:
@balavenkatesan I made some changes and added test coverage. Feel free to take a look before I finalize and put it in for a review in the next day or so

yellowbrick/boundaries.py
```diff
@@ -0,0 +1,286 @@
...
+        process, but can and should be set as early as possible.
+
+        These parameters can be influenced later on in the visualization
+        process, but can and should be set as early as possible.
+        """
+        super(DecisionBoundariesVisualizer, self).__init__(self)
+
+        self.colors = colors
+        self.classes_ = classes
+        self.features_ = features
```
ndanielsen: need to fix the doc string

yellowbrick/boundaries.py
```diff
@@ -0,0 +1,286 @@
...
+
+class DecisionBoundariesVisualizer(ModelVisualizer):
+    """
+    DecisionBoundariesVisualizer is a bivariate data visualization algorithm
+    that plots the decision boundaries of each class.
+    """
+
+    def __init__(self,
+                 model,
+                 colors=PALETTES['set1'],
```
ndanielsen: add x,y to init, and logic for either x,y or features

yellowbrick/boundaries.py
```diff
@@ -0,0 +1,286 @@
...
+from matplotlib.patches import Patch
+
+from yellowbrick.exceptions import YellowbrickTypeError
+from yellowbrick.base import ModelVisualizer
+from yellowbrick.utils import get_model_name
+from yellowbrick.style.colors import resolve_colors
+from yellowbrick.utils import is_dataframe
+from yellowbrick.style.palettes import PALETTES
+
+
```
ndanielsen: add quick method

yellowbrick/boundaries.py
```diff
@@ -0,0 +1,286 @@
...
+            )
+        # Handle the feature names if they're None.
+        if self.features_ is None:
+            # If X is a data frame, get the columns off it.
+            if is_dataframe(X):
+                self.features_ = X.columns
+
+            # Otherwise create numeric labels for each column.
+            else:
+                self.features_ = [str(cdx) for cdx in range(ncols)]
```
ndanielsen: change to np.unique

yellowbrick/boundaries.py
```diff
@@ -0,0 +1,286 @@
...
+                for kls_num, label in enumerate(self.classes_)
+            }
+            self.class_labels = dict(zip(set(y), self.classes_))
+        else:
+            raise YellowbrickTypeError(
+                """Number of classes must be the same length of number of
+                target y"""
+            )
+        # Handle the feature names if they're None.
+        if self.features_ is None:
```
ndanielsen: add in feature selection logic from ScatterViz

ndanielsen:
@bbengfort @NealHumphrey ready for a review when you've got time

tests/test_classifier/test_boundaries.py
```diff
@@ -0,0 +1,337 @@
+# tests.test_boundaries
+# Ensure that the decision boundary visualizations work.
+#
+# Author:   Author:   Nathan Danielsen <nathan.danielsen@gmail.com>
+# Created:  Sun Mar 19 13:01:29 2017 -0400
+#
+# Copyright (C) 2017 District Data Labs
+# For license information, see LICENSE.txt
+#
+# ID: test_knn.py [] nathan.danielsen@gmail.com $
```
bbengfort: test_boundaries.py || test_knn.py? 

A note on these ID tags, I have a script that detects them and fills in the brackets with the commit to master that included the original change. This is part of a workflow for a previous place I worked at, but eventually we may want to study our revisions and user growth.

tests/test_classifier/test_boundaries.py
```diff
@@ -0,0 +1,337 @@
...
+from yellowbrick.exceptions import YellowbrickTypeError
+from yellowbrick.exceptions import YellowbrickValueError
+
+from sklearn import datasets
+from sklearn import neighbors
+from sklearn import naive_bayes
+##########################################################################
+# Data
+##########################################################################
+# yapf: disable
```
bbengfort: what's yapf?

tests/test_classifier/test_boundaries.py
```diff
@@ -0,0 +1,337 @@
...
+##########################################################################
+# Residuals Plots test case
+##########################################################################
+
+
+class DecisionBoundariesVisualizerTest(VisualTestCase):
+    """Testcases for the DecisionBoundariesVisualizers """
+
+    def test_decision_bounardies(self):
+        """Assert no errors occur during KnnDecisionBoundariesVisualizer
```
bbengfort: With the -v flag on nosetests set, the first line of the docstring is used in the test description; this isn't a big deal, but just wanted to note that the word "integration" would be chopped off during tests. I think it's pretty readable even so.

yellowbrick/classifier/boundaries.py
```diff
@@ -0,0 +1,457 @@
...
+
+class DecisionBoundariesVisualizer(ModelVisualizer):
+    """
+    DecisionBoundariesVisualizer is a bivariate data visualization algorithm
+    that plots the decision boundaries of each class.
+
+    Parameters
+    ----------
+
+        model : the Scikit-Learn estimator
```
bbengfort: not sure if the indent here will cause problems with the docstring build (will check after merge; unless you want to take a quick build of the docs and see what happens).

yellowbrick/classifier/boundaries.py
```diff
@@ -0,0 +1,457 @@
...
+        x : string, default: None
+            The feature name that corresponds to a column name or index postion
+            in the matrix that will be plotted against the x-axis
+
+        y : string, default: None
+            The feature name that corresponds to a column name or index postion
+            in the matrix that will be plotted against the y-axis
+
+        colors : string or matplotlib cmap, default: yellowbrick palette 'set1'
+            Sets the color palette of the visualizer
```
bbengfort: These are the colors for the classes right? Can it also be a list of colors?

yellowbrick/classifier/boundaries.py
```diff
@@ -0,0 +1,457 @@
...
+        self.features_ = features
+        self.estimator = model
+        self.name = get_model_name(self.estimator)
+        self.show_scatter = show_scatter
+        self.step_size = step_size
+        self.markers = itertools.cycle(
+            kwargs.pop('markers', (',', 'o', 'd', '*', 'v', 'h', '+')))
+        self.pcolormesh_alpha = pcolormesh_alpha
+        self.scatter_alpha = scatter_alpha
+        self.title = title
```
bbengfort: Note that title is handled in the super class, actually in the very base class Visualizer, and should be omitted here to ensure that `set_title()` works.

yellowbrick/classifier/boundaries.py
```diff
@@ -0,0 +1,457 @@
...
+    visualizer.fit_draw_poof(X, y, **kwargs)
+
+    # Return the axes object on the visualizer
+    return visualizer.ax
+
+##########################################################################
+# Static ScatterVisualizer Visualizer
+##########################################################################
+
+class DecisionBoundariesVisualizer(ModelVisualizer):
```
bbengfort: Why doesn't this extend from `ClassificationVisualizer`?

yellowbrick/classifier/boundaries.py
```diff
@@ -0,0 +1,457 @@
...
+        """
+        super(DecisionBoundariesVisualizer, self).__init__(self)
+
+        self.x = x
+        self.y = y
+        self.colors = colors if colors is not None else PALETTES['set1']
+        self.classes_ = classes
+        self.features_ = features
+        self.estimator = model
+        self.name = get_model_name(self.estimator)
```
bbengfort: self.name and self.estimator (and self.classes_) are set in `ClassificationScoreVisualizer` -- I guess this is not a score visualizer, but do we want to share this functionality somehow?

yellowbrick/classifier/boundaries.py
```diff
@@ -0,0 +1,457 @@
...
+    """
+    DecisionBoundariesVisualizer is a bivariate data visualization algorithm
+    that plots the decision boundaries of each class.
+
+    Parameters
+    ----------
+
+        model : the Scikit-Learn estimator
+            Should be an instance of a classifier, else the __init__ will
+            return an error.
```
bbengfort: So the ClassificationVisualizer does raise an exception if it's not a classifier; but it's an empty threat here, because as far as I can tell the ModelVisualizer won't raise an exception ;-)

yellowbrick/classifier/boundaries.py
```diff
@@ -0,0 +1,457 @@
...
+        if features is not None:
+            if len(features) != 2:
+                raise YellowbrickValueError(
+                    'DecisionBoundariesVisualizer only accepts two features.')
+
+    def _select_feature_columns(self, X):
+        """ """
+
+        if len(X.shape) == 1:
+            X_flat = X.view(np.float64).reshape(len(X), -1)
```
bbengfort: X.view has been giving me some trouble; apparently a future version of numpy is going to change it's behavior. Are you getting any warnings when using this? If not, would you mind `pip install -U numpy` and seeing if you get any warnings after? I've been trying to figure it out, but no luck yet. It may just have to do with named arrays.

yellowbrick/classifier/boundaries.py
```diff
@@ -0,0 +1,457 @@
...
+                target y"""
+            )
+
+        # ensure that only
+        self.estimator.fit(X, y)
+
+        # Plot the decision boundary. For that, we will assign a color to each
+        # point in the mesh [x_min, x_max]x[y_min, y_max].
+        x_min, x_max = X[:, 0].min() - 1, X[:, 0].max() + 1
+        y_min, y_max = X[:, 1].min() - 1, X[:, 1].max() + 1
```
bbengfort: So I get that you're trying to add some padding with the -1 and +1 on the min and the max. The problem is that if the range of feature is very small e.g. between 0.01 - 0.25, then adding or subtracting 1 here is going to result in a massive imbalance in the mesh. Also, though less bad, if the range of the feature is 10 to 1000, then the adding and subtracting of one will be barely visible. 

Potentially it might make sense to normalize X to the range [0, 1] -- this would solve the first problem. However, I might have a different suggestion depending on why you adjusted the range (perhaps the mesh should just go from min to max or be scaled by * 1.01 or something like that). Let me know!

yellowbrick/classifier/boundaries.py
```diff
@@ -0,0 +1,457 @@
...
+
+        # Create the axes if they don't exist
+        if self.ax is None:
+            self.ax = plt.gca(xlim=[x_min, x_max], ylim=[y_min, y_max])
```
bbengfort: doing a check on self.ax is no longer necessary; and although you're setting limits using gca, I'm not sure that will matter since your mesh is bounded by the range already; what do you think?

yellowbrick/classifier/boundaries.py
```diff
@@ -0,0 +1,457 @@
...
+        Called from the fit method, this method creates a decision boundary
+        plot, and if self.scatter is True, it will scatter plot that draws
+        each instance as a class or target colored point, whose location
+        is determined by the feature data set.
+        """
+        # ensure that if someone is passing in another X such as X_test, that
+        # features will be properly handled
+        X = self._select_feature_columns(X)
+
+        num_colors = len(self.classes_) * 2
```
bbengfort: Is this one color for the scatter point and one for the background for each class?

yellowbrick/classifier/boundaries.py
```diff
@@ -0,0 +1,457 @@
...
+
+        Parameters
+        ----------
+        kwargs: generic keyword arguments.
+
+        """
+        # Divide out the two features
+        feature_one, feature_two = self.features_
+
+        if self.title is None:
```
bbengfort: This is also handled by the base class `set_title` function; you should feel free to just pass your title into set_title.

yellowbrick/classifier/boundaries.py
ndanielsen: okay, making the change

yellowbrick/classifier/boundaries.py
ndanielsen: yeah, updated numpy and getting these warnings. Not sure about a workaround for this. I'll keep an eye out

PR #207, opened 2017-04-27
Threshold visualizer

ndanielsen:
This visualizer creates a visualizations from n_trials that plots thresholds against model variability percentages of precision, recall and queue for binary labeled data.

Example viz here is from the UCI spam classification dataset 
thresholdviz

I had flagged this blog post awhile ago, and mostly reimplemented in in yb: 
http://blog.insightdatalabs.com/visualizing-classifier-thresholds/


bbengfort:
@ndanielsen let me know when you're ready for me to review.

ndanielsen:
@bbengfort @NealHumphrey ready for a review

tests/test_classifier/test_rocauc.py
```diff
@@ -167,7 +167,7 @@ def test_rocauc_quickmethod(self):
         data = load_breast_cancer()
         model = DecisionTreeClassifier()
 
-        # TODO: impage comparison of the quick method 
+        # TODO: impage comparison of the quick method
```
bbengfort: imp-rage-b-color

I read this "imp rage" -- Google happily supplied me with this.

yellowbrick/classifier/threshold.py
```diff
@@ -0,0 +1,351 @@
...
+
+    colormap : string or cmap, default: None
+        Optional string or matplotlib cmap to colorize lines
+        Use either color to colorize the lines on a per class basis or
+        colormap to color them on a continuous scale.
+
+    color : String, default: None
+        Optional string or matplotlib cmap to colorize lines
+        Use either color to colorize the lines on a per class basis or
+        colormap to color them on a continuous scale.
```
bbengfort: What's the relationship between colormap and color? Should we specify it in the documentation? 

Also minor nit: recommend lower case `String`, `Sequence` and `Integer` --> `int` just for consistency with other docstrings. Here they make me think of a specific class rather than the primitive types or interfaces.

yellowbrick/classifier/threshold.py
```diff
@@ -0,0 +1,351 @@
...
+        if not isclassifier(model):
+            raise YellowbrickTypeError(
+                "This estimator is not a classifier; try a regression or clustering score visualizer instead!"
+            )
+        super(ThresholdVisualizer, self).__init__(model, **kwargs)
+
+        self.estimator = model
+        self.colormap = kwargs.pop('colormap', None)
+        self.color = kwargs.pop('color', None)
+        self.title = title
```
bbengfort: Note that the super class, `Visualizer` handles title and implements `set_title()`. [yellowbrick.base L71](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/base.py#L71)

yellowbrick/classifier/threshold.py
```diff
@@ -0,0 +1,351 @@
...
+    return visualizer.ax
+
+
+##########################################################################
+# Static ThresholdVisualizer Visualizer
+##########################################################################
+
+
+class ThresholdVisualizer(ModelVisualizer):
+    """Visualizes precision, recall and queue rate at different different
```
bbengfort: This description should probably be expanded a bit, as well as a `.. seealso::` with the link to the blog post you mentioned in the spec. At the very least I'm not sure what the different thresholds are, and the specificity of a binary classifier should be firmly established.

yellowbrick/classifier/threshold.py
```diff
@@ -0,0 +1,351 @@
...
+            Returns the instance of the visualizer
+        """
+        self.plot_data = []
+
+        for trial in range(self.n_trials):
+            train_X, test_X, train_y, test_y = train_test_split(
+                X,
+                y,
+                test_size=self.test_size_percent,
+                random_state=self.random_state)
```
bbengfort: If the random_state is always 0 (the default in the kwargs) won't this always return the same X and y? I just ran it in my terminal and that seemed to be the result.

yellowbrick/classifier/threshold.py
```diff
@@ -0,0 +1,351 @@
...
+            color_values = resolve_colors(
+                num_colors=3, colormap=self.colormap, color=self.color)
+        else:
+            color_values = get_color_cycle()
+
+        # unpack first three colors for visualization
+        recall_color, precision_color, queue_rate_color = color_values[:3]
+
+        if self.ax is None:
+            self.ax = plt.gca(xlim=[-1, 1], ylim=[-1, 1])
```
bbengfort: This is no longer required with the new `self.ax` property in the descriptor, which automatically gets the correct axis. In fact, self.ax will never be None; so this will never get called! To ensure the limits are set correctly, set them in `finalize()`.

yellowbrick/classifier/threshold.py
```diff
@@ -0,0 +1,351 @@
...
+            uniform_precision_plots.append(uniform_precision)
+            uniform_recall_plots.append(uniform_recall)
+            uniform_queue_rate_plots.append(uniform_queue_rate)
+
+        lower_precision, median_precision, upper_precision = mstats.mquantiles(
+            uniform_precision_plots, prob=self.quantiles, axis=0)
+        lower_recall, median_recall, upper_recall = mstats.mquantiles(
+            uniform_recall_plots, prob=self.quantiles, axis=0)
+        lower_queue_rate, median_queue_rate, upper_queue_rate = mstats.mquantiles(
+            uniform_queue_rate_plots, prob=self.quantiles, axis=0)
```
bbengfort: My feeling is that most the above can be moved to its own function, since none of this data relies on the axes or drawing objects. Refactoring this might make it easier to debug. Also; would the user be interested in this data? If so, they could access that function directly without drawing or we could store this data after `fit()` rather than the dictionaries in plot data -- depends on what you think would be more valuable.

yellowbrick/classifier/threshold.py
```diff
@@ -0,0 +1,351 @@
...
+            color=recall_color)
+        self.ax.fill_between(
+            uniform_thresholds,
+            upper_queue_rate,
+            lower_queue_rate,
+            alpha=0.5,
+            linewidth=0,
+            color=queue_rate_color)
+
+        self.ax.axis('auto')
```
bbengfort: Doesn't this beat up on the [-1,1] lims set before?

yellowbrick/classifier/threshold.py
```diff
@@ -0,0 +1,351 @@
...
+            alpha=0.5,
+            linewidth=0,
+            color=recall_color)
+        self.ax.fill_between(
+            uniform_thresholds,
+            upper_queue_rate,
+            lower_queue_rate,
+            alpha=0.5,
+            linewidth=0,
+            color=queue_rate_color)
```
bbengfort: Might I suggest this tiny simplification, that might be aided by the refactoring of the computation of thresholds above? 

```python 
uplots = (uniform_precision_plots, uniform_recall_plots, uniform_queue_rate_plots) 

# Compute the curves for precision, recall, and queue rate and plot them 
for uniform_plot, color in zip(uplots, color_values): 
    # Compute the lower, median, and upper plots 
    lower, median, upper = mstats.mquantiles(uniform_plot, prob=self.quantiles, axis=0)
 
    # Draw the median line 
    self.ax.plot(uniform_thresholds, median, color=color)

    # Draw the fill between the lower and upper bounds 
    self.ax.fill_between(uniform_thresholds, upper, lower, alpha=0.5, linewidth=0, color=color) 
```

May make it a bit easier to understand and make changes to.

yellowbrick/classifier/threshold.py
```diff
@@ -0,0 +1,351 @@
...
+        The user calls poof and poof calls finalize.
+
+        Parameters
+        ----------
+        kwargs: generic keyword arguments.
+
+        """
+        # Set the title
+        if self.title is not None:
+            self.set_title(self.title)
```
bbengfort: Title handled by super class if you want to call super on final. Also recommend you give a generic title in case one doesn't exist like "Threshold Plot of Binary Classifier" or something like that.

yellowbrick/classifier/threshold.py
ndanielsen: @bbengfort recommend that we better document colormap and color usage. I'm still unclear on the intended use for both.

ndanielsen: I've been putting sticking both of them in resolve_colors(), but perhaps the usage and pattern should be pulled up into a higher class to simplify implementation?

bbengfort: So I think color vs colormap is a hold over from Pandas (and not the current version) where either a color or colormap could be passed in. This made a lot of sense for feature visualizers where the target could either be discrete (a classifier or clusterer) and therefore required a color cycle or discrete list of colors or it could be continuous (a regression) and a colormap was required. 

So here are Ben's (possibly erroneous) notes on color and `resolve_colors()` as I can think of them off the top of my head:

1. `color` refers to a discrete sequence of colors 
    1. color can be a string, a tuple, a float, or a list of strings, list of floats or a list of tuples 
    2. in most cases, it should be a list of strings or a list of tuples 
    3. a float indicates a grayscale value 
    4. a tuple indicates an RGB/A  value 
    5. a string can be either a color name or reference, e.g. `bgrmcyk` 
    6. in the case of lists, a string can also be the name of a palette, e.g. `"Set1"`
2. `colormap` refers to a continuous sequence of colors 
    1. colormap can be a string or a collection 
    2. a string is the name of a colormap, e.g. `"RdBu"`) 
    3. a string with an `_r` at the end reverses the colormap, e.g. `"RdBu_r"`. 
    4. a string can be the name of a palette, which is converted to a `ListedColorMap` 
    4. a collection of colors is converted into a `ListedColorMap` 
3. Because we need to get a distinct n - # of colors two things happen:
    1. We get a slice of discrete colors passed to `color` or a cycle for n > len(colors), repeating colors as we go. 
    2. For a colormap, n determines the exact color values in the heatmap
4. With a colormap, we need to add a `colorbar`, with `color` we need a legend. 

Ok, so that's a start of my thoughts - is this jibing more or less?

bbengfort: also, we should get this in the docs somewhere.

bbengfort:
@ndanielsen Will you be around 7/5? I know it's right after a holiday but we're having an office hours at Georgetown and perhaps we can take your two PRs across the finish line together?

ndanielsen:
@bbengfort won't be able to make the 5th as I'm be out of DC on vacation. I made all the changes on the DecisionBoundaryViz and merged it in. I just haven't had much time to finish up the ThresholdViz, it seems like a small amount of time to finish it up

bbengfort:
No worries, that's a good time to take a vacation out of DC! Glad to hear things are progressing, just wanted to offer my help.

yellowbrick/classifier/threshold.py
ndanielsen: okay, set the default to None (no set random state unless user defines it in init)

ndanielsen:
@bbengfort I added in documentation for this one. Take a look when you get the chance

PR #208, opened 2017-04-27
Ready: 151 refactor classifier

### LINKED ISSUE #151: Refactor classifier.py into module
As classifier has grown, should refactor it into a folder with individual .py files for the classes. Should also look into hoisting functionality to create parent classes as necessary.

NealHumphrey:
This refactor of the classifier module resolves #151. It should be close, and can be merged as is, though I'd like to fix a few more things in it. Making the pull request now so you can see where it stands. 

Note, this adds the enhancement of using the find_text_color util for the ClassificationReport. It also includes a partial fix of #141 - the code no longer errors out and a report is displayed, but it has weird dimensions (half the screen area isn't used).

Still to do:
1) @bbengfort in your comment in 151 you said "don't forget to update the documentation autodoc functionality." I'm not sure what that refers to?
2) I'd also like to do a quick pass at cleaning up the docstrings while I'm here, plus maybe do a bit of mild inheritance adjustment to hoist some stuff up to the ClassificationScoreVisualizer, but if I don't get to it I'll add a a new issue. 
3) Hoping to do complete fix of #141 by converting it to a mesh instead of imshow, the way I did for the ConfusionMatrix.


tests/test_classifier/__init__.py
```diff
@@ -0,0 +1,6 @@
+#Backend must be set before first use. 
+# Setting backend here allows us to run tests just in this folder, without running the whole yellowbrick.tests folder
+# This command will have no effect if backend has already been set previously. 
+import matplotlib
+matplotlib.use('Agg')
+
```
bbengfort: Did the [VisualTestCase](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/base.py#L28) not work to ensure that the backend was set correctly? It looks like that's the base class of all the tests, and it was supposed to set the backend on setUp.

tests/test_classifier_old.py
```diff
@@ -18,7 +18,7 @@
 import unittest
 
 from tests.base import VisualTestCase
-from yellowbrick.classifier import *
+from yellowbrick.classifier_old import *
```
bbengfort: What is `classifier_old`?

bbengfort: So it looks like it's just the original file; will you remove this before merging it in?

NealHumphrey: Correct, will remove before merging.

tests/test_classifier/__init__.py
NealHumphrey: When I tried to run only this folder of tests, it didn't let me because it didn't have proper backend. If you run the whole tests folder I think it is OK. If you include this line in the subfolder, it doesn't affect running the whole tests folder because it ignores later requests to set the back end. Let me know if you want me to remove, though.

bbengfort: No it's fine, no need to remove it - but if you could add an issue for me to investigate why the base test case is not working, I'd appreciate it.

tests/test_classifier_old.py
bbengfort: Thanks, merge when you're ready.

NealHumphrey:
Fixed documentation. Will merge as soon as tests pass

PR #222, opened 2017-05-15
Update CONTRIBUTING.md

ndanielsen:
Just made a minor formatting fix and added how someone can run their own individual test file


bbengfort:
@ndanielsen ok this does not have @rebeccabilbro's changes in it - would you mind adding your changes to the new version of the contrib documentation and we'll try again?

rebeccabilbro:
Just double checked and it looks like this actually does have my changes, though I did add to the docs again yesterday so I'm attempting to update now.

PR #232, opened 2017-05-22
WIP pca_feature_visualization

cjmorale:
I would like to help out with the PCA Feature Visualization. I worked on some initial code today, below is a plot the code creates using the iris data set. Would love to get some feed back.

```
    from yellowbrick.features.pca import PCA2D
    from sklearn import datasets
    iris = datasets.load_iris()
    X = iris.data
    y = iris.target

    params = {'scale': True, 'center': True, 'col': y}
    visualizer = PCA2D(**params)
    visualizer.fit(X)
    visualizer.transform(X)
    visualizer.poof()
```
iris_pca_plot

Current features
   - User can choose to center and scale data

Needs to be completed
   - documentation
   - test cases


bbengfort:
@cjmorale awesome, thank you! Are you still going to be around for the sprints tomorrow (@rebeccabilbro mentioned she spoke to you at PyCon); if so we can do the code review together; otherwise I'll plan to do it tomorrow morning.

cjmorale:
@bbengfort I will be leaving for the airport around 9, so if you are an early riser we could do the code review together.

cjmorale:
Committed some new code. There is now an ipynb example if the folder `examples/cjmorale`

rebeccabilbro:
@cjmorale - excellent, taking a look now!

yellowbrick/features/pca.py
```diff
@@ -0,0 +1,103 @@
...
+
+from yellowbrick.features.base import DataVisualizer
+
+from sklearn.pipeline import Pipeline
+from sklearn.decomposition import PCA
+from sklearn.preprocessing import StandardScaler
+##########################################################################
+##2D PCA Visualizer
+##########################################################################
+class PCA2D(DataVisualizer):
```
bbengfort: This is _exactly_ what we hoped to see from new visualizers - good naming, good documentation/docstrings, tests; it's the whole package!

tests/test_features/test_pca.py
```diff
@@ -0,0 +1,71 @@
...
+from tests.dataset import DatasetMixin
+from yellowbrick.features.pca import PCA2D
+
+##########################################################################
+## RadViz Base Tests
+##########################################################################
+
+class PCA2DTests(unittest.TestCase, DatasetMixin):
+
+    def test_scale_t_center_t(self, scale=True, center=True):
```
bbengfort: Our test package requires a one line description of the tests, e.g. 

```python
def test_scale_t_center_t(self):
    """
    Test the PCA2D visualizer scaled and centered 
    """
```

Because this line in the docstring is printed in the test report. 

Also, in terms of arguments to the test, e.g. (scale=True, center=True), our test runner won't be able to modify the arguments, so they're there for documentation only, and I think the name of the method does a good job of documenting it if you want to remove them. Otherwise you could create a function that takes the arguments that is not named `test_` and then simply pass the arguments to the function from your two test cases.

yellowbrick/features/pca.py
```diff
@@ -0,0 +1,103 @@
...
+    >>> X = iris.data
+    >>> y = iris.target
+    >>> params = {'scale': True, 'center': False, 'col': y}
+    >>> visualizer = PCA2D(**params)
+    >>> visualizer.fit(X)
+    >>> visualizer.transform(X)
+    >>> visualizer.poof()
+
+
+    Notes
```
bbengfort: Not necessary to include a Notes header if you don't have any!

yellowbrick/features/pca.py
```diff
@@ -0,0 +1,103 @@
...
+
+    ax : matplotlib Axes, default: None
+        The axes to plot the figure on. If None is passed in the current axes
+        will be used (or generated if required).
+    
+    scale : bool, default: True
+    
+    center : bool, default: True
+    
+    col : list or tuple, default: None
```
bbengfort: In other visualizers we call this argument "color" -- e.g. in [ParallelCoordinates](https://github.com/cjmorale/yellowbrick/blob/feature/pca_feature_visualizer/yellowbrick/features/pcoords.py#L59); any reason not to maintain this here?

yellowbrick/features/pca.py
```diff
@@ -0,0 +1,103 @@
...
+    >>> visualizer.poof()
+
+
+    Notes
+    -----
+    
+    
+    """
+    def __init__(self, ax=None, scale=True, center=True, col=None,
+                 colormap='RdBu_r', **kwargs):
```
bbengfort: We do have a default colormap, that we'd prefer you'd use, [`DEFAULT_SEQUENCE`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/style/palettes.py#L347); it's actually `RdBu` -- is reversed important? If so, keep this as it is!

yellowbrick/features/pca.py
```diff
@@ -0,0 +1,103 @@
...
+        self.pca_features_ = self.pca_transformer.transform(X)
+        self.draw()
+        return self.pca_features_
+
+    def draw(self, **kwargs):
+        X = self.pca_features_
+        if self.ax is None:
+            self.ax = self.gca()
+
+        self.ax.scatter(X[:, 0], X[:, 1], c=self.col, cmap=plt.cm.Paired)
```
bbengfort: In the init you save a colormap, but don't use it here, instead using the Paired map; any reason?

yellowbrick/features/pca.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def transform(self, X, y=None, **kwargs):
+        self.pca_features_ = self.pca_transformer.transform(X)
+        self.draw()
+        return self.pca_features_
+
+    def draw(self, **kwargs):
+        X = self.pca_features_
+        if self.ax is None:
+            self.ax = self.gca()
```
bbengfort: In a recent commit, we changed our axes management; so `if self.ax is None: self.ax = self.gca()` is no longer necessary; you can remove these two lines of code!

bbengfort:
@cjmorale we've been updating the code base, please make sure you hit the update branch button when you're ready to make sure that the tests pass for the 0.4.2 release. For example, the axes thing I mentioned in the code review depends on it.

yellowbrick/features/pca.py
cjmorale: Good point. updated line 33 to 

```{python}
    color : list or tuple of colors, default: None
        Specify the colors for each individual class
```

yellowbrick/features/pca.py
cjmorale: That is good to know. I am still getting familiar with the code base. Added 
`from yellowbrick.style import palettes` at imports and replaced line 64-65 with

```{python}
    def __init__(self, ax=None, scale=True, col=None, proj_dim=2,
                 colormap=palettes.DEFAULT_SEQUENCE, **kwargs):
```

yellowbrick/features/pca.py
cjmorale: I intended to have `cmap=self.colormap`. Good eye!

yellowbrick/features/pca.py
cjmorale: I might need some help on this one. I tried to update my branch but when I deleted those two lines the code would run completely :/

tests/test_features/test_pca.py
cjmorale: I removed the arguments and added a one line docstrings to the functions. 

I like to use `pytest.raises` to write test cases that should catch thrown exceptions. 

 Any thoughts about adding `pytest` to the requirements.txt?

yellowbrick/features/pca.py
cjmorale: Thanks!

yellowbrick/features/pca.py
bbengfort: Hmm, I went into your branch and it looks like you have the latest code. These two lines of code if they're in should cause an error at this point. Can you describe the error you were seeing?

cjmorale: @bbengfort it looks like the my code runs now after excluding lines 90-91. So that issue is now resolved! Below are two errors I am getting when I run `make test`

I am trying to wrap up some of the loose ends n my code tonight. The function now supports 3d plots and I am making test cases for the *quick method* you mentioned in your code review.

```{python}
ERROR: Test radviz on the real, occupancy data set
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/carlo/Documents/yellowbrick/tests/test_features/test_radviz.py", line 89, in test_integrated_radviz
  File "/Users/carlo/Documents/yellowbrick/myenv/lib/python3.5/site-packages/sklearn/base.py", line 497, in fit_transform
    return self.fit(X, y, **fit_params).transform(X)
  File "/Users/carlo/Documents/yellowbrick/yellowbrick/features/base.py", line 193, in fit
    self.draw(X, y, **kwargs)
  File "/Users/carlo/Documents/yellowbrick/yellowbrick/features/radviz.py", line 226, in draw
    self.ax.text(xy[0] + 0.025, xy[1] + 0.025, name, ha='left', va='bottom', size='small')
TypeError: text() missing 1 required positional argument: 's'


ERROR: Assert no errors occur during radviz visualizer integration
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/carlo/Documents/yellowbrick/tests/test_features/test_radviz.py", line 71, in test_radviz
    visualizer.fit_transform(self.X, self.y)
  File "/Users/carlo/Documents/yellowbrick/myenv/lib/python3.5/site-packages/sklearn/base.py", line 497, in fit_transform
    return self.fit(X, y, **fit_params).transform(X)
  File "/Users/carlo/Documents/yellowbrick/yellowbrick/features/base.py", line 193, in fit
    self.draw(X, y, **kwargs)
  File "/Users/carlo/Documents/yellowbrick/yellowbrick/features/radviz.py", line 226, in draw
    self.ax.text(xy[0] + 0.025, xy[1] + 0.025, name, ha='left', va='bottom', size='small')
TypeError: text() missing 1 required positional argument: 's'
```

cjmorale:
@rebeccabilbro @bbengfort I pushed some new code. I tried to address all of the code comments by @bbengfort. I added more test cases and also added the quick function. The function supports 2 and 3 dimensional plots. I added some examples in the example folder see the notebook at `examples/cjmorale/PCA_Examples.ipynb`

requirements.txt
```diff
@@ -4,6 +4,7 @@ scipy>=0.18
 scikit-learn>=0.18
 numpy>=1.11.0
 cycler>=0.10.0
+pytest>=3.0.7
```
bbengfort: We're really kind of wary of adding too many dependencies to Yellowbrick; we've had a tough time with the dependencies we do have in terms of getting a Conda package together, and dealing with different versions. At the very least, I'd ask you to move this dependency to `tests/requirements.txt` -- however, the way I see you using pytest below, there is actually already a solution using `unittest`; let me know what you think.

tests/test_features/test_pca.py
```diff
@@ -0,0 +1,233 @@
...
+             [2.110, 3.609, 4.330, 7.985, 5.595, ],
+             [2.110, 3.626, 4.330, 8.203, 5.621, ],
+             [2.110, 3.620, 4.470, 8.210, 5.612, ]]
+        )
+
+        y = np.array([1, 1, 0, 1, 0, 0])
+
+        params = {'scale': True, 'center': False, 'proj_dim': 4, 'col': y}
+
+        with pytest.raises(ValueError) as exception:
```
bbengfort: This line can be easily replaced by:

```python
with self.assertRaises(ValueError) as exception:
    ... 
```

tests/test_features/test_pca.py
```diff
@@ -0,0 +1,233 @@
...
+             [2.110, 3.609],
+             [2.110, 3.626],
+             [2.110, 3.620]]
+        )
+
+        y = np.array([1, 0])
+
+        params = {'scale': True, 'center': False, 'proj_dim': 3, 'col': y}
+
+        with pytest.raises(ValueError) as exception:
```
bbengfort: This line can be easily replaced by:

```python
with self.assertRaises(ValueError) as exception:
    ... 
```

tests/test_features/test_pca.py
```diff
@@ -0,0 +1,233 @@
+##########################################################################
+## Imports
+##########################################################################
+
+import unittest
+import numpy as np
+import numpy.testing as npt
+import pytest
```
bbengfort: Recommend removing this dependency, see below.

yellowbrick/features/pca.py
```diff
@@ -0,0 +1,176 @@
...
+    >>> visualizer.transform(X)
+    >>> visualizer.poof()
+
+    """
+    def __init__(self, X=None, y=None, ax=None, scale=True, color=None, proj_dim=2,
+                 colormap=palettes.DEFAULT_SEQUENCE, **kwargs):
+        super(PCADecomposition, self).__init__(ax=ax, **kwargs)
+        # Data Parameters
+        if proj_dim not in (2, 3):
+            raise ValueError("proj_dim object is not 2 or 3.")
```
bbengfort: It would be my preference that you raised a `YellowbrickValueError` (from the exceptions.py module) so that we can disambiguate between Yellowbrick exceptions and other exceptions. Note that `YellowbrickValueError` extends both `YellowbrickException` and `ValueError` so you can test for both as you're doing in the tests.

PR #238, opened 2017-05-22
add decomposition explained variance visualizer

georgerichardson:
First shot at visualizing the explained variance from a sklearn decomposition to help choose n_components. Currently uses PCA only.

``` python
from sklearn import datasets

bc = datasets.load_breast_cancer()
X = bc.data

from yellowbrick.features.decomposition import ExplainedVariance

p = ExplainedVariance()
p.fit_transform(X)
p.poof()
```
expvar


bbengfort:
@georgerichardson make sure you ping me when you're ready for a code review!

georgerichardson:
@bbengfort Sounds good. Been getting back into work, but should have some time this weekend, so feel free to give me comments!

yellowbrick/features/decomposition.py
```diff
@@ -0,0 +1,66 @@
...
+    >>> visualizer.fit(X)
+    >>> visualizer.transform(X)
+    >>> visualizer.poof()
+
+    Notes
+    -----
+    
+    """
+
+    def __init__(self, ax=None, n_components=None, scale=True, center=True, colormap='RdBu_r',
```
bbengfort: In the `yellowbrick.style.palettes` module we have a constant, `DEFAULT_SEQUENCE` (which is currently set to `"RdBu"`), which I'd recommend you use over a string constant for the colormap argument. Colors are so difficult, and it's also difficult to diagnose when things go wrong, so we try to handle most of our colors as variables or as simple 'b', 'g', 'r', 'm', 'y', 'c', 'k' strings which are set by the environment.

yellowbrick/features/decomposition.py
```diff
@@ -0,0 +1,66 @@
...
+        self.pipeline = Pipeline([('scale', StandardScaler(with_mean=self.center,
+                                                                   with_std=self.scale)), 
+                                                                  ('pca', PCA(n_components=self.n_components))])
+
+    def fit(self, X, y=None):
+        self.pipeline.fit(X)
+        return self
+
+    def transform(self, X):
+        self.draw(self.pipeline.steps[-1][1].explained_variance_)
```
bbengfort: You have a property, `explained_variance_` but are not setting it; I recommend that you do and allow draw to access it directly rather than passing it into draw. Also, you're not doing anything on transform, so perhaps it would be better to put this into `fit()`? 

```python
def fit(self, X, y=None):
    self.pipeline.fit(X) 
    self.explained_variance_ = self.pipeline.steps[-1][1].explained_variance_ 
    self.draw() 
```

Another option is to simply make `explained_variance_` a property so that you always access it off of the pipeline, even during updates:

```python
@property 
def explained_variance_(self):
    return self.pipeline.steps[-1][1].explained_variance_ 
```

This has the advantage of ensuring that even if the visualizer is fit multiple times, the right visualization will be drawn.

bbengfort:
@georgerichardson looks great so far!

bbengfort:
@georgerichardson better do a pull before you get started ;-)

georgerichardson:
Made the above recommendations and added a quick method.

I guess some tests should be next, along with some extra features for the visualizer like labelling and identifying key values.

bbengfort:
@georgerichardson awesome! Let us know when you're ready for us to review!

bbengfort:
@georgerichardson this is looking great so far, excited to see it completed!

bbengfort:
@georgerichardson any interest in finishing up those tests, or would you prefer one of us to carry this to the home stretch?

bbengfort:
@georgerichardson if it's alright, I'm going to go ahead and merge what you have as an undocumented feature then we can let someone else work on tests, etc. I'll open up an issue to carry this to the end.

PR #246, opened 2017-05-24
JtPlotVis: Only warn of matplotlib maj ver 1 if that is so (fix #243)

### LINKED ISSUE #243: JointPlotVisualizer insists on Matplotlib "2.0.0" exactly.
Warns if on mpl.__version__ == "2.0.2". That's overly picky, yes?

Line 133 in features/jointplot.py.

JimStearns206:
* Add unit to detect issuance of UserWarning about matplotlib.
* Verify unit test fails with matplotlib 2.0.2.
* Fix version check to check for major version >= 2, not exact match to 2.0.0.
* Change warning message from "Please upgrade to continue" to "Please upgrade";
this is warning, not error.


JimStearns206:
I believe the new unit test is passing - after some flubs on my part. 

The last Travis failure appears to be a network problem while pip installing scipy.

Any way to kick off another build, short of doing a no-op commit?

bbengfort:
@JimStearns206 yes, there is a restart button on the travis page; I just kicked it off again. I'll do the code review now.

yellowbrick/features/jointplot.py
```diff
@@ -129,13 +129,14 @@ def __init__(self, ax=None, feature=None, target=None,
                  xy_plot='hist', xy_args=None,
                  size=6, ratio=5, space=.2, **kwargs):
 
-        #check matplotlib version - needs to be version 2.0.0
-        if mpl.__version__ == "2.0.0":
+        # Check matplotlib version - needs to be version 2.0.0 or greater.
+        mpl_vers_maj = int(mpl.__version__.split(".")[0])
+        if mpl_vers_maj >= 2:
```
bbengfort: This is really minor, but I would propose a refactor as follows:

```python
if mpl_vers_maj < 2:
    warnings.warn((
        "{} requires matplotlib major version 2 or greater. "
        "Please upgrade."
     ).format(self.__class__.__name__))
```

I'm just a bit worried that in the current else form the logic may be accidentally broken by another contributor.

yellowbrick/features/jointplot.py
```diff
@@ -129,13 +129,14 @@ def __init__(self, ax=None, feature=None, target=None,
                  xy_plot='hist', xy_args=None,
                  size=6, ratio=5, space=.2, **kwargs):
 
-        #check matplotlib version - needs to be version 2.0.0
-        if mpl.__version__ == "2.0.0":
+        # Check matplotlib version - needs to be version 2.0.0 or greater.
+        mpl_vers_maj = int(mpl.__version__.split(".")[0])
```
bbengfort: It seems like it might be a good idea to put a helper function in `version.py`, e.g. something like `parse_version` so that we can reuse this in multiple places and also add exception handling (`int()` can throw a `TypeError`). 

Not for this PR, but perhaps you could also write an issue or a TODO for a `check_version` function that raises a warning or exception based on a version of a specific dependency? I think it might be something that comes up more, and other libraries have deprecation warning decorators and the like.

bbengfort:
Also make sure to close #243 when ready!

bbengfort:
@JimStearns206 I'm going to go ahead and merge since @cjmorale reported not seeing a squash and merge button.

PR #247, opened 2017-05-25
WIP pca_feature_visualization 

cjmorale:
Closed old pull request and issued another one. The old request was getting messy from multiple commits.

See old pull request here https://github.com/DistrictDataLabs/yellowbrick/pull/232


cjmorale:
@bbengfort I did not your comments about pytest and `YellowbrickValueError`. Can you recomment them here. I would like to get this pull request finished up by Friday!

requirements.txt
```diff
@@ -4,6 +4,7 @@ scipy>=0.18
 scikit-learn>=0.18
 numpy>=1.11.0
 cycler>=0.10.0
+pytest>=3.0.7
```
bbengfort: We are a bit wary about adding dependencies to Yellowbrick; we have had a lot of trouble with dependency management for different versions as well as building other tools like Conda packages. At the very least, this should be an optional dependency for developers and placed into `tests/requirements.txt` rather than the global  requirements. However, I don't believe that pytest is required here, and your use case can be easily replaced with unittest; see below.

cjmorale:
@bbengfort Do you have any ideas why `The Travis CI build failed`.

tests/test_features/test_pca.py
```diff
@@ -0,0 +1,233 @@
+##########################################################################
+## Imports
+##########################################################################
+
+import unittest
+import numpy as np
+import numpy.testing as npt
+import pytest
```
bbengfort: I'm always interested in new packages to make things better, but I think we're not getting any additional benefit in this particular case.

tests/test_features/test_pca.py
```diff
@@ -0,0 +1,233 @@
...
+             [2.110, 3.609, 4.330, 7.985, 5.595, ],
+             [2.110, 3.626, 4.330, 8.203, 5.621, ],
+             [2.110, 3.620, 4.470, 8.210, 5.612, ]]
+        )
+
+        y = np.array([1, 1, 0, 1, 0, 0])
+
+        params = {'scale': True, 'center': False, 'proj_dim': 4, 'col': y}
+
+        with pytest.raises(ValueError) as exception:
```
bbengfort: I propose that this test could be replaced with:

```python
with self. assertRaisesRegexp(ValueError, "proj_dim object is not 2 or 3"):
    PCADecomposition(**params)
```

tests/test_features/test_pca.py
```diff
@@ -0,0 +1,233 @@
...
+             [2.110, 3.609],
+             [2.110, 3.626],
+             [2.110, 3.620]]
+        )
+
+        y = np.array([1, 0])
+
+        params = {'scale': True, 'center': False, 'proj_dim': 3, 'col': y}
+
+        with pytest.raises(ValueError) as exception:
```
bbengfort: I propose that this test could be replaced with:

```python
with self. assertRaisesRegexp(ValueError, "n_components=3 must be between 0 and n_features"):
    PCADecomposition(**params)
    pca.fit(X)
```

yellowbrick/features/pca.py
```diff
@@ -0,0 +1,176 @@
...
+    >>> visualizer.transform(X)
+    >>> visualizer.poof()
+
+    """
+    def __init__(self, X=None, y=None, ax=None, scale=True, color=None, proj_dim=2,
+                 colormap=palettes.DEFAULT_SEQUENCE, **kwargs):
+        super(PCADecomposition, self).__init__(ax=ax, **kwargs)
+        # Data Parameters
+        if proj_dim not in (2, 3):
+            raise ValueError("proj_dim object is not 2 or 3.")
```
bbengfort: My preference is to raise a `YellowbrickValueError` found in the `yellowbrick.exceptions` module; this will allow us to easily detect yellowbrick exceptions from numpy or scikit-learn exceptions. Also `YellowbrickValueError` extends both `YellowbrickError` and `ValueError` so can be caught using:

```python
try:
    PCADecomposition() 
except ValueError:
    # catches YBValueError
```

tests/test_features/test_pca.py
cjmorale: Thanks for looking into this for me.

requirements.txt
cjmorale: I updated the unit tests. Thanks for your feedback in the code review

tests/test_features/test_pca.py
cjmorale: updated test_pca.py

tests/test_features/test_pca.py
cjmorale: updated test_pca.py

cjmorale:
@bbengfort I discovered something interesting today. On my local setup the RadViz unit tests were failing when I was running `make test`.

My unit tests lived in a file named `test_pca.py`. If I removed this file from the folder `tests/test_features` and ran `make test` again like magic the RadViz unit tests were passing and there were 0 errors.

If I put my `test_pca.py` file back into `tests/test_features` and ran `make test` 2 errors would occur.

Since I named my unit test file `test_pca.py` they would run before `test_pcoords.py` and `test_radviz.py`. I thought maybe if my test cases ran last in the `test_features` folder that things would be okay. I was curious if there may be some hidden ordering dependency for the tests. If that was true I could see how introducing my tests could break something else.

Therefore I renamed  `test_pca.py` to `test_x_pca.py` to see if running my new unit tests last in the `test_features` folder would maybe work. And well now the test run successfully.

bbengfort:
Huh, that's very strange; all of the tests should be encapsulated in their own test cases and have setup and teardown functions called before and after each one. I was looking into the RadViz errors but could not reproduce them, so this is good information. 

I looked at your test file and the RadViz tests and cannot see any shared dependencies right off the bat. Let's merge this feature in, and I'll take a look at it. 

Thanks for all your hard work on this PR - it looks great!

Issue #244, opened 2017-05-23
ConfusionMatrix classes: one label must be specified in y true. 


bbengfort:
When trying to passes class names to the ConfusionMatrix, I'm getting the following error:

```
/usr/local/lib/python3.6/site-packages/sklearn/metrics/classification.py:248: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if np.all([l not in y_true for l in labels]):
Traceback (most recent call last):
  File "rc.py", line 63, in <module>
    viz.score(X_test, y_test)
  File "/Users/benjamin/Repos/ddl/yellowbrick/yellowbrick/classifier/confusion_matrix.py", line 107, in score
    y, y_pred, labels=self.classes_, sample_weight=sample_weight
  File "/usr/local/lib/python3.6/site-packages/sklearn/metrics/classification.py", line 249, in confusion_matrix
    raise ValueError("At least one label specified must be in y_true")
ValueError: At least one label specified must be in y_true
```

This is do to the fact that the y_true vector is label encoded (e.g. ones and zeros) but that the labels are strings, e.g. "occupied" and "unoccupied". 

Here is the code to produce the error:

```python 
df = pd.read_csv('examples/data/occupancy/occupancy.csv')

features = ["temperature", "relative humidity", "light", "C02", "humidity"]
target = "occupancy"

X = df[features]
y = df[target]
classes = ["unoccupied", "occupied"]

X_train, X_test, y_train, y_test = tts(X, y, test_size=0.2)

viz = ConfusionMatrix(MultinomialNB(), classes=classes)
viz.fit(X_train, y_train)
viz.score(X_test, y_test)
viz.poof()
```

Note that:

```
viz = ConfusionMatrix(MultinomialNB())
```

Does not raise an exception.

cjmorale:
@bbengfort Is anyone working on this bug yet? I was able to reproduce the error last night on my local environment and wanted to offer my help.

ndanielsen:
@cjmorale I'm one of the maintainers of yb, you're welcome to take a stab at it! 💯

cjmorale:
@ndanielsen How should the system handle examples like @bbengfort provided above? I was a little unsure what type of behavior the yb team is looking for in these situations.

NealHumphrey:
@cjmorale I'm another maintainer and also authored the original ConfusionMatrix. 

[...]

PR #253, opened 2017-05-27
initial coding to resolve issue 244

cjmorale:
See issue https://github.com/DistrictDataLabs/yellowbrick/issues/244

The initial coding includes:
   - `ConfusionMatrix label_encoder example.ipynb` file in `examples/cjmorale` folder
   - two new test cases for class `ConfusionMatrixTests`
   - `character_mapping` util helper function
   - added another optional parameter to the init `label_mapping`
 
TODO:
   - Try to get test coverage to 100% for confusion_matrix.py. 
   - Update documentation
  
```{python}
Name                                               Stmts   Miss  Cover

yellowbrick/classifier/confusion_matrix.py           86      3    97%
```


tests/test_classifier/test_confusion_matrix.py
```diff
@@ -61,3 +60,28 @@ def test_one_class(self):
...
+
+    def test_inverse_mapping(self):
+        model = LogisticRegression()
+        classes = ['zero', 'one', 'two', 'three', 'four', 'five', 'six', 'seven', 'eight', 'nine']
+        cm = ConfusionMatrix(model, classes=classes)
+        cm.fit(self.X_train, self.y_train)
+        cm.score(self.X_test, self.y_test)
+
+    def test_isclassifier(self):
+        model = PassiveAggressiveRegressor()
```
bbengfort: Hehe, I had no idea a `PassiveAggressiveRegressor` existed ... that's quite the name!

yellowbrick/classifier/confusion_matrix.py
```diff
@@ -80,6 +84,7 @@ def __init__(self, model, ax=None, classes=None, **kwargs):
         self.cmap.set_under(color='w')
         self.cmap.set_over(color='#2a7d4f')
         self.edgecolors=[] #used to draw diagonal line for predicted class = true class
+        self.label_mapping= label_mapping
```
bbengfort: This is an extremely interesting idea, and I like what you've done. My question though is that I wonder if the user will actually have a mapping like the one you suggest. Later on you use the `LabelEncoder` for guessing the mapping between classes and numeric values, but likely that's what the user used to do the encoding (unless they used something like `pandas.get_dummies`). What would you think of having the user optionally pass in their label encoder, rather than a dictionary?

yellowbrick/classifier/confusion_matrix.py
```diff
@@ -103,6 +108,20 @@ def score(self, X, y, sample_weight=None, percent=True):
         """
         y_pred = self.predict(X)
 
+
+        if self.label_mapping:
+            # if a mapping is passed to class apply it here.
+            y = character_mapping(y, self.label_mapping)
+            y_pred = character_mapping(y_pred, self.label_mapping)
```
bbengfort: This makes sense, and I like it! It's a good solution, if the user is able to supply a mapping (see above).

yellowbrick/classifier/confusion_matrix.py
```diff
@@ -103,6 +108,20 @@ def score(self, X, y, sample_weight=None, percent=True):
...
+
+        if len(set(y_pred)) == len(set(self.classes_)):
+            if set(y_pred) != set(self.classes_):
+                self._label_encoder = LabelEncoder()
+                self._label_encoder.fit(self.classes_)
+
+                y = self._label_encoder.inverse_transform(y)
+                y_pred = self._label_encoder.inverse_transform(y_pred)
```
bbengfort: I get what you're trying to do here, however I'm wondering if the logic might be a tiny bit off. I'd like to discuss what I think the inputs to both `y_pred` and `self.classes_` are to see what you think. 

`self.classes_` can be either:

1. a user supplied list of classes that is converted into a `np.array`
2. the underlying encoding by the estimator, stored in `self.estimator.classes_`, also a `np.array`. 

In both cases `classes_` is a collection of string values (names). In the `self.estimator.classes_` case, we assume that `classes_[0]` maps to an encoded class `0` and so on and so forth, however it also means that we're expecting `y_pred` to contain strings! If `self.estimator.classes_` are integers, it means that `y_train` was encoded and therefore `y_pred` should be encoded as well. 

And that brings us to `y_pred` which will either be:

1. A vector of ints if `y_train` was encoded 
2. A vector of strings if it wasn't 

So my worry here is that label encoding the `classes_` might actually produce a false mapping depending on what the user passes to the constructor. (Of course, I'm not sure, and I'm happy to discuss it). 

Additionally, this method should also help us take care of "slicing" -- that is when the user suggests a subset of classes to display in the confusion matrix. 

Any thoughts?

yellowbrick/utils/helpers.py
```diff
@@ -28,6 +28,29 @@
...
+    array: iterable
+        an iterable object containing predictions from a classifier
+
+    Returns
+    -------
+    array : list
+        a list of predicted values after applying a given mapping
+
+    """
+    array = [str(x) for x in array]
```
bbengfort: Any reason for mapping `str` to every element in the array? This is a lot of overhead for larger collections and should be unnecessary as far as I can tell.

yellowbrick/utils/helpers.py
```diff
@@ -28,6 +28,29 @@
...
+    for key, value in mapping.items():
+        array = [x.replace(value, key) for x in array]
```
bbengfort: Ah, so `str.replace` -- replaces every instance if a character with the value from the key. E.g. `"bcb".replace("b", "a")` results in the string `"aca"`. Given the use case for target/class mapping, may I suggest:

```python
array = [mapping[x] for x in array] 
```

Also I'd note that the input is likely to be a `np.array`. I investigated `np` replace methods, but most of them use masks, e.g. `array[array == x] = mapping[x]` will replace all values of x with the mapping character, but this is probably slower than looping through every element in the array since you'll be replacing all values anyway.

yellowbrick/classifier/confusion_matrix.py
NealHumphrey: @bbengfort I suggested the dictionary in #244. But, I agree that it should be user choice - either the `LabelEncoder` itself OR a dictionary mapping. It could use duck typing with the same variable (label_mapping) - use a `try:` block with inverse_transform, and if `self.label_mapping` doesn't have that method (because it's not a `LabelEncoder`, treat it as a dictionary.

cjmorale: I made a code change where the user can pass a dictionary mapping or a trained `LabelEncoder`

yellowbrick/utils/helpers.py
cjmorale: I use the str.replace function later to apply the mapping

yellowbrick/classifier/confusion_matrix.py
cjmorale: @bbengfort I made several changes to the code from the first code review. I was wondering if you could take a look at the new code and see if this comment is still relevant.

cjmorale:
@bbengfort @NealHumphrey I made several code changes from the first code review. Would either of you have a chance to look over the new code?

Thanks!

tests/test_classifier/test_confusion_matrix.py
```diff
@@ -61,3 +60,30 @@ def test_one_class(self):
...
         cm.score(self.X_test, self.y_test)
+
+    def test_defined_mapping(self):
+        model = LogisticRegression()
+        classes = ['one', 'two', 'three', 'four', 'five', 'six', 'seven', 'eight', 'nine']
+        mapping = {0: 'zero', 1: 'one', 2: 'two', 3: 'three', 4: 'four', 5: 'five',
+                   6: 'six', 7: 'seven', 8: 'eight', 9: 'nine'}
+        cm = ConfusionMatrix(model, classes=classes, label_encoder = mapping)
+        cm.fit(self.X_train, self.y_train)
+        cm.score(self.X_test, self.y_test)
```
bbengfort: In both this test and the one below, X and y are a matrix and vector of integers, is that correct?

yellowbrick/classifier/confusion_matrix.py
```diff
@@ -55,6 +55,8 @@ class ConfusionMatrix(ClassificationScoreVisualizer):
         indicated by that function. It may be used to reorder or select a subset of labels.
         If None, values that appear at least once in y_true or y_pred are used in sorted order.
         Default: None
+        
+    label_mapping : dict, a dictionary in the form {class: encoded_class}
```
bbengfort: this docstring needs to be updated, I propose the following:

```
label_mapping : dict or LabelEncoder, default: None 
    When specifying the ``classes`` argument, the input to ``fit()`` and ``score()`` must match the 
    expected labels. If the ``X`` and ``y`` datasets have been encoded prior to training and the 
    labels must be preserved for the visualization, use this argument to provide a mapping from the 
    encoded class to the correct label. Because typically a Scikit-Learn ``LabelEncoder`` is used to 
    perform this operation, you may provide it directly to the class to utilize its fitted encoding. 
```

I would also encourage you to review the `classes : ` docstring to make sure it matches this.

yellowbrick/classifier/confusion_matrix.py
```diff
@@ -77,9 +79,10 @@ def __init__(self, model, ax=None, classes=None, **kwargs):
         self.confusion_matrix = None
 
         self.cmap = color_sequence(kwargs.pop('cmap', 'YlOrRd'))
-        self.cmap.set_under(color='w')
-        self.cmap.set_over(color='#2a7d4f')
-        self.edgecolors=[] #used to draw diagonal line for predicted class = true class
+        self.cmap.set_under(color = 'w')
+        self.cmap.set_over(color = '#2a7d4f')
```
bbengfort: Minor, and I'm not sure when this got added, but since I'm looking at it -- the hard coded color hex values make me nervous, they're difficult to find and debug. Would you mind making this a module variable, e.g. 

```python
CMAP_OVERCOLOR = '#2a7d4f' 

class ConfusionMatrix(Visualizer):
    ... 

    def __init__(self):
        self.cmap.set_over(color = CMAP_OVERCOLOR) 
```

I'd really appreciate it !

bbengfort: Note that the `'w'` is completely fine.

yellowbrick/classifier/confusion_matrix.py
```diff
@@ -103,6 +106,16 @@ def score(self, X, y, sample_weight=None, percent=True):
...
+
+        if self.label_encoder:
+            try :
+                y = self.label_encoder.inverse_transform(y)
+                y_pred = self.label_encoder.inverse_transform(y_pred)
+            except AttributeError:
+                # if a mapping is passed to class apply it here.
+                y = [self.label_encoder[x] for x in y]
+                y_pred = [self.label_encoder[x] for x in y_pred]
+
```
bbengfort: Perfect!

tests/test_classifier/test_confusion_matrix.py
cjmorale: Correct

yellowbrick/classifier/confusion_matrix.py
cjmorale: Easy fix!

bbengfort:
@cjmorale looks great; I'm going to go ahead and merge. Thanks!

Issue #69, opened 2016-10-07
Rank 1D


bbengfort:
Rank 1D is an analysis of a single feature; usually by inspecting a histogram. Rankings of features are conducted though a metric that takes a single feature and returns a score between [-1, 1] or [0, 1]. 

The visualizer can either show these ranks as a column of features or as a bar chart; still undecided. 

Ranking mechanisms to be implemented:

- [ ] Normality of distribution (Shapiro-Wilk and Kolmogorov-Smirnov) 
- [ ] Uniformity of distribution (entropy)
- [ ] Number of potential outliers 
- [ ] Number of hapaxes 
- [ ] Size of gap 

See: Seo, Jinwook, and Ben Shneiderman. "A rank-by-feature framework for interactive exploration of multidimensional data." Information visualization 4.2 (2005): 96-113.

rebeccabilbro:
As a start, we could implement a simple a histogram, a bit like what is described in this [post](http://www.openjems.com/enhanced-feature-importance-plot-in-scikit-learn-with-rs-ggplot2/).

Issue #18, opened 2016-06-01
Single feature analysis tools (augmenting Rank1D)


rebeccabilbro:
Radviz, parallel coordinates, and Rank2D are already implemented. We should expand the feature visualization toolset to include single feature analysis tools:
- [ ] boxplots
- [ ] histograms
- [ ] violinplots

Create a `SingleFeatureVisualizer` class whose `poof` method takes as input the name(s) or position(s) of one of the features. Display the appropriate visualization allowing multiple `poof` calls with different features. Make sure to integrate this with Rank1D.

Open questions: do we need to store the data? Or can we create all visualization and then display on demand, or is there some middle ground? 

Note: similar to parallel coordinates and radviz, these need to be native implementations, emulating other libs (seaborn, pandas) as necessary.

rebeccabilbro:
Also, add in violinplots because they are awesome: https://stanford.edu/~mwaskom/software/seaborn/generated/seaborn.violinplot.html

rebeccabilbro:
Note: for any contributors interested in taking these on, we can break these into three separate issues!

ChrisPappalardo:
Hello.  I'm at Pycon 2018 dev sprints today and will take a stab at this issue.

lschumm:
pycon 2019 here. taking a look at this...

rebeccabilbro:
Covered by [`Rank1D`](https://www.scikit-yb.org/en/develop/api/features/rankd.html#rank-1d) as well as follow-on issues #69 #83  #435

PR #258, opened 2017-06-08
Rank 1D (WIP)

pbs929:
Hello!  I'm new here but I saw Rebecca's talk at PyData last year and loved the concept.  I recently started working with @ayota, who encouraged me to contribute. 

I thought issue #69 would be a good place to get my feet wet.  As I understand it, the aim is to compute a single summary score for each feature and to display the scores (preferably in order) in a bar chart.  

Side note - this visualizer seems to me to be distinct from the one proposed in #18.  There, the aim is to come up with some more complex visualization for each feature (e.g. a histogram), whereas here we are looking for a single number.  The descriptions for the issues seemed to suggest that the two visualizers were related somehow, but I couldn't figure out exactly what you all have in mind.  Let me know if I'm missing something. :)


bbengfort:
@pbs929 welcome! We're glad that you're interested in contributing to yellowbrick, and absolutely rank 1D is an excellent place to start. 

You're absolutely correct that the spec for #69 creates a single metric per feature; a bar chart would be an excellent way to visualize this, keeping in mind that some metrics can be from [-1,1]. One of the places this question gets really interesting is what happens when there are 35-70 features (as is typical of most problems). I also wouldn't worry too much about sorting, since feature identification can become a challenge that we'll discuss later on! 

As for #18 - this is more of an interactive augmentation, e.g. you click on a bar and a histogram or a box plot is displayed for the feature. We need #69 before #18, so feel free not to worry about that issue until the future! 

I'd recommend building your Visualizer in `yellowbrick/features/rankd.py` but a first step might be to get the metrics together for the bar chart. If you'd like you can create a module `yellowbrick/metrics.py` and we can start tossing them in there. Alternatively you can find the metrics in Scipy, Numpy, or Scikit-Learn and we'll just get started organizing them. 

Let me know if there is any other advice I can give!

pbs929:
Great, thanks for the clarification.  The `metrics` module sounds a good idea.  My thought is that I'll start by implementing one or two very simple metrics and go ahead and get the visualizer sorted out.  Then I or someone else can add in more metrics as they feel so inclined.  

Looking forward to getting started!  I'm sure I'll have more questions as I start digging into the code.

pbs929:
@bbengfort, see if this looks good as a start.  I emulated the existing Rank2D as closely as I could.  Because Rank1D and Rank2D are similar, I also refactored the common methods into a base class that both can inherit.  

I implemented the Shapiro-Wilk metric as a lambda for now similar to how Rank2D metrics are done.  I also took the chance to make a couple improvements on Rank2D including using the feature names as axis labels.  Let me know if you'd prefer to see that in a different PR.  :)

yellowbrick/features/rankd.py
```diff
@@ -80,23 +128,21 @@ def rank2d(X, y=None, ax=None, algorithm='pearson', features=None,
 
 
 ##########################################################################
-## Rank 2D Feature Visualizer
+## Base Feature Visualizer
 ##########################################################################
 
-class Rank2D(FeatureVisualizer):
+class RankDBase(FeatureVisualizer):
```
bbengfort: Yep, a base class for these two visualizers makes a lot of sense!

yellowbrick/features/rankd.py
```diff
@@ -126,18 +172,13 @@ class Rank2D(FeatureVisualizer):
     process, but can and should be set as early as possible.
     """
 
-    ranking_methods = {
```
bbengfort: We should leave a `ranking_methods` class property on the base as just an empty dictionary I think; to ensure that a `KeyError` is raised by subclasses that don't implement this rather than an `AttributeError` -- I think that would help us more quickly debug.

yellowbrick/features/rankd.py
```diff
@@ -126,18 +172,13 @@ class Rank2D(FeatureVisualizer):
...
-        'pearson': lambda X: np.corrcoef(X.transpose()),
-        'covariance': lambda X: np.cov(X.transpose()),
-    }
-
-    def __init__(self, ax=None, algorithm='pearson', features=None,
+    def __init__(self, ax=None, algorithm='shapiro', features=None,
```
bbengfort: The default here should probably be `algorithm=None` since both `shapiro` and `pearson` are subclass-specific.

yellowbrick/features/rankd.py
```diff
@@ -166,11 +207,6 @@ def fit(self, X, y=None, **kwargs):
         self : instance
             Returns the instance of the transformer/visualizer
         """
-
-        # TODO: This class is identical to the Parallel Coordinates version,
```
bbengfort: Is this TODO no longer valid?

yellowbrick/features/rankd.py
```diff
@@ -240,27 +275,173 @@ def rank(self, X, algorithm=None):
                 "'{}' is unrecognized ranking method".format(algorithm)
             )
 
+        # Extract matrix from dataframe if necessary
+        if is_dataframe(X):
```
bbengfort: If someone specifies a list of features to rank (e.g. a subset of the total features) in both the dataframe or the named object array, it looks like this is the point where the filtering has to be done. Not sure if you did that further down, but it's something to consider.

yellowbrick/features/rankd.py
```diff
@@ -205,12 +240,12 @@ def transform(self, X, **kwargs):
...
         -------
-        Xp : ndarray
+        X' : ndarray
             The transformed matrix, X'
         """
         # Rank and draw the input matrix
-        Xp = self.rank(X)
-        self.draw(Xp, **kwargs)
+        R = self.rank(X)
+        self.draw(R, **kwargs)
```
bbengfort: So why `R` here, for rank matrix? I get that `X'` stands for X prime, as does presumably `Xp`. My first reaction was that this is a bit of a faux pas, simply changing a variable name to something you prefer is bad practice with open source as it can lead to confusion or worse, start a turf fight. However, if the original names were leading to confusion, then it is a good thing to introduce clarity, and in this case, I think some clarity is needed, and should be enhanced with additional documentation. In that spirit I suggest the following:

```python
def transform(self, X, **kwargs):
    """
    Calling the ``transform()`` method is the primary hook for drawing feature rankings. 

    ...

    Returns 
    -------
    X : ndarray 
        Typically a transformed matrix, X' is returned. However, this method
        performs no transformation on the original data, instead simply 
        ranking the features that are in the input data and returns the original 
        data, unmodified. 
    """
    
    # R is a n dimensional, symmetric array where n is the number of features. 
    # E.g. for 1D ranking, R is (n,) and for 2D ranking R is (n,n) and so forth. 
    R = self.rank(X)
    self.draw(R, **kwargs)
    ... 
``` 

What do you think? Does that help make things more obvious to both users and developers?

yellowbrick/features/rankd.py
```diff
@@ -240,27 +275,173 @@ def rank(self, X, algorithm=None):
...
+        order features as well as visualize the result.
+        """
+        super(Rank1D, self).__init__(ax=None, algorithm=algorithm, features=features,
+                         colormap=colormap, **kwargs)
+
+    def draw(self, R, **kwargs):
+        """
+        Draws the heatmap of the ranking matrix of variables.
+        """
+        self.ax.bar(np.arange(len(R)), R, color=self.colormap[0])
```
bbengfort: This is very minor, but I'm worried that using the first indexed color of the colormap will produce variable results (and this can make testing very tricky, beyond simply annoying users a bit). Since you're only using a single color, I recommend simply using the value `'b'` here -- the discrete color cycle in matplotlib is `'bgrmyck'` and so `'b'` is always the first color we use for yb visualizations (giving them coordinated look and feel). Note that using this notation the `'b'` will be equivalent to the `set_palette()` color `'b'` and will therefore maintain the user's style preferences.

yellowbrick/features/rankd.py
```diff
@@ -240,27 +275,173 @@ def rank(self, X, algorithm=None):
...
+        self.ax.set_xticks(np.arange(len(R)))
+        self.ax.xaxis.grid(False)
+        self.ax.set_xticklabels(self.features_, rotation=90, ha='left')
```
bbengfort: Tick labels seem to be a `finalize()` thing to me; and I have comments related to this in the `draw()` of Rank2D. I don't think any change is necessary right now, but I wanted to alert you to the fact that in the future some refactoring may occur along those lines. 

Also, thank you for adding the tick labels for both 1 and 2D, however there are a couple of scenarios that we should consider (and are the reason they were originally omitted from 2D). 

1. A user might not specify feature names 
2. There may be (and is usually the case) > 35 features 

In the first case, what will happen if you do `self.ax.set_xticklabels(None, rotation=90, ha='left')`? This is, after all the default case (and a good test to write!) 

In the second case, the features get jammed together and become unreadable. This was constantly happening to me with my models in Rank2D. We do need to figure something out, and I'm open to suggestions. For now, however (to get your PR merged quickly) I suggest the following:

1. Create an option in `RankDBase` for `show_feature_names` or something like that, which defaults to `True` but can be turned off if the labels get jammed up. 
2. Create an issue "Better feature name labeling for RankD methods" that discusses various strategies. 

Strategies include:

1. Showing on the first and last feature on all axes (in the Rank2D and if we ever have a Rank3D) 
2. Eliminating every other feature name from the axes after a certain number of dimensions 
3. Using interactive graphs with hover to show which feature/pair of features exist.

yellowbrick/features/rankd.py
bbengfort: Also, you mentioned sorting in the original PR; if the user specifies an order in `features`, it should be maintained; but if any sorting occurs (probably an option to pass to `__init__`, then I think it should also be done here! What do you think?

yellowbrick/features/rankd.py
bbengfort: I noticed that you are using a horizontal bar chart here, where the features are on the X-axis and the rank score is on the Y-Axis. Generally speaking, most models will employ >35 features, I'd suggest that potentially having a taller image will be a bit more readable than a landscape image (e.g. features on the Y-axis and ranks on the X-axis). However, I also think this might be best left to the user. 

What do you think about putting in an option in `__init__` that allows the user to chose the orientation? Pandas allows this kind of specification via the `kind` argument ([pandas plot methods](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.plot.html#pandas.DataFrame.plot))  where `bar` is a vertical plot and `barh` is a horizontal plot. We could create an `orientation='h'` or `'v'` arguments and do the same?

yellowbrick/features/rankd.py
```diff
@@ -240,27 +275,173 @@ def rank(self, X, algorithm=None):
...
+    def __init__(self, ax=None, algorithm='pearson', features=None,
+                 colormap='RdBu_r', **kwargs):
+        """
+        Initialize the class with the options required to rank and
+        order features as well as visualize the result.
+        """
+        super(Rank2D, self).__init__(ax=None, algorithm=algorithm, features=features,
+                         colormap=colormap, **kwargs)
+
+    def draw(self, R, **kwargs):
```
bbengfort: RE: discussion from above; if you wouldn't mind adding docstring info about what `R` is in this case! 

Alternatively, we could remove the R argument to the method, and instead store it in `self.scores_` -- this would make the scores available to the user, which they might appreciate. What do you think about that?

yellowbrick/features/rankd.py
```diff
@@ -271,20 +452,8 @@ def draw(self, X, **kwargs):
...
-    def finalize(self, **kwargs):
-        """
-        Finalize executes any subclass-specific axes finalization steps.
-        The user calls poof and poof calls finalize.
-
-        Parameters
-        ----------
-        kwargs: dict
-            generic keyword arguments
+        # Reverse the rows to get the lower left triangle
```
bbengfort: Maybe I just can't see it in the review (loc are omitted in the diff), but don't forget to call `super()` that's where the title gets set!

yellowbrick/features/rankd.py
```diff
@@ -271,20 +452,8 @@ def draw(self, X, **kwargs):
...
+        self.ax.invert_yaxis()
 
-        """
-        # Set the title
-        self.set_title(
-            "{} Ranking of {} Features".format(
-                self.ranking_.title(), len(self.features_)
-            )
-        )
+        self.ax.set_xticklabels(self.features_, rotation=90, ha='left')
```
bbengfort: Same comment as before about passing in self.features_ --> what if it's None, etc.

pbs929:
@bbengfort Those are great comments, thanks!  I'll take some time to go through and respond over the next several days.  

As for testing, I'm willing to take a stab at that as well, just wanted to get your feedback on this part of it first.  Stay tuned.

bbengfort:
@pbs929 awesome, looking forward to it! Ping me when you're ready to take another look/ need help with writing tests. 

Also, we've set up a Google group if you want to join that:

https://groups.google.com/forum/#!forum/yellowbrick

Ben

yellowbrick/features/rankd.py
pbs929: As far as I can tell, yes.  ParallelCoordinates and RadViz now inherit their `fit` methods from `features.base.DataVisualizer`, but the functionality there is not applicable to RankD. 

From a quick scan over the other modules in the `features` package, I don't see any others that do the same thing on `fit` that we're doing here, so there's probably no need to hoist this higher for now.

yellowbrick/features/rankd.py
pbs929: done

yellowbrick/features/rankd.py
pbs929: agreed, done

yellowbrick/features/rankd.py
pbs929: Yes, my aim here was to add clarity.  The docstring in `transform` specified that `Xp` was the method's return value, but `Xp` actually referred to the rank matrix in the code, and that variable is not returned.  On the other hand the docstring for the `rank` method referred to the same rank matrix as `R`, so I thought it was best to keep consistent and call it the same thing everywhere.  

In any case, I think your explanations will help mitigate confusion in the future.  I've added them in.

yellowbrick/features/rankd.py
pbs929: My interpretation of the original intent for the `features` argument was that it was to be used to pass in labels for display on the plot, not for filtering or re-ordering columns.  However, if you'd like to add that functionality, we could do it.  

The only issue that I see is that this leads to an asymmetry in the treatment of numpy arrays vs. dataframes.  In the dataframe case, passing a `features` argument will potentially reorder the columns, while in the numpy case the same `features` argument will leave the order the same but just label the columns.  So, the same data and same `features` argument will give you different results depending on which type the data is.  Might be confusing, but maybe that's something we can live with?  Just let me know what you think and we can go either way.

yellowbrick/features/rankd.py
pbs929: Sounds good - this was a bit of a hack since I wasn't sure how color palletes and everything work :)

pbs929: Great idea!  Will do.

yellowbrick/features/rankd.py
pbs929: Thanks for that context.  

I included the axis labeling in the `draw` method because to me having labels is central to making the graph interpretable, and the `finalize` method is not called by the "quick" methods.  So, if labels are not added in `draw`, this makes the quick methods much less useful IMO. 

If the user does not provide feature names, the class is already set up to label the ticks with integers.  I should have included a demo of this, but it's already taken care of. 

On your second point about cases with many features, I agree that's a bit of a hairy issue but the approach you proposed should work for now.  I think it's useful to have the ability to put in the labels even if you have to turn them off for high-dimensional data.

yellowbrick/features/rankd.py
pbs929: I like it!  I would prefer to name it something like `self.ranks_` instead, though, to avoid confusion with results of sklearn's `score` method.

pbs929: Related general question on style... In Scikit-learn, attribute names with trailing underscores are reserved for attributes assigned during `fit`, while attributes passed through the constructor do not have underscores.  I'm guessing there's no similar convention in yellowbrick?

yellowbrick/features/rankd.py
pbs929: As in the other case the feature names default to integers... I think that was a feature that you built!  :)

yellowbrick/features/rankd.py
pbs929: yup, got it

pbs929: The title gets set in `finalize`, not in `draw`.

yellowbrick/features/rankd.py
bbengfort: Also Seaborn has an `orient=` argument that is also `h` or `v`: http://seaborn.pydata.org/tutorial/categorical.html#plotting-wide-form-data

bbengfort:
@pbs929 ping me when you're ready to go through the review again - wasn't sure if your last commit was ready for me to look at or not.

yellowbrick/features/rankd.py
bbengfort: Ok, sounds good!

yellowbrick/features/rankd.py
bbengfort: Yeh, we're dealing with that challenge in a lot of places. Let's not worry about it for this PR but make an issue regarding it so that we can explore it in the future. I'm more keen to get this merged sooner rather than delay on figuring stuff like this out completely!

yellowbrick/features/rankd.py
bbengfort: Sorry, thought that was in finalize, I'm with you.

yellowbrick/features/rankd.py
bbengfort: Yeh, that totally makes sense - my initial reaction was wrong and I'm glad I thought it through before responding! Thanks for updating the docstring and comments.

pbs929:
@bbengfort, ready for review I think. 

I was going to take a crack at tests but I may take you up on your offer for you to stub them out since your'e more familiar with what best practices you're going for with the project and so forth.  The ipynb runs through the major non-error cases.  For error handling we might want to check what happens when someone gives a bad argument for `orient` or `algorithm` (both should raise `YellowbrickValueError`).

yellowbrick/features/rankd.py
bbengfort: @pbs929 did you/could you make this issue?

Issue #59, opened 2016-10-06
Improve Parallel Coordinates


bbengfort:
See #50 for a more detailed discussion. 

There are several things that need to be done to improve the parallel coordinates visualizer:
- [x] Benchmarking, it is currently SLOW
- [x] Add a `draw_instance` method so that instances can be added to the figure at any time
- [x] Add `DataFrame` support so that the visualizer can accept either an `ndarray` or a `DataFrame` as input.
- [x] Create a subclass `NormalizedParallelCoordinates` that normalizes the data to the space 0 to 1 before drawing the coordinates. 
- [x] Add subsampling of instances to reduce clutter and improve performance 
- [x] Add fast vs. slow drawing methods for performance
- [x] Add alpha so that you can see instances through the other lines 
- [ ] <strike>Create an optimization technique that reorders the columns such that the overlap of two instances by the same class is minimized</strike>

There are probably several more things and there are comments in the Parallel Coordinates codebase as well.

thekylesaurus:
At PyCon 2018 sprints.  I'm going to take a pass at some of these items.

thekylesaurus:
@bbengfort - should we also include an item to make the "sample" capability select random rows, as opposed to ordered rows.  For data sets with sorted targets, you don't end up with a true representative sample.

bbengfort:
@thekylesaurus I think that would be a great idea - [I think we were planning on making sample a uniform random sample](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/features/pcoords.py#L244) but the number of choices for how to conduct sampling led us to simply use `head`. If there is a sklearn-method that does sampling we could use that as a guide for how to treat the `sample` parameter, otherwise perhaps we could simply add a `shuffle=True` argument which would make `head` a uniform random sample? What do you think?

thekylesaurus:
@bbengfort - shuffle=True would definitely get the job done; but I'll check sklearn to see if there is a different convention that we should inherit.

PR #259, opened 2017-06-10
Normalized pcoords and pcoords improvements

pbs929:
from issue #59:
* add dataframe compatibility
* reduce alpha for better plot appearance
* add NormalizedParallelCoordinates subclass


pbs929:
I added NormalizedParallelCoordinates as a subclass as specified in the issue, but normalization could just as well be specified as a boolean argument to ParallelCoordinates itself IMO.

pbs929:
Another idea regarding the speed/performance issue... Drawing a line for each of potentially thousands or millions of examples in a dataset is always going to be slow in matplotlib.  A better approach might be to provide a `max_samples` argument that defaults to, say, 1000, and to select a random subset of examples for display. I don't think human eyes can visually process many more lines than that, anyway.

ndanielsen:
@pbs929 this is a great contribution. I'm flagging @bbengfort to review this as he was the original contributor for this visualizer

bbengfort:
I agree that this should be a boolean argument to the ParallelCoordinates; the idea behind the additional subclass is if we wanted to chain them together transformer style, but that doesn't seem to be where the API is going. 

I also like the idea of sampling, though I'd prefer `sample=1.0` (or None) as the default, and have the argument be a floating point number of the percent of the data set to sample; e.g. 0.1 or 0.2 -- what do you think about that idea?

yellowbrick/features/pcoords.py
```diff
@@ -231,3 +294,82 @@ def finalize(self, **kwargs):
...
+        coordinates canvas and draws each instance and vertical lines on it.
+        """
+        # Convert from dataframe
+        if is_dataframe(X):
+            X = X.as_matrix()
+
+        # Normalize
+        maxes = X.max(axis=0)
+        mins = X.min(axis=0)
+        X_normed = (X - mins) / (maxes - mins)
```
bbengfort: This is certainly a decent method of normalization, but there are quite a few methods and we may want to provide the user the option to choose normalization or standardization. I'd also prefer if we used the built in Sklearn Transformers, in this case: [MinMaxScaler](http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html#sklearn.preprocessing.MinMaxScaler), [MinAbsScalar](http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MaxAbsScaler.html#sklearn.preprocessing.MaxAbsScaler), the [Normalizer](http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.Normalizer.html#sklearn.preprocessing.Normalizer), or the [StandardScalar](http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler). This would work something like an internal pipeline, which other methods like the TSNE method or the PCA decomposition have done.

yellowbrick/features/pcoords.py
```diff
@@ -231,3 +294,82 @@ def finalize(self, **kwargs):
         # Set the legend and the grid
         self.ax.legend(loc='best')
         self.ax.grid()
+
+
+##########################################################################
+## Normalized Static Parallel Coordinates Visualizer
+##########################################################################
+
+class NormalizedParallelCoordinates(ParallelCoordinates):
```
bbengfort: I think you're right, this should just be an argument to the ParallelCoordinates visualizer, the subclass doesn't add much exception confusion!

pbs929:
All good ideas, I will implement those changes.

pbs929:
I added an argument 'sample' that allows the user to specify either a fraction or an integer number of samples to display.  I also refactored normalization as an optional argument and added several normalization methods as requested.  See what you think!

yellowbrick/features/pcoords.py
```diff
@@ -56,14 +59,20 @@ def parallel_coordinates(X, y=None, ax=None, features=None, classes=None,
     classes : list of strings, default: None
         The names of the classes in the target
 
+    normalize : string or None, default: None
+        specifies which normalization method to use, if any
+
+    sample : float or int, default: 1.0
+        specifies how many examples to display from the data
+
```
bbengfort: It would be helpful if the normalize and sample docstrings matched that of the visualizer.

yellowbrick/features/pcoords.py
```diff
@@ -169,6 +200,37 @@ def draw(self, X, y, **kwargs):
         Called from the fit method, this method creates the parallel
         coordinates canvas and draws each instance and vertical lines on it.
         """
+        # Convert from dataframe
+        if is_dataframe(X):
+            X = X.as_matrix()
+
+        # Choose a subset of samples
```
bbengfort: My feeling is that normalization and sampling should go in `transform()` or `fit()` - what do you think?

yellowbrick/features/pcoords.py
```diff
@@ -169,6 +200,37 @@ def draw(self, X, y, **kwargs):
...
+        else:
+            raise YellowbrickTypeError(
+                "`sample` parameter must be int or float"
+            )
+        X = X[:n_samples, :]
+
+        # Normalize
+        if self.normalize is not None:
+            if self.normalize not in self.normalizers:
+                raise YellowbrickValueError(
```
bbengfort: This seems a little late for an exception to be raised, and also I'd like to keep the fitted normalizer around in case we need to use `inverse_transform` on it. My suggestion would be to hoist this to `__init__` and instead of keeping self.normalize, actually making it the normalizer as follows:

```python
class ParallelCoordinates(DataVisualizer):

    normalizers = {
        'minmax': MinMaxScalar,
        'minabs': MinAbsScalar, 
        'standard': StandardScalar, 
        'l1': lambda: Normalizer('l1'), 
        'l2': lambda: Normalizer('l2'), 
    }

    def __init__(self, **kwargs):
        ....

        if normalize and normalize not in self.normalizers:
            raise YellowbrickValueError() 
        self.normalize = self.normalizers.get(normalize, None) 

    def fit(self, X, y):
        if self.normalize:
            self.normalize.fit(X) 
        # maybe also compute indices of the sample 

    def transform(self, X):
        # maybe also perform sampling 
        if self.normalize:
            return self.normalize.transform(X) 
        return X
```

Of course, we need to figure out where `draw()` is called to make this work.

yellowbrick/features/pcoords.py
```diff
@@ -169,6 +200,37 @@ def draw(self, X, y, **kwargs):
...
+                    "`sample` parameter of type `int` must be greater than 1"
+                )
+            n_samples = min([self.sample, len(X)])
+        elif isinstance(self.sample, float):
+            if self.sample <= 0 or self.sample > 1:
+                raise YellowbrickValueError(
+                    "`sample` parameter of type `float` must be between 0 and 1"
+                )
+            n_samples = int(len(X) * self.sample)
```
bbengfort: We should probably keep track of n_samples to display in the visualizer somehow, e.g. the title or a legend.

yellowbrick/features/pcoords.py
```diff
@@ -169,6 +200,37 @@ def draw(self, X, y, **kwargs):
...
+        else:
+            raise YellowbrickTypeError(
+                "`sample` parameter must be int or float"
+            )
+        X = X[:n_samples, :]
```
bbengfort: Is this a random uniform sample? I think the sampling method is important, or we can add a `shuffle=True` option and randomly shuffle the deck, then take the first `n_samples`.

yellowbrick/features/pcoords.py
pbs929: I was following the existing convention in this module where the docstrings in the quick method are abbreviated.  I'll go ahead and make all of them match.

bbengfort: Hmm, I didn't realize that was a convention - perhaps I just put in docstring stubs and meant to update them later? Needless to say, I'd appreciate if you made the match, but don't let it block the merge.

pbs929: Okay, no problem.

yellowbrick/features/pcoords.py
pbs929: I kind of like having it in `draw` because this preprocessing is a required prerequisite to the actual drawing code but is not specific to either `fit` or `transform`.  So, e.g., if later on someone wanted to move the call to `draw` from `fit` to `transform` (say, in a derived class) they could do so without having to reproduce all the preprocessing by hand.  It also means that here I don't have to worry about which of those methods in the base class calls `draw`, meaning someone could change the base class and no updates would be needed here.  

So, in my opinion it is more elegant this way, but all that being said I'm open to doing it the other way too if you'd prefer.

yellowbrick/features/pcoords.py
pbs929: Good call, will do.

yellowbrick/features/pcoords.py
pbs929: Nope I did this in a deterministic way on this cut (it takes the first n_samples from the data) since it makes it easier to see what is going on when you see the same samples visualized on each run.  I like the idea of an optional `shuffle` argument though.

yellowbrick/features/pcoords.py
pbs929: So, correct me if I'm way off here, but my understanding of the purpose of this and other transformers in the package is that these should produce visualizations via various methods, but they should *not* actually perform transformations on the data passed through the `transform` method.  So here, when we perform normalization, that normalization applies only to the way that we visualize the data, but we are not actually returning normalized data in the `transform` method. 

It seems to me that this is the desired behavior because there are probably plenty of situations when I would want to *visualize* my data in a normalized way but I would want to build the visualization into a pipeline where the data either 1) was not normalized or 2) was normalized/preprocessed by a different method.  In other words, the normalization that makes for the best diagnostic visualization may not be the same normalization that I actually want to use for my ML model.  

From looking at the existing transformers, it seems like this is the approach for everything except the `pca` visualizer, which does return PCA-transformed features. So, up to you which way you want the API to be headed.  I kind of like keeping the visualizers as visualizers and allowing the user to do the actual transformations by other means.

pbs929: One other scenario to consider: suppose I wanted to make a VisualPipeline consisting of various feature visualizers, some of which plot the data in a normalized way and some which don't.  If the visualizer actually returns normalized data, that prevents any later visualizers in the pipeline from plotting the data in a non-normalized way, which to me seems undesirable.

yellowbrick/features/pcoords.py
bbengfort: One of the most difficult things about developing YB is figuring out the semantics between the Scikit-Learn API and the matplotlib API. I totally get what you're saying, and I'm happy to go with it to see what happens when we start deriving subclasses.

yellowbrick/features/pcoords.py
bbengfort: That's a good point - I think if we add the shuffle argument, plus a random state argument like in the [Sklearn `KFold`](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html) class then we can do both at the same time. If shuffle is True, then we're pretty close to a uniform random sample, if shuffle is False or a random state is set then we have deterministic behavior that we can test.

yellowbrick/features/pcoords.py
bbengfort: Very, very good points. 

So I do consider these feature visualizers transformers; I want them to be useful in a pipeline to show exactly what's happening to the data.  The problem is, I think that if we do normalization in the visualizer but then don't pass that on, this could also be misleading. I guess this would mean that there is an argument for simply putting the normalizer transformer ahead of the pcoords in a pipeline rather than do something mysterious on the user's behalf.

I'll consider the idea of "visualizers are visualizers" and that they stand-alone without doing transformation, we've already walked down the road of "visual transformer == visualizer" but we're not so far down it we can't walk back. In the meantime, let's get these features merged!

pbs929: Alright, sounds good :)  I realize this is getting into deeper questions of where the package is headed, which you may have more of a view of than I do.  But I agree with sentiment of let's get this merged in the meantime!

yellowbrick/features/pcoords.py
pbs929: Yeah, that's a good thought.  I think I'll mark this as a TODO for now and come back to it soon so we can get the existing work checked in.

pbs929:
Stand by for one more commit, then this should be ready to go.

bbengfort:
Ok, let me know when ready.

pbs929:
@bbengfort, sorry for the delay on this. I pushed some changes based on the discussion above and also added some tests. Should be good to go I think.

bbengfort:
@pbs929 no worries at all, thank you so much! I'm planning to review on Wednesday at the DDL office hours. Are you in DC? You're more than welcome to join - it's at Georgetown University in Chinatown.

pbs929:
Great, glad this was useful!  BTW, couldn't make it out last night, but I will definitely try to join you guys some time in the future.

bbengfort:
Awesome, we meet biweekly on Wednesdays; ping me when you're interested so we can get you on the list for security!

Issue #12, opened 2016-05-18
Grid Search Viz


bbengfort:
Create a package for visualizing grid search results. How to generalize this?

rebeccabilbro:
This one might be tricky, mainly because grid search takes so long. But here is a rough draft:

```python
def visual_gridsearch(model, X, y):
    C_range = np.logspace(-2, 10, 5)
    gamma_range = np.logspace(-5, 5, 5)
    param_grid = dict(gamma=gamma_range, C=C_range)
    grid = GridSearchCV(SVC(), param_grid=param_grid)
    grid.fit(X, y)

    scores = [x[1] for x in grid.grid_scores_]
    scores = np.array(scores).reshape(len(C_range), len(gamma_range))

    plt.figure(figsize=(8, 6))
    plt.subplots_adjust(left=.2, right=0.95, bottom=0.15, top=0.95)
    plt.imshow(scores, interpolation='nearest', cmap=cmap)
    plt.xlabel('gamma')
    plt.ylabel('C')
    plt.colorbar()
    plt.xticks(np.arange(len(gamma_range)), gamma_range, rotation=45)
    plt.yticks(np.arange(len(C_range)), C_range)
    plt.title(
        "The best parameters are {} with a score of {:0.2f}.".format(
            grid.best_params_, grid.best_score_)
    )
    plt.show()
```

TomAugspurger:
Thoughts on something [like this](http://nbviewer.jupyter.org/github/TomAugspurger/postlearn/blob/master/examples/IPywidgets.ipynb#Grid-Search-Results) for grid search? It's pretty much just a wrapper around seaborn's factorplot, pulling the data off the `GridSearch.grid_scores_`. The implementation is [here](https://github.com/TomAugspurger/postlearn/blob/1d96b8874ab6717e749d99ebaec961b8603966a6/postlearn/reporter.py#L98), from a library I started working on but haven't had a chance to develop any further.

bbengfort:
Actually, that's really helpful thanks! We're having a lot of angst around
whether to add seaborn as a dependency or to implement our own FactorPlot
somehow, but the unpack_grid_scores function is going to be really helpful!

bbengfort:
Completed in #261

PR #261, opened 2017-06-12
Grid search color plot viz (WIP)

pbs929:
Don't mean to inundate you all with PRs, but I'm having fun :)

I put together a color plot visualizer for grid search - see the ipynb for a demo.  Also aiming to create a structure into which other grid search visualizers can fit.  

Unit tests are definitely needed... I will come back to this after getting some experience with the testing workflow on my other PRs.  General feedback on the approach is welcome in the meantime. 

Called for in issue #12.


bbengfort:
@NealHumphrey do you have time to take a look at this by any chance? 

@pbs929 keep them coming!

bbengfort:
@pbs929 still interested in finishing this PR? I'm happy to do a review.

pbs929:
Yes, still interested!  As I recall I was waiting on some general feedback on the approach (see demo in the ipynb).

yellowbrick/gridsearch/pcolor.py
```diff
@@ -0,0 +1,225 @@
...
+    Parameters
+    ----------
+    cv_results : dict
+        A dictionary of results from the `GridSearchCV` object's `cv_results_`
+        attribute
+
+    param_1 : string
+        The name of the parameter to be visualized on the horizontal axis.
+
+    param_2 : string
```
bbengfort: I can see the motivation for `param_1` vs `param_2` - but because these names are connected to the vertical and horizontal axes, maybe it would be easier to name them "x_param" and "y_param" - I know this is not really the x and y axes but they do it on bar graphs so I think it would be ok here.

bbengfort: I can see the motivation for `param_1` vs `param_2` - but because these names are connected to the vertical and horizontal axes, maybe it would be easier to name them "x_param" and "y_param" - I know this is not really the x and y axes but they do it on bar graphs so I think it would be ok here.

yellowbrick/gridsearch/pcolor.py
```diff
@@ -0,0 +1,225 @@
...
+        The name of the parameter to be visualized on the vertical axis.
+
+    Returns
+    -------
+    ax : matplotlib axes
+        Returns the axes that the classification report was drawn on.
+    """
+    # Get unique values of the two display parameters
+    x_vals = sorted(list(set(cv_results['param_' + param_1].compressed())))
+    y_vals = sorted(list(set(cv_results['param_' + param_2].compressed())))
```
bbengfort: Ok, so just to make sure I understand what's happening here - you're fetching the param mask from the cv results dictionary, compressing it to get the real values instead, then calling set to find all unique values. These are the categories for the grid, right? 

I've got a couple of notes:

1. `'param_' + param_1` is used twice, it might be easier to store these in variables and note what they are or just fetch the mask array from the results and store that. 
2. If the parameter doesn't exist in the grid search, this will return a `KeyError` - it would be better if we raised a yellowbrick key error, e.g. `YellowbrickKeyError(YellowbrickError, KeyError)` that says "parameter not found in grid search" so it's more obvious to the user what happened.

yellowbrick/gridsearch/pcolor.py
```diff
@@ -0,0 +1,225 @@
...
+    idx_x = [int_mapping_1[value] if value else None
+             for value in cv_results['param_' + param_1]]
+    idx_y = [int_mapping_2[value] if value else None
+             for value in cv_results['param_' + param_2]]
+
+    # Create an array of all scores for each value of the display parameters.
+    # This is a n_x by n_y array of lists with `None` in place of empties
+    # (my kingdom for a dataframe...)
+    all_scores = [[None for _ in range(n_x)] for _ in range(n_y)]
+    for x, y, score in zip(idx_x, idx_y, cv_results['mean_test_score']):
```
bbengfort: It feels like we should allow the user to choose which metric as an input to the visualizer (which defaults to `mean_test_score` but could also be - `mean_train_score`, `mean_fit_time`, and `mean_score_time`.

yellowbrick/gridsearch/pcolor.py
```diff
@@ -0,0 +1,225 @@
...
+        if x is not None and y is not None:
+            if all_scores[y][x] is None:
+                all_scores[y][x] = []
+            all_scores[y][x].append(score)
```
bbengfort: I certainly understand the motivation for creating this intermediate data structure, but is it not possible to compute max score at this point?

yellowbrick/gridsearch/pcolor.py
```diff
@@ -0,0 +1,225 @@
...
+    "GridSearchColorPlot",
+    "gridsearch_color_plot"
+]
+
+
+##########################################################################
+## Dimension reduction utility
+##########################################################################
+
+def param_projection(cv_results, param_1, param_2):
```
bbengfort: This function seems extremely useful for all `GridSearchVisualizers` - perhaps this should be in `base.py` -- possibly as a method of `GridSearchVisualizer`?

bbengfort: Or as a standalone function, with an aliased (static?) method in `GridSearchVisualizer`?

yellowbrick/gridsearch/base.py
```diff
@@ -0,0 +1,59 @@
...
+from ..utils import is_gridsearch
+from ..base import ModelVisualizer
+from ..exceptions import YellowbrickTypeError
+
+
+##########################################################################
+## Base Grid Search Visualizer
+##########################################################################
+
+class GridSearchVisualizer(ModelVisualizer):
```
bbengfort: Very good set up of this into the Yellowbrick class hierarchy!

yellowbrick/gridsearch/base.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(self, model, ax=None, **kwargs):
+        """
+        Check to see if model is an instance of GridSearchCV.
+        Should return an error if it isn't.
+        """
+        # A bit of type checking
+        if not is_gridsearch(model):
```
bbengfort: Nice!

yellowbrick/gridsearch/pcolor.py
bbengfort: Also: future work - add more params?

yellowbrick/gridsearch/pcolor.py
```diff
@@ -0,0 +1,225 @@
...
+        self.colormap = colormap
+
+    def draw(self):
+        # Project the grid search results to 2 dimensions
+        x_vals, y_vals, best_scores = param_projection(
+            self.cv_results_, self.param_1, self.param_2
+        )
+
+        # Mask nans so that they can be filled with a hatch
+        data = np.ma.masked_invalid(best_scores)
```
bbengfort: I like this a lot! I'm going to have to use this in other things! The hatch nans thing I mean.

yellowbrick/gridsearch/pcolor.py
```diff
@@ -0,0 +1,225 @@
...
+        self.ax.patch.set(hatch='x', edgecolor='black')
+
+        # Ticks and tick labels
+        self.ax.set_xticks(np.arange(len(x_vals)) + 0.5)
+        self.ax.set_yticks(np.arange(len(y_vals)) + 0.5)
+        self.ax.set_xticklabels(x_vals, rotation=45)
+        self.ax.set_yticklabels(y_vals, rotation=45)
+
+        # Add the colorbar
+        cb = self.ax.figure.colorbar(mesh, None, self.ax)
```
bbengfort: This is a running debate that I hope you might weigh in on: should color bars have the range (min, max) or should they have the range (0, 1)? Should this be an option we supply to the user?

Arguments for (min, max):

- they create greater color disparity so it's easier to tell values apart 
- They bound the area you're looking in 

Arguments for (0,1):

- You can easily compare multiple heatmaps with each other 
- The closer min is to max, the more similar they appear in this as opposed to the other.

bbengfort: @ndanielsen @rebeccabilbro -- just wanted to highlight this discussion for you again here.

yellowbrick/gridsearch/pcolor.py
```diff
@@ -0,0 +1,225 @@
...
+        cb.outline.set_linewidth(0)
+
+        self.ax.set_aspect("equal")
```
bbengfort: Nice - glad you included this!

yellowbrick/gridsearch/base.py
pbs929: 👍

yellowbrick/gridsearch/pcolor.py
pbs929: sure, easy enough

yellowbrick/gridsearch/pcolor.py
pbs929: Alright, I moved it up to the base module and aliased it in the base GridSearchVisualizer class.  I didn't make the aliased method static, rather letting it access the gridsearch model object as `self.estimator`.

Definitely could add more params as future work.

yellowbrick/gridsearch/pcolor.py
pbs929: Yes, that's right - I've added a couple comment lines to try to clarify and stored the masked array like you suggested in (1).  Also did what you suggested in (2).

yellowbrick/gridsearch/pcolor.py
pbs929: In this case I'd advocate for min, max.  With grid search, you can use any of a number of error metrics, which may or may not be bounded by 1.  Furthermore when fine tuning parameters you may be optimizing over the 3rd or 4th decimal of the error metric, so fixing the range on a larger scale would make the visualization not very useful.

yellowbrick/gridsearch/pcolor.py
pbs929: 👍

pbs929:
Alright @bbengfort I think I've addressed your comments.  Let me know if you catch anything else - I'm pretty tied up these days with a newborn but I will try to find the time.  :)

PR #304, opened 2017-12-14
302 RadViz now works with missing data.

Aylr:
PR to start conversation per contributing guidelines.

I'm investigating where this nan/None fix check should live. Should it go in the fit method of DataVisualizer?


Aylr:
Initially, it looks like having this in `DataVisualizer.fit()` solves the problem.

screen shot 2017-12-14 at 1 31 44 am

tests/test_utils/test_nan_warnings.py
```diff
@@ -0,0 +1,77 @@
...
+    def test_raise_warning_if_nans_exist(self):
+        """
+        Test that a warning is raised if any nans are in the data
+        """
+        data = np.array([
+            [1, 2, 3],
+            [1, 2, np.nan],
+        ])
+
+        self.assertWarns(UserWarning, warn_if_nans_exist, data)
```
bbengfort: We're actually in the process of switching over to Pytest - would prefer `pytest.warns` ([assert warnings with pytest](https://docs.pytest.org/en/latest/warnings.html#warns))  -- it'll be one less thing we have to do as we move towards a complete change over!

tests/test_utils/test_nan_warnings.py
```diff
@@ -0,0 +1,77 @@
...
+            [1, np.nan, 3],
+        ])
+
+        data3 = np.array([
+            [np.nan, 2, 3],
+            [1, np.nan, 3],
+            [np.nan, np.nan, np.nan],
+        ])
+
+        self.assertEqual(0, count_nan_rows(data0))
```
bbengfort: Same here, since we're moving away from unittest, you can just do assert `count_nan_rows(data0) == 0`

yellowbrick/utils/nan_warnings.py
```diff
@@ -0,0 +1,31 @@
...
+
+def warn_if_nans_exist(data):
+    """Warn if nans exist in a numpy array."""
+    null_count = count_nan_rows(data)
+    total = len(data)
+    percent = null_count / total
+
+    # if np.isnan(data).any():
+    if null_count > 0:
+        warnings.warn(
```
bbengfort: If possible, I'd like to create a `YellowbrickWarning` class in `yellowbrick.exceptions`, then we can subclass a `DataWarning` to use here. This will allow us to more easily capture and suppress warnings if needed.

bbengfort: This is very similar to how [sklearn does things](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/exceptions.py#L40), so something like:

```python
class YellowbrickWarning(UserWarning):
    """
    Warning class used to notify users of Yellowbrick-specific issues
    """
    pass

class DataWarning(YellowbrickWarning):
    """
    There is an issue with supplied data that may produce unexpected visualizations
    """
    pass
```

yellowbrick/utils/nan_warnings.py
```diff
@@ -0,0 +1,31 @@
...
+            'Warning! Found {} rows of {} ({}%) with nan/null/None '
+            'values which are cannot be plotted.'.format(null_count, total,
+                                                         percent))
```
bbengfort: Percent is not multiplied by 100, though the format string uses `{}%`, so the result will look something like `0.3%` instead of `30%` if I'm not mistaken ... ? 

Also, with floats we should probably limit the no of decimal places, e.g. `"{:0.2f}%"`

yellowbrick/features/base.py
```diff
@@ -189,6 +191,10 @@ def fit(self, X, y=None, **kwargs):
                     str(cdx) for cdx in range(ncols)
                 ]
 
+        # Drop nan-containing rows and warn that this has taken place
+        nan_warnings.warn_if_nans_exist(X)
+        X = nan_warnings.drop_rows_containing_nans(X)
```
bbengfort: Should we allow the user to stop this behavior? E.g. have an argument to the visualizer `drop_nans=True`, that we condition on? It feels a little squishy to me to modify the user's dataset, even with a warning; but I'd like to discuss it.

yellowbrick/utils/nan_warnings.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def count_nan_rows(data):
+    """Count the number of rows that contain any nan values."""
+    if data.shape[0] >= 2:
```
bbengfort: Out of curiosity, why do we condition our count to require a minimum of two rows?

tests/test_utils/test_nan_warnings.py
```diff
@@ -0,0 +1,77 @@
...
+        Test that a warning is raised if any nans are in the data
+        """
+        data = np.array([
+            [1, 2, 3],
+            [1, 2, np.nan],
+        ])
+
+        self.assertWarns(UserWarning, warn_if_nans_exist, data)
+
+    def test_count_rows_with_nans(self):
```
bbengfort: Excellent cases - just wanted to suggest a few more; one where the # of rows is < 2 as per the condition in count_nan_rows. We should probably also send in a 1D vector as well to make sure that we can use this tool for y in addition to X?

yellowbrick/features/base.py
Aylr: Squishy is the perfect word for it! To answer you question, I need to know if I should move the nan removal deeper inside so that the modified data is not returned and still used for plotting. If so, where should I put that? If not, then I agree that the argument is a good plan.

yellowbrick/utils/nan_warnings.py
Aylr: Nice catch. This was a late night error meant to check for 2D arrays. 😄

tests/test_utils/test_nan_warnings.py
Aylr: 1. Fixed the row count (which was an error meant to check for dimensions).
2. Added a new function with a test that is for 1D arrays.

I'm going to modify the null remover to check across X and y correctly.

yellowbrick/features/base.py
Aylr: Ok. I've thought and played with this and my current preference is to leave the user's data intact and only temporarily clean it in the `RadViz.draw()` method. Now that all the functionality is in place it is trivial to move this around if you disagree or have another suggestion.

Aylr:
Thanks for the fantastic review and code snippets. I learned a lot about numpy working on this! Attached is a screenshot and source of a jupyter notebook demonstrating the fixes.

screen shot 2017-12-16 at 3 51 11 pm
[radviz_missing_values.ipynb.zip](https://github.com/DistrictDataLabs/yellowbrick/files/1565162/radviz_missing_values.ipynb.zip)

yellowbrick/features/base.py
bbengfort: @Aylr ok that sounds good to me!

tests/test_utils/test_nan_warnings.py
```diff
@@ -0,0 +1,145 @@
...
+        [4, 5, 6],
+        [7, 8, 9],
+    ])
+
+    observed = clean_data(X)
+    np.testing.assert_array_equal(X, observed)
+
+
+def test_clean_data_X_only():
+    """Test that arrays with missing data are returned without nans."""
```
bbengfort: Same as above for these docstrings, though they do actually describe what's happening!

yellowbrick/utils/nan_warnings.py
```diff
@@ -0,0 +1,49 @@
...
+        return np.where(np.isnan(X).sum(axis=1) != 0, 1, 0).sum()
+
+
+def count_nan_elements(data):
+    """Count the number of elements in 1D arrays that are nan values."""
+    if data.ndim == 1:
+        return np.isnan(data).sum()
+
+
+def clean_data(X, y=None):
```
bbengfort: I'll be honest, I'm not 100% sure about the name here; when used in combination with the module, nan_warnings.clean_data it seems to make sense; but elsewhere I think it might not be communicative. Perhaps, `filter_missing`, `remove_nan_rows` or `clean_nans` would be better? 

Also, I think this function will get used in a few places, so perhaps it would be good to expand the docstring? Something like:

```python 
def clean_data(X, y=None):
    """
    Removes rows that contain np.nan values in data. If y is given, 
    this function ensures that X and y will be filtered together so that 
    their shape remains identical, e.g. rows in X with nans will also 
    remove rows in y, or rows in y with np.nans will also remove 
    corresponding row in X. 

    Parameters
    ------------
    X : array-like 
        Data in shape (m, n) that possibly contains np.nan values 

    y : array-like, optional 
        Data in shape (m, 1) that possibly contains np.nan values 

    Returns
    --------
    X' : np.array 
       Possibly transformed X with any row containing np.nan removed 

    y' : np.array 
        If y is given, will also return possibly transformed y to match the 
        shape of X'. 

    Notes
    ------ 
    This function will return either a np.array if only X is passed or a tuple 
    if both X and y is passed. Because all return values are indexable, it is
    important to recognize what is being passed to the function to determine
    its output. 
    """
```

I know that's a mouthful - feel free to edit as you see fit. We can then add it to the documentation; perhaps in the next PR?

bbengfort:
@Aylr Thanks so much for your hard work on this; it's a very good contribution!

tests/test_utils/test_nan_warnings.py
```diff
@@ -0,0 +1,145 @@
...
+        [1, 2, 3],
+        [1, 2, np.nan],
+    ])
+
+    with pytest.warns(DataWarning):
+        warn_if_nans_exist(data)
+
+
+def test_count_rows_in_2d_arrays_with_nans():
+    """Test that a warning is raised if any nans are in the data."""
```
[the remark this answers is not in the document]
Aylr: This isn't trivial at all. Great catch!

tests/test_utils/test_nan_warnings.py
```diff
@@ -0,0 +1,145 @@
...
+    data0 = np.array([
+        [1, 2, 3],
```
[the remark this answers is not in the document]
Aylr: Can you please clarify this - I'm not following you here.

yellowbrick/utils/nan_warnings.py
Aylr: 1. Thanks for being honest! I totally agree on the name being unclear. :smile:
> There are only two hard things in Computer Science: cache invalidation and naming things.
> -- Phil Karlton

2. I'll add the nice docstring.

Aylr: Thinking more about this main "switch" function, I might prefer to remove it (and simply make two functions) so that reasoning about different returns is eliminated.

tests/test_utils/test_nan_warnings.py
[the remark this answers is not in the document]
bbengfort:
```
data0 = np.array([
    [1,2,3],
    [4,5,6], 
]) 
``` 

I've forgotten what I thought the bug would be with only one row, but I think having a 2x3 array would be a slightly better test and protect us from injecting whatever bug I was thinking of in the future.

Aylr:
@bbengfort thank you for the great insights, comments and questions! Experiences like this are why I love open source.

bbengfort:
@Aylr cheers - thanks again for contributing, we'd love for you to participate as much as you're able!

PR #309, opened 2018-02-09
Features refactor

pbs929:
Back in #258, I removed a TODO about hoisting more shared functionality from the `features` visualizers higher in the inheritance hierarchy, thinking it was already done.  On closer inspection, there was a bit more to do; this PR is my attempt at that.  

The main observation here was that both the `DataVisualizer` base class and the `RankDBase` class did identical extraction of feature names from dataframe columns.  I have created a new intermediary base class `MultiFeatureVisualizer` from which both can inherit this functionality.  To sum up the inheritance structure for this subpackage, we now have:
* `FeatureVisualizer` - inherits from `base.Vizualizer` and `TransformerMixin`, provides `transform` and `fit_transform_poof` methods.  `PCADecomposition` and `JointPlotVisualizer` inherit directly from this.   
* `MultiFeatureVisualizer(FeatureVisualizer)` - provides `self.features_` attribute as well as a `fit` method that extracts feature names from dataframe columns if necessary.  `RankDBase` inherits directly from this. 
* `DataVisualizer(MultiFeatureVisualizer)` - provides `self.classes_` attribute and extends the `fit` method to extract the classes from `y` data if necessary.  `ScatterVisualizer`, `ParallelCoordinates`, and `RadialVisualizer` inherit from this.  (Side note: IMO this class should be renamed to ClassVisualizer, as DataVisualizer is a bit too general.)

While testing out the refactoring above, I attempted to put all these visualizers into a visual pipeline and noticed a few small problems that I have also taken the liberty of addressing. 
* I added the PCA class and quick method to the imports in the `__init__.py`.
* `FeatureVisualizer` defined a `fit` method that was identical to the one it already inherited from `base.Visualizer`, which I removed for clarity.  
* The constructor for `PCADecomposition` took X and y arguments and set a `self.pca_features_` attribute that were not used for anything, which I removed. 
* `RadialVisualizer` did not work with dataframe inputs, so I fixed this to be similar to the other visualizers. 
* The constructors for `Rank1D` and `Rank2D` did not pass on their `ax` arguments to their super-methods, which I fixed. 
* The constructor for ScatterVisualizer was not setting the `color` and `colormap` attributes from the arguments that were passed in, even though these attributes were used by the `draw` method.  

The included jupyter notebook shows that after the described changes, the visualizers now all work as part of a `VisualPipeline`.  (Except for `JointPlotVisualizer`, which requires X to be a vector/series, making it an outlier here... I may open an issue to address this.)

@bbengfort since you reviewed my previous work on this subpackage - LMK if this looks good or if I'm overstepping bounds on anything.


pbs929:
BTW, the commit history has a bunch of detritus from my previous PRs (already merged) and my git foo is not good enough to squash them... the last 5 commits are the applicable ones.

bbengfort:
Thanks @pbs929 for the pull request -- things are a bit busy right now, but I'll review as soon as I can - potentially later this week. Hope that's ok.

pbs929:
No rush on my end - just wrapping this up into a PR since it had been sitting in a local branch on my machine for 6 months.  :)

yellowbrick/features/base.py
```diff
@@ -42,16 +42,6 @@ class FeatureVisualizer(Visualizer, TransformerMixin):
     def __init__(self, ax=None, **kwargs):
         super(FeatureVisualizer, self).__init__(ax=ax, **kwargs)
 
-    def fit(self, X, y=None, **fit_params):
```
bbengfort: This is the unnecessary extra fit method removed for clarity, right?

yellowbrick/features/base.py
```diff
@@ -75,11 +65,68 @@ def fit_transform_poof(self, X, y=None, **kwargs):
         return Xp
 
 
+class MultiFeatureVisualizer(FeatureVisualizer):
```
bbengfort: So one point of feedback I've been getting is that the inheritance chain is too complicated and it's hard to reason about what methods are doing what. Not necessarily for this PR, but perhaps we could open an issue to create a series of Mixins that do the work we're describing for things like this.

yellowbrick/features/base.py
```diff
@@ -75,11 +65,68 @@ def fit_transform_poof(self, X, y=None, **kwargs):
...
+        If a DataFrame is passed to fit and features is None, feature
+        names are selected as the columns of the DataFrame.
+
+    kwargs : dict
+        Keyword arguments that are passed to the base class and may influence
+        the visualization as defined in other Visualizers.
+
+    """
+
+    def __init__(self, ax=None, features=None, **kwargs):
```
bbengfort: I would prefer if features was renamed to labels to be more consistent with Scikit-Learn. This probably involves cleaning up some other code. Also, labels should also accept a LabelEncoder. 

You don't have to do it in this PR, but would you mind opening up an issue with this and mention the locations of where things might have to change?

yellowbrick/features/radviz.py
```diff
@@ -159,6 +160,9 @@ def draw(self, X, y, **kwargs):
         draws each instance as a class or target colored point, whose location
         is determined by the feature data set.
         """
+        # Convert from dataframe
+        if is_dataframe(X):
```
bbengfort: Scikit-Learn has some utilities called `_check_target` and `_check_instances` that do this kind of conversion. This is just a general comment to remind me to think about how to deal with this in a more meaningful way in the future.

yellowbrick/features/rankd.py
```diff
@@ -383,7 +343,7 @@ def __init__(self, ax=None, algorithm='shapiro', features=None,
         order features as well as visualize the result.
         """
         super(Rank1D, self).__init__(
-            ax=None, algorithm=algorithm, features=features,
+            ax=ax, algorithm=algorithm, features=features,
```
bbengfort: Wow, great catch!

Issue #297, opened 2017-11-30
Check if an estimator is fitted and visualize without modification


bbengfort:
Don't call fit on an already fitted estimator.

`ModelVisualizer` objects should be able to accept either a fitted or unfitted model; we shouldn't call fit on the model if it's already fitted, especially if `fit()` is a hook for drawing. 

Moreover there are a series of visualizers that use `fit()` to draw themselves but only need information from the model, there should be a shortcut function for drawing these things (unless they need data).

dylanking42:
Is this related to the sklearn.utils.validcation.check_is_fitted?
http://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html

SO discussion on related topic:
https://stackoverflow.com/questions/39884009/whats-the-best-way-to-test-whether-an-sklearn-model-has-been-fitted

check is fitted will raise an error if a model IS NOT fitted, but maybe it could be use to bypass fit() within ModelVisualizer?

bbengfort:
@dylanking42 that's a great idea - I think `check_is_fitted` is exactly what we need for this! Did you want to submit a PR that fixes this?

dylanking42:
If I knew how I would have already done it -_-

bbengfort:
Totally understood, no worries! The links and notes are very helpful!

bbengfort:
The check fitted method requires params to be known in advance. We can have visualizers specify those params using an internal `_is_fitted` or `fit_params` mechanism. We should also capture `sklearn.exceptions.NotFittedError` and raise `yellowbrick.exceptions.NotFitted` instead. 

Here is some sample utility code from a previous project (though this needs to be refined a lot:

```
import sklearn 
import warnings 
from sklearn.utils.validation import check_is_fitted

def is_fitted(estimator):
    if hasattr(estimator, '_is_fitted'):
         return estimator._is_fitted() 

    warnings.warn("performing a broad search of fit params", UserWarning) 
    try:
         check_is_fitted(estimator, ["coef_", "estimator_"], all_or_any="any")
         return True 
    except sklearn.exceptions.NotFittedError:
          return False 
```

bbengfort:
This is also related to quick methods which would benefit from being able to draw an already fitted model without having to pass train-test splits directly into the method.

amueller:
[...]

PR #317, opened 2018-03-03
FeatureImportances Visualizer

bbengfort:
In this PR I've implemented the `FeatureImportances` visualizer as described by #58 and as loosely implemented in #195. This visualizer uses both `feature_importances_` and `coef_` on a fitted model to create a bar chart with each feature's relative importance. Important features include:

- feature discovery: can pass a list of labels or will attempt to infer features from data 
- ranking: fitted params `features_` and `feature_importances_` are sorted by importance 
- relative: display in two modes either as a percentage relative to max or actual values 

One thing we should discuss, this is a `ModelVisualizer` but I've placed it in `yellowbrick.features`. This feels ok to me even though it's a little weird since it's really about feature engineering and because it doesn't really fit into either classifiers or regressors since it can do both. 

Features and tasks TODO:

- [ ] implement standard deviation for ensemble models 
- [x] create absolute mode for coefs that can be negative 
- [ ] top n features to filter number displayed (both pos and neg)
- [ ] color negative coefs 
- [x] write documentation 

Note that tests are written for this, including image comparison tests! 

Here is a classifier on the occupancy dataset using all default values:

feature_importances_

And here is a regressor on the concrete dataset with relative=False and named features:

coef_


ndanielsen:
Will review tomorrow

bbengfort:
oops, struggles with merges master and develop -- should be ok; but there are more commit messages than I wanted.

yellowbrick/exceptions.py
```diff
@@ -39,6 +39,13 @@ class ModelError(YellowbrickError):
     pass
 
 
+class NotFitted(ModelError):
```
ndanielsen: great addition

yellowbrick/features/base.py
```diff
@@ -104,21 +106,17 @@ def fit(self, X, y=None, **fit_params):
...
-        # Get the shape of the data
-        nrows, ncols = X.shape
-
         # Handle the feature names if they're None.
         if self.features_ is None:
 
             # If X is a data frame, get the columns off it.
             if is_dataframe(X):
-                self.features_ = X.columns
+                self.features_ = np.array(X.columns)
```
ndanielsen: do we want to put support in here for structured / named arrays here at some point?

yellowbrick/features/importances.py
```diff
@@ -0,0 +1,250 @@
...
+            self.feature_importances_ *= 100.0
+
+        # Create labels for the feature importances
+        # NOTE: this code is duplicated from MultiFeatureVisualizer
+        if self.labels is None:
+            # Use column names if a dataframe
+            if is_dataframe(X):
+                self.features_ = np.array(X.columns)
+
+            # Otherwise use the column index as the labels
```
ndanielsen: do we want to add support for structured arrays here?

yellowbrick/features/importances.py
```diff
@@ -0,0 +1,250 @@
...
+        return self
+
+    def draw(self, **kwargs):
+        """
+        Draws the feature importances as a bar chart; called from fit.
+        """
+        # Quick validation
+        for param in ('feature_importances_', 'features_'):
+            if not hasattr(self, param):
+                raise NotFitted("missing required param '{}'".format(param))
```
ndanielsen: perhaps we should put this validations in __init__?

yellowbrick/features/importances.py
```diff
@@ -0,0 +1,250 @@
...
+            "could not find feature importances param on {}".format(
+                self.estimator.__class__.__name__
+            )
+        )
+
+    def _is_fitted(self):
+        """
+        Returns true if the visualizer has been fit.
+        """
+        return hasattr(self, 'feature_importances_') and hasattr(self, 'features_')
```
ndanielsen: maybe this should be moved into utils.helpers.py

tests/test_features/test_importances.py
```diff
@@ -0,0 +1,326 @@
...
+
+    def test_integration_coef(self):
+        """
+        Integration test of visualizer with coef param
+        """
+
+        concrete = self.load_data('concrete')
+        feats = ['cement','slag','ash','water','splast','coarse','fine','age']
+
+        # TODO: find some way to convert this structured array sanely
```
ndanielsen: 👍

yellowbrick/features/importances.py
```diff
@@ -0,0 +1,250 @@
...
+        Keyword arguments that are passed to the base class and may influence
+        the visualization as defined in other Visualizers.
+
+    Examples
+    --------
+
+    >>> from sklearn.ensemble import GradientBoostingClassifier
+    >>> visualizer = FeatureImportances(GradientBoostingClassifier())
+    >>> visualizer.fit(X, y)
+    >>> visualizer.poof()
```
ndanielsen: a thought, do we want the fit method to call poof()? it would be one less step to produce the visualization

yellowbrick/features/base.py
bbengfort: Yes, we probably do; man, are they a pain in the butt though. I pulled out all of the code we had that used `X.view` since we were getting massive numbers of deprecation warnings from Numpy and I replaced it with `np.array(X.tolist())` which isn't great. 

StackOverflow post on the subject here if you have an answer: https://stackoverflow.com/questions/49085469/convert-structured-array-to-numpy-array-for-use-with-scikit-learn

yellowbrick/features/importances.py
bbengfort: I thought about it, however, those params won't appear on the estimator until after they're fitted - so we can't do anything until then. 

I wanted some advice on this though - it would be nice if we could take an already fitted model and draw the feature importance without having to run through the data again.  Any thoughts on how to incorporate this here?

yellowbrick/features/importances.py
bbengfort: That's a good thought, the problem is that poof calls `plt.show()` and we need it to not do that for image comparison and other reasons. However, poof also calls `finalize` and I'm thinking that should be part of `draw()` not part of poof. I'll think about how to do this.

yellowbrick/features/importances.py
bbengfort: Yeh, this is something I do at work with the estimators I make. Basically, this is an estimator-specific definition of what being fitted means. I do have utility code that goes something like this:

```python
import sklearn 
import warnings 
from sklearn.utils.validation import check_is_fitted

def is_fitted(estimator):
    if hasattr(estimator, '_is_fitted'):
         return estimator._is_fitted() 

    warnings.warn("performing a broad search of fit params", UserWarning) 
    try:
         check_is_fitted(estimator, ["coef_", "estimator_"], all_or_any="any")
         return True 
    except sklearn.exceptions.NotFittedError:
          return False 
```

The problem is that you have to guess what the internal params are for `check_is_fitted` to work. I'm all for bringing this into Yellowbrick though, I'll do another PR that basically wraps the sklearn method and raises a yellowbrick exception.

yellowbrick/exceptions.py
rebeccabilbro: +1

yellowbrick/features/importances.py
rebeccabilbro: Maybe we want to add a `fit_poof` method similar to sklearn's `fit_transform`?

yellowbrick/features/importances.py
rebeccabilbro: It would be nice to have some kind of general utility to use in the score and model visualizers that would check to see if the model is fitted already, and if it is, return some useful information (e.g. hyperparam settings, coefficients/feature importances) that could be used by the visualizers.

yellowbrick/features/base.py
bbengfort: Added issue #319 in response

yellowbrick/features/importances.py
bbengfort: Updated #297 in response to these comments.

yellowbrick/features/importances.py
bbengfort: That's a good idea, I thought we had that but they don't seem to be there anymore.

bbengfort:
Ok, I've added documentation. I just wanted to document how I did the documentation. 

The basic documentation process is as follows: 

1. Create .rst file in specified `docs/api` folder 
2. Create a .py file in `docs/api` folder to generate images 
3. Generate images and write API document 
4. Review the code documentation and ensure `automodule` is set up correctly 

Index the document by adding references in the following locations:

- `docs/api` folder index.rst (ensure inclusion in TOC) and listing
- `docs/index.rst` for high level over view of visualizers 
- DESCRIPTION.rst (for PyPI)
- README.md (for GitHub)

So that's a lot to do and we should probably put it in `CONTRIBUTORS.md` if it's not already.

bbengfort:
I'm going to implement the last few features then I'm going to squash and merge. I think the current documentation needs some editing; but I'll pass that off on the research labs folks ;-) 

Let me know if you guys have any other thoughts.

yellowbrick/features/base.py
ndanielsen: well aware of how painful they can be to work with =)

PR #321, opened 2018-03-05
Class Balance Heat Map Visualizer

lwgray:
In this PR I've implemented a ClassBalanceHeatMap visualizer based on @bbengfort  prototype code .  I was asked by @rebeccabilbro to take a crack at this implementation.  

This visualizer is a heatmap implementation of the class balance visualizer.  It is being added because the heatmap provides a way to quickly understand how good your classifier is at predicting the right classes.

Take for instance,  the sklearn digits data set (see below) with 10 classes.  The heatmap displays data as a stacked bar chart.  On the X axis are the different classes. The Y axis is a count of how many times your model predicted a certain class.  For example... The second bar shows the total number of instances where a Class 1 prediction was assigned.  Additionally, each bar is color coded  to show the number of predicted class.  In the case of the second bar, it shows three colors with the following counts, Class 1(Purple, 50), Class 8 (Red, 3), Class 9(Grey, 1).  This means that of the total number of "Class 1" predictions, 92.5% were actually "Class 1", 5.5% were "Class 8" and 2% were "Class 9"

heatmap

#### Major Issue
If I build the visualizer in a python notebook as seen [here](https://github.com/lwgray/ddl/blob/master/yellowbrick/heatmap.ipynb) it works.  

However, if I write the visualizer into yellowbrick/classifier/class_balance_heat_map.py (in this pull request) I receive a RuntimeError: Invalid DISPLAY variable.  I can provide the traceback if required.

#### To Do

- [x] Graphics - 
**Strike throughs are being left as user-defined parameters**
    - [x] ~~Fix axis labels~~
    - [x] ~~Fix axis ticks~~
    - [x] ~~Draw box around legend~~
    - [x] ~~Get rid of grid lines in Figure~~
    - [x] Set ylim of chart so that there exists space between bars and chart border

- [ ] Tests
    - [ ] Test1: Test that indices and classes are the same length.  Here are two instances to test
        - [ ] This would be the case where more labels are supplied in our test data, via score than in our training data; 
        - [ ] if there are less labels in y and y_pred, that's fine - we just need to ensure there are zeros for the fit classes that do not have any associated values
- [ ] Documentation
     - [x] Create Example Notebook
     - [ ] Review docstrings - Bad wording
     - [ ] class_prediction_error.py/rst doc files
- [ ] Issues
    - [x] Open issue regarding self.score_ and self.classes_ nomenclature
           throughout yellowbrick.  see @bbengfort @ndanielsen comments below 
    - [ ] Open up an issue to create the target module as per @rebeccabilbro
- [x] Misc
    - [x] Correct RuntimeError
    - [x] Perform _check_target on y and y_pred - raise YellowbrickValueError


lwgray:
### Side Note / RuntimeError Correction
I can correct the RuntimeError if I include "%matplotlib inline" in jupyter notebook

```python
import matplotlib
from yellowbrick.classifier import ClassBalanceHeatMap
from sklearn.datasets import load_digits
from sklearn.ensemble import RandomForestClassifier
%matplotlib inline

digits = load_digits()
X_train, X_test, y_train, y_test = tts(digits.data, digits.target, test_size=0.33, random_state=42)
visualizer = ClassBalanceHeatMap(RandomForestClassifier())
visualizer.fit(X_train, y_train)
visualizer.score(X_test, y_test)
visualizer.poof(outpath='heatmap.png')
```
heatmap2

bbengfort:
@lwgray in terms of the error you mentioned, this has to do with your operating system and the environment you're running matplotlib in. See [RuntimeError: Invalid DISPLAY variable](https://stackoverflow.com/questions/35737116/runtimeerror-invalid-display-variable) for more details. Are you running on Ubuntu or Windows? Needless to say, using `%matplotlib inline` or better yet, `%matplotlib notebook` sets the matplotlib backend to the correct display mode. 

I'm going to go ahead and do a brief code review of the current pull request -- you should see that momentarily!

bbengfort:
By the way - very good PR intro text and thanks for including an image!

yellowbrick/classifier/class_balance_heat_map.py
```diff
@@ -0,0 +1,169 @@
+# yellowbrick.classifier.class_balance_heat_map
```
bbengfort: I think one thing we need to do is have a discussion of whether this is its own visualizer or if it should simply extend the existing `ClassBalance` visualizer. Either way, I don't believe this needs to be in a separate file, I recommend adding it to the `yellowbrick/classifier/class_balance.py` module. 

This makes things a bit easier for you as well since the documentation and test files are already set up!

yellowbrick/classifier/class_balance_heat_map.py
```diff
@@ -0,0 +1,169 @@
...
+# Author:   Rebecca Bilbro <rbilbro@districtdatalabs.com>
+# Author:   Benjamin Bengfort <bbengfort@districtdatalabs.com>
+# Author:   Neal Humphrey
+# Author:   Larry Gray
+# Created:  Sun March 4 20:45:00 2018 -0400
+#
+# Copyright (C) 2017 District Data Labs
+# For license information, see LICENSE.txt
+#
+# ID: class_balance_heat_map.py [5388065] lwgray@gmail.com $
```
bbengfort: Don't worry about changing this line - it is automatically updated with a git hook when we push to master during releases!

yellowbrick/classifier/class_balance_heat_map.py
```diff
@@ -0,0 +1,169 @@
+# yellowbrick.classifier.class_balance_heat_map
+# Class balance visualizer for showing per-class support.
+#
+# Author:   Rebecca Bilbro <rbilbro@districtdatalabs.com>
+# Author:   Benjamin Bengfort <bbengfort@districtdatalabs.com>
+# Author:   Neal Humphrey
+# Author:   Larry Gray
```
bbengfort: Very good - adding your name here is the right thing to do!

yellowbrick/classifier/class_balance_heat_map.py
```diff
@@ -0,0 +1,169 @@
...
+
+from sklearn.model_selection import train_test_split
+from sklearn.utils.multiclass import unique_labels
+
+
+##########################################################################
+## Class Balance Heat Map Chart
+##########################################################################
+
+class ClassBalanceHeatMap(ClassificationScoreVisualizer):
```
bbengfort: If we do decide to make this a separate class, I recommend that we make it a subclass of `ClassBalance` 

Also, I'd like to talk about the name - this is technically a heat map since it is using different colors to represent data, however, I'm afraid that it implies that the strength or hue of the color implies the value's magnitude. Because we're talking about classes, they are discrete. 

For me, programming design is about the naming of things; once you nail the name function follows. If we can't think of a good name for this, then I believe that strongly implies we make this an option of `ClassBalance` rather than an independent visualizer. 

My suggested names are:

- `PredictionDistribution` 
- `PredictionBalance` 
- `ClassPredictionError` 

Thoughts? Do any of these things jump out at you?

yellowbrick/classifier/class_balance_heat_map.py
```diff
@@ -0,0 +1,169 @@
...
+    kwargs: dict
+        Keyword arguments passed to the super class. Here, used
+        to colorize the bars in the histogram.
+    Notes
+    -----
+    These parameters can be influenced later on in the visualization
+    process, but can and should be set as early as possible.
+    """
+
+    def score(self, X, y=None, **kwargs):
```
bbengfort: As part of the discussion, one thing I will say is that I think `ClassBalance` is actually a feature visualizer, or maybe something new... a target visualizer! Having `score` as the entry point always seemed a little too far downstream, as did involving a model, since knowledge of the class balance is usually a preliminary to selecting a model. 

This version, however, is definitely a `ModelVisualizer` and its entry point is definitely `score` (though you could also make a case for `predict`.

yellowbrick/classifier/class_balance_heat_map.py
```diff
@@ -0,0 +1,169 @@
...
+            A matrix of n instances with m features
+        y : ndarray or Series of length n
+            An array or series of target or class values
+
+        Returns
+        -------
+
+        ax : the axis with the plotted figure
+        """
+
```
bbengfort: Would you mind adding a comment here:

```python 
# We're relying on predict to raise NotFitted
y_pred = self.predict(X)
```

yellowbrick/classifier/class_balance_heat_map.py
```diff
@@ -0,0 +1,169 @@
...
+        y_pred = self.predict(X)
+        self.indices = unique_labels(y, y_pred)
+        self.classes_ = self.classes_ if self.classes_ is not None else self.indices
```
bbengfort: I would advise not storing indices on `self` -- changing this line to:

```
indices = unique_labels(y, y_pred)
if self.classes_ is None:
    self.classes_ = indices 
```

Estimator attributes should either be hyperparameters (supplied by the user and having reasonable defaults) or parameters, things that are added during `fit` and are [suffixed by an underscore](http://scikit-learn.org/stable/tutorial/statistical_inference/settings.html#estimators-objects). 

However, I'm not sure if we actually do need to do anything about `self.classes_` - in what cases would we fit and `self.classes_` is None? Instead, this should probably be a check to align indices with `self.classes_` -- if there is a mismatch we need to raise an error ourselves rather than raising an uninterpretable error. (This would be the case where more labels are supplied in our test data, via score than in our training data; if there are less labels in y and y_pred, that's fine - we just need to ensure there are zeros for the fit classes that do not have any associated values). 

I'd also recommend we add the above two cases to the tests for this.

yellowbrick/classifier/class_balance_heat_map.py
```diff
@@ -0,0 +1,169 @@
...
+        self.scores = np.array([
```
bbengfort: Following the comment above, I'd say it is valid to store this as `self.scores_` -- many of the `*CV` estimators do something like this, though they compute the scores on fit. I think we can get away with that here. 

That also reminds me, there are several other visualizers that are not consistent with this behavior and need to be updated. @ndanielsen, @rebeccabilbro  - do you think we should make this an issue to deal with it in the future?

yellowbrick/classifier/class_balance_heat_map.py
```diff
@@ -0,0 +1,169 @@
...
+    def draw(self):
+        """
+        Renders the class balance heap map across the axis.
+        Returns
+        -------
+        ax : the axis with the plotted figure
+        """
+
+        prev = np.zeros(len(self.classes_))
+        colors = [cm.spectral(x) for x in np.linspace(0, 1, len(self.indices))]
```
bbengfort: I had advised you not to store `self.indices` above, and that would definitely cause a problem here! During score you validated the classes in `y_true` and `y_pred` so at this point, indices can simply be `np.arange(len(self.classes_))`, which is easy enough to recompute in the draw function.

bbengfort: Additionally, we should allow the user to specify the color palette rather than hardcoding `cm.spectral`. See `resolve_colors` for more on this.

rebeccabilbro:
@bbengfort @ndanielsen @NealHumphrey @jkeung @lwgray - my sense is that both `ClassBalance` and `ClassBalanceHeatMap` should be `DataVisualizers` (subclass of `Visualizer`), housed in a module separate from `ModelVisualizers` and `FeatureVisualizers`, focused purely on target visualization.

yellowbrick/classifier/class_balance_heat_map.py
```diff
@@ -0,0 +1,169 @@
...
+        ax : the axis with the plotted figure
+        """
+
+        y_pred = self.predict(X)
+        self.indices = unique_labels(y, y_pred)
+        self.classes_ = self.classes_ if self.classes_ is not None else self.indices
+        self.scores = np.array([
+            [(y_pred[y == label_t] == label_p).sum()
+            for label_p in self.indices]
+            for label_t in self.indices
```
ndanielsen: minor request to break this up nested list comprehension up as it is a bit hard to read. at a minimum, a nice comment explaining it would be great

yellowbrick/classifier/class_balance_heat_map.py
```diff
@@ -0,0 +1,169 @@
...
+    Returns
+    -------
+    ax : matplotlib axes
+        Returns the axes that the class balance plot was drawn on.
+    """
+    # Instantiate the visualizer
+    visualizer = ClassBalanceHeatMap(model, ax, classes, **kwargs)
+
+    # Create the train and test splits
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2)
```
ndanielsen: adding test_size to the function kwargs would allow users to specify their test_size

yellowbrick/classifier/class_balance_heat_map.py
ndanielsen: I agree that we need to make an issue, and likely conduct an audit to ensure that all visualizers are conforming to this spec for self.score_ but also for things like self.classes_

lwgray:
Thanks for reviewing so quickly; there's no need to thank me - you did all the leg work.  I will try my best to expediently address your comments ..

yellowbrick/classifier/class_balance_heat_map.py
lwgray: I'm split between ClassPredictionError and PredictionDistribution- but I think the former is a more informative name.

yellowbrick/classifier/class_balance_heat_map.py
lwgray: I agree, I found it weird that I couldn't get class information until the very end of my workflow.

yellowbrick/classifier/class_balance_heat_map.py
lwgray: 1. I was not sure what exception to raise, but I choose to go with ModelError
2. Also, I wan't sure what message to to raise with Error

lwgray: @bbengfort if i don't store indices on self, the the scope of the indices variable is limited to the score function and can't be accesses by draw() Is there a work around for this?

lwgray: @bbengfort disregard my last comment, you addressed it below

yellowbrick/classifier/class_balance_heat_map.py
lwgray: I have run into an issue with the colors .  @bbengfort you asked that I not hard code in cmap thus I wrote this snippet(see below) and I ran the code in jupyter notebook (code below).  I have two different outcomes depending on whether or not I specify a color in the visualizer.  If I don't specify a color,  I want to display a default color scheme.  However, I get an IndexError when no color is specified.

``python
```
if self.color is not None:
    colors = resolve_colors(
                  color=self.color,
                  num_colors=len(self.classes_))
else:
    # colors = [cm.spectral(x) for x in np.linspace(0, 1, len(indices))]
    colors = get_color_cycle()
```

digits = load_digits()
X_train, X_test, y_train, y_test = tts(digits.data, digits.target, test_size=0.33, random_state=42)
visualizer = ClassBalanceHeatMap(RandomForestClassifier(), color='bgrmyck')
visualizer.fit(X_train, y_train)
visualizer.score(X_test, y_test)
visualizer.poof()


### Specify Color: BGRMYCK
heatmap_rgb

### No Color Specified: Default scheme
heatmap_indexerror

yellowbrick/classifier/class_balance_heat_map.py
lwgray:
```python
def class_balance_heat_map(model, X, y=None, ax=None, classes=None,
                             test_size=0.2, **kwargs):      
      # Instantiate the visualizer
      visualizer = ClassBalanceHeatMap(model, ax, classes, **kwargs)

      # Create the train and test splits
      X_train, X_test, y_train, y_test = train_test_split(X, y,
                                                          test_size=test_size)

      visualizer.fit(X_train, y_train, **kwargs)
      visualizer.score(X_test, y_test)

      # Return the axes object on the visualizer
      return visualizer.ax
```

yellowbrick/classifier/class_balance_heat_map.py
lwgray: I am working on it... it was left over from @bbengfort  and I am still having trouble decoding it.

lwgray:
I finished all the corrections except one... I was wondering if a consensus was reached on how to treat this visualizer?

yellowbrick/classifier/class_balance_heat_map.py
bbengfort: scikit-learn does some target verification in [`_check_targets`](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/metrics/classification.py#L44) 

Interestingly it does not necessarily do what I suggested above, but perhaps it is sufficient to just run `y` and `y_pred` through that check. They raise `ValueError` when things are off, I would prefer to raise `MantleValueError` if possible.

yellowbrick/classifier/class_balance_heat_map.py
bbengfort: Ok - we'll have to make an issue for it; @lwgray would you mind doing that?

yellowbrick/classifier/class_balance_heat_map.py
bbengfort: This is setting up a two-dimensional array, so I think it's going to be tough to break it up. Might I suggest the following?

```python
# Create a table of scores whose rows are the true classes
# and whose columns are the predicted classes; each element
# is the count of predictions for that class that match the true 
# value of that class. 
self.scores_ = np.array([
    [
        (y_pred[y == label_t] == label_p).sum()
        for label_p in self.indices
    ]
    for label_t in self.indices
])
```

yellowbrick/classifier/class_balance_heat_map.py
bbengfort: I'm pretty sure you can pass None into `resolve_colors` and it will give you the default color cycle for the specified number of classes. Would you mind trying that?

NealHumphrey:
@bbengfort and @rebeccabilbro @lwgray @ndanielsen @jkeung 

A brief but related digression - one thing that has always bothered me about the plain `ClassBalance` visualizer is that it uses `precision_recall_fscore_support` and then only uses the `.support` portion of that to get the class balance. Technically to just draw the class balance all you need is to bin the `y_actual` into their classes - no fitted model needed, and in fact the same class balance would apply across models. My vote would be to rewrite the regular `ClassBalance` visualizer to be based purely on the data itself, which also opens it up to being used on training data not just test data.

In any case `ClassPredictionError` visualizer (i.e. this new one by @lwgray) clearly needs a fitted model, so  I agree with your suggestion to keep it as a new visualizer in `class_balance.py`

yellowbrick/classifier/class_balance_heat_map.py
lwgray: @bbengfort :+1:

bbengfort:
@NealHumphrey yeh I agree and so does @rebeccabilbro -- I think we'll create a `target` module to go along with the `features` module to do this. 

For the current class balance, there is a score version that shows the balance of y_true vs the balance of y_pred, though it would need to be a CV scoring method, not a single split method:

https://gist.github.com/bbengfort/bd524672aff751f4340be58833f256ec

Which produces:

cb_preds

@lwgray check out the above gist, it has some of the _check_target and validation stuff I mentioned in earlier comments (though it is not tested prototype code and I've noted a few places where some assumptions are made).

yellowbrick/classifier/class_balance_heat_map.py
lwgray: I think resolve_colors func doesn't work as expected.   If both color and colormap are set to None then invoking resolve_colors triggers a TypeError.


```python

def resolve_colors(num_colors=None, colormap=None, color=None):
            colormap = cm.get_cmap(colormap)

    # Work with the colormap
    if color is None and colormap is None:
        if isinstance(colormap, str):
            cmap = colormap
            colormap = cm.get_cmap(colormap)

            if colormap is None:
                raise YellowbrickValueError(
                    "Colormap {0} is not a valid matploblib cmap".format(cmap)
                )

        colors = list(map(colormap, np.linspace(0, 1, num=num_colors)))

    # Work with the color list
    elif color is not None:

        if colormap is not None:
            warnings.warn(
                "'color' and 'colormap' cannot be used simultaneously! Using 'color'."
            )

        colors = list(color) # Ensure colors is a list

    # Get the default colors
    else:
        colors = get_color_cycle()

    if len(colors) != num_colors:
        multiple = num_colors // len(colors) - 1
        mod = num_colors % len(colors)
        colors += multiple * colors                                                                                          colors += colors[:mod]

    return colors
```

bbengfort: Whoops, I guess I was wrong about that ... thanks for checking though. Ok, so there is a way to get the default colors and I'm going to have to search for it. One sec ...

bbengfort: Ok, after much investigation, it does look like it's supposed to be `resolve_colors` but the function isn't working. I tried to change it but it broke some tests so I'll need more time to get a fix going. I'll work on it and ping you when I do the PR to resolve it. 

In the meantime, perhaps try:

```
# HACK: Waiting for @bbengfort to fix resolve_colors 
color_values = get_color_cycle()
n_colors = len(color_values) 
colors = [color_values[i % n_colors] for i in np.arange(len(indices))]
```

It's ugly, but it should get the job done.

lwgray: 👍

yellowbrick/classifier/class_balance_heat_map.py
lwgray: @bbengfort What is MantleValueError?

yellowbrick/classifier/class_balance_heat_map.py
lwgray: @ndanielsen should i simply create my own folder in the examples directory?

yellowbrick/classifier/class_balance_heat_map.py
lwgray: @bbengfort The default color scheme is leading to information being lost in the heatmap.  This is because the scheme consists of only 6 colors that repeat themselves until all classes have been assigned a color.   Therefore, all datasets with more than 6 classes will have multiple bars with the same primary color thus data loss.

This is why I liked your original color scheme(SPECTRAL) choice; I think it consists of 11 colors.   

How many classes will ClassPredictionError support?   And can we adopt a color scheme that is flexible to larger number of classes while still being visually decipherable ?

yellowbrick/classifier/class_balance_heat_map.py
bbengfort: Whoops, sorry - [`YellowbrickValueError`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/exceptions.py#L56) - mantle is a project at my work.

yellowbrick/classifier/class_balance_heat_map.py
bbengfort: You can, using your GitHub username preferably, but that's not required. Documentation for this would go in `docs/api/classification/class_balance.rst` or you could make an .rst file next to that if you prefered.

yellowbrick/classifier/class_balance_heat_map.py
bbengfort: Yeh, that is a problem - and that's why the user really does need to specify the color for > 6 classes; which is less common but definitely a possibility. I'm not sure you have to do anything about it, but you could pass `spectral` as the name of the colormap for your example. 

Generally, we follow [Cynthia Brewer's advice](http://colorbrewer2.org/#type=sequential&scheme=BuGn&n=3) when it comes to color, so if you'd prefer you could use `Set1` as the default; when I fix `resolve_colors` I do plan on having a color cycle that is a bit longer using this scheme.

yellowbrick/classifier/class_balance_heat_map.py
lwgray: Took care of it 👍 

#326

lwgray:
@bbengfort Thank you for all the attention you have giving this PR.  
When you review this most recent push, could you pay particular attention on the lines of code in class_balance.py where I am comparing the lengths of indices and classes then raising an Error if they don't match (lines 239-241).

yellowbrick/classifier/class_balance.py
```diff
@@ -168,3 +175,171 @@ def class_balance(model, X, y=None, ax=None, classes=None, **kwargs):
...
+        y_type, y_true, y_pred = _check_targets(y, y_pred)
+
+        if y_type not in ("binary", "multiclass"):
+            raise YellowbrickValueError("%s is not supported" % y_type)
+
+        indices = unique_labels(y_true, y_pred)
+
+        if len(self.classes_) != len(indices):
+            raise ModelError(
+                'Please correct label mismatch.')
```
bbengfort: So I'm trying to reason about this outloud: first, we know that `self.classes_` is either specified by the user as a list of labels or it is extracted from [`self.estimator.classes_`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/base.py#L108). We should note that we do no further checks on that, so if `self.classes_` is None, then we're in trouble, but I don't think that's your problem, I think that's a problem for `ClassificationScoreVisualizer`. 

If `self.classes_` is supplied by the user:

- if `len(self.classes_) < len(indices)` it could be that the user is trying to filter labels, but this is currently not implemented. 
- if `len(self.classes_) > len(indices)` then potentially the user-supplied bad class information or potentially the reason below for the same condition.

If `self.classes_` is derived from the underlying estimator:

- if `len(self.classes_) < len(indices)` it could be that the estimator does not know about classes in y_true, but presumably this would have errored during the computation of y_pred and will get caught in `_check_targets` 

- if `len(self.classes_) > len(indices)` it could be that `y` and `y_pred` are missing a class that was contained in `y_train`. In that case, we don't know which class is missing without knowing if we have label encoded data or not. The bad thing that could happen is that we mislabel the classes, which would be ultra bad.  We have two choices here, raise an exception (current choice) or make a choice and warn the user. Making a choice would likely involve comparing the contents of `self.classes_` with `indices`. 

Does that reasoning make sense? 

Currently I propose the following:

```python
if len(self.classes_) > len(indices):
    raise ModelError('y and y_pred contain zero values of one of the specified classes') 
elif len(self.classes_) < len(indices):
    raise NotImplementededError('filtering classes is currently not supported') 
```

Thoughts?

lwgray: Your logic seems fine...  While fooling around today, I accidentally raised this error.  

I had built a classification dataset (sklearn - make_classification) with 5 classes ['a','b','c','d','e'] however when I ran ClassPredictionError I only specified 4 classes

```python 
viz = ClassPredictionError(RandomForestClassifier(), classes=['a','b','c','d']
viz.fit(X_train, y_train)
viz.score(X_test, y_test)
viz.poof()

---------------------------------------------------------------------------
ModelError                                Traceback (most recent call last)
<ipython-input-8-dc1390c3a6b8> in <module>()
      4                                   classes=['a', 'b', 'c', 'd'])
      5 visualizer.fit(X_train, y_train)
----> 6 visualizer.score(X_test, y_test)
      7 visualizer.poof(outpath='heatmap.png')
      8 plt.show()

~/portfolio/yellowbrick/yellowbrick/classifier/class_balance.py in score(self, X, y, **kwargs)
    239         if len(self.classes_) != len(indices):
    240             raise ModelError(
--> 241                 'Please correct label mismatch.')
    242 
    243         # Create a table of scores whose rows are the true classes

ModelError: Please correct label mismatch.


```

bbengfort:
@lwgray tests have been fixed - if you hit the "update branch" button and then do a pull your local branch should have the tests fixed as well.

lwgray:
@bbengfort 
1. I created the test stubs that you requested and placed them in test/test_classifiers/test_class_prediction_error.py
2.  I Fixed the broken code, caused by your work on resolve_colors function
3.  Created an example notebook under /examples/lwgray 
4. Travis is still not working... I updated branch them pulled to my local branch

tests/test_classifier/test_class_prediction_error.py
```diff
@@ -0,0 +1,98 @@
...
+import pytest
+import numpy as np
+import numpy.testing as npt
+
+from tests.base import VisualTestCase
+from tests.dataset import DatasetMixin
+from yellowbrick.classifier.rocauc import *
+from yellowbrick.exceptions import ModelError
+
+from sklearn.svm import LinearSVC
```
bbengfort: For the pyflake tests to pass, we'll have to get rid of all unused imports; that may have been one issue with the tests not passing.

tests/test_classifier/test_class_prediction_error.py
```diff
@@ -0,0 +1,98 @@
...
+y = np.array([1, 1, 0, 1, 0, 0])
+
+##########################################################################
+##  Tests
+##########################################################################
+
+class ClassPredictionErrorTests(VisualTestCase):
+
+    @pytest.mark.skip(reason="not implemented yet")
+    def test_class_report(self):
```
bbengfort: `test_class_prediction_error`?

tests/test_classifier/test_class_prediction_error.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        Assert no errors occur during classification report integration
```
bbengfort: Assert no errors occur during class prediction error integration?

tests/test_classifier/test_class_prediction_error.py
lwgray: not quite sure what you are asking here... 😕

tests/test_classifier/test_class_prediction_error.py
lwgray: I cut and paste... And didn't change the text

tests/test_classifier/test_class_prediction_error.py
lwgray: done 😄

lwgray:
I must have deleted the rocauc baseline images by accident.  In all honest, I am not quite sure what you are referring to

lwgray:
@bbengfort All the test have passed now

lwgray:
@bbengfort Thanks for letting me help out.  
Also, I don't want you guys to forget about this issue you wanted created.  I would but I don't know enough to post this issue.
#### ISSUE: 
"Then let's open up an issue to create the target module as per @rebeccabilbro "

bbengfort:
@lwgray thanks! see #334

PR #327, opened 2018-03-08
Fixes to the resolve colors function

bbengfort:
This PR provides fixes to the `resolve_colors` function and adds a test suite to guard against regression. I've also reimplemented `resolve_colors` on the feature visualizers that had it commented out to use `get_color_cycle` instead. 

This PR unblocks #321 -- however, a more extensive overhaul of color and color handling in YB is required. We have 4 different methodologies that are mixed and matched in kind of a strange fashion. More than that, Yellowbrick-specific requirements have emerged. Briefly, they are:

- We need discrete colors that are visually separate for classes 
- We need continuous colors for regression values and heatmaps 

Both of these requirements are specified by the target variable. Finally, we need helpers for the other utilities we use; but those can be largely handled with `resolve_colors` and `get_color_cycle`.

### DIFF
--- a/yellowbrick/style/colors.py
+++ b/yellowbrick/style/colors.py
@@ -54,58 +54,71 @@ def get_color_cycle():
     return mpl.rcParams['axes.color_cycle']
 
 
-def resolve_colors(num_colors=None, colormap=None, color=None):
+def resolve_colors(n_colors=None, colormap=None, colors=None):
     """
-    Resolves the colormap or the color list with the number of colors.
-    See: https://github.com/pydata/pandas/blob/master/pandas/tools/plotting.py#L163
+    Generates a list of colors based on common color arguments, for example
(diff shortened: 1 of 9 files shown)


yellowbrick/style/colors.py
```diff
@@ -54,58 +54,71 @@ def get_color_cycle():
...
+    plotting library that has been removed in the new version of the library.
     """
 
-    # Work with the colormap
-    if color is None and colormap is None:
-        if isinstance(colormap, str):
-            cmap = colormap
-            colormap = cm.get_cmap(colormap)
+    # Work with the colormap if specified and colors is not
+    if colormap is not None and colors is None:
```
lwgray: Great, you caught the logic here that was preventing usage of default color scheme.

bbengfort:
Going to merge this PR even though tests are failing; it looks like something's up with Travis; will open a new branch to try to sort that out.

Issue #322, opened 2018-03-05
Improvements for FeatureImportances visualizer 


bbengfort:
There are a couple of enhancements for the `yellowbrick.features.FeatureImportances` visualizer that should be made to really make it stick out. They are as follows:

*Note to contributors: items in the below checklist don't need to be completed in a single PR; if you see one that catches your eye, feel to pick it off the list!*

- [ ] color negative coefs 
- [ ] top n features to filter number displayed (both pos and neg)
- [ ] implement standard deviation for ensemble models 

### Color negative coefs 

The first item is relatively straightforward, currently, the bar chart is a single color, but it might be nice to show negative `coef_` values as a different color, e.g. blue for positive green for negative as below:

![figure_1](https://user-images.githubusercontent.com/745966/36986141-1b9d5a58-2067-11e8-983e-eadb2acac686.png)

To do this, you'll have to create a color array to pass as the color argument, e.g. 

```python
colors = np.array(['b' if v > 0 else 'g' for v in self.feature_importances_]) 
self.ax.barh(pos, self.feature_importances_, color=colors, align='center')
```
We should also create arguments to provide a way to specify the colors. 

### Top N Features 

For the second item, I'm picturing something similar to [most informative features with scikit-learn](https://stackoverflow.com/questions/11116697/how-to-get-most-informative-features-for-scikit-learn-classifiers) (though not exactly this code). Here, an argument `topn` which defaults to None specifies a filter to only plot the `N` best features. 

This should also be relatively straightforward, but gets complicated in the case of negative values. We have two options, we can rank all values including negative values and plot the N best values either positive or negative, or we can do the N best positive and N best negative coefficients 

### Standard Deviation for Ensembles 

Ensemble models like Random Forest and Gradient Boosting have an underlying `estimators_` attribute that describes each feature's importance in a different way. The global feature importances are the mean, but it would be nice to add an `xerr` bar with the standard deviation as in [plot forest importances](http://scikit-learn.org/stable/auto_examples/ensemble/plot_forest_importances.html). 

This could also be useful for `CV` models that also have an underlying `estimators_` attribute. 

The idea with this one is to compute the standard deviations for each feature using `estimators_` and np.std, and storing the value in a `confidences_` attribute during fit. Note that it will also have to be sorted using the `sort_index` -- the confidences are drawn during `ax.barh` with `xerr=self.confidences_`. 

Right now it looks like the example above is no longer working exactly as expected, so some deeper review is necessary. 

[...]

PR #335, opened 2018-03-14
Feature importance Visualizer Improvement - Colors of Coef

lwgray:
This PR aims to resolve a singular problem of issue #322 
The problem I have chosen to resolve deals with the coloring of  the bars of the graph.
This problem consists of 2 aims:
#### Aim 1:
Problem: Apply different colors to the positive and negative coefficients.
Solution:   A slight alteration of the code provided by @bbengfort 

#### Aim 2:
Problem:  Allow user to specify which colors to use for positive and negative coefficients
Solution:  Used Yellowbrick.style function "resolve color"

### Example
importance


bbengfort:
Also, it does look like this change broke the tests for feature importance; which may not be a bad thing - we just have to see what changed in the image and create a new baseline image if the change is reasonable.

lwgray:
@bbengfort It's funny that you asked these questions b/c I have thought about most of these items.  In fact I had originally code exceptions but decided to leave them out b/c I thought the resolve_colors function could handle most of theses.  However, I will address each of these in a subsequent commit.

bbengfort:
Hehe, you don't necessarily have to handle it with exceptions - documentation and test cases is fine also!

ndanielsen:
just quick note that this is looking great so far and a fantastic contribution!

bbengfort:
@lwgray please do a pull before you continue to work on this; I updated your branch pending the release of 0.6

Issue #331, opened 2018-03-10
RadViz does not support Pandas DataFrames


Kautumn06:
I tried to use a Pandas DataFrame with RadViz and received the following error:

`ValueError: Unable to coerce to DataFrame, shape must be (20560, 5): given (1, 5)`

Here is the code I used:

```
# Load the classification data set
data = load_data('occupancy')

# Specify the features of interest and the classes of the target
features = ["temperature", "relative humidity", "light", "C02", "humidity"]
classes = ['unoccupied', 'occupied']

X = data[features]
y = data.occupancy

# Import the visualizer
from yellowbrick.features import RadViz

# Instantiate the visualizer
visualizer = RadViz(classes=classes, features=features)

visualizer.fit(X, y)      # Fit the data to the visualizer
visualizer.transform(X)   # Transform the data
visualizer.poof()         # Draw/show/poof the data
```

ndanielsen:
@Kautumn06 thanks for creating this issue!

ndanielsen:
@Kautumn06 I added some tests to the RadViz and it seems to support pandas dataframes in the tests that I wrote / adapted from your example. Can you take another look at this?

https://github.com/DistrictDataLabs/yellowbrick/pull/338/files

ndanielsen:
closing this issue as it now has test coverage to confirm it works

PR #338, opened 2018-03-15
Add pandas test to radviz

ndanielsen:
Related to https://github.com/DistrictDataLabs/yellowbrick/issues/331

A quick PR to confirm that Radviz supports the use of pandas dataframes, and fix some of the visual unit tests that didn't seem to be properly implemented.


tests/test_features/test_radviz.py
```diff
@@ -16,13 +16,17 @@
...
-
+import unittest
 import numpy.testing as npt
 
 from tests.base import VisualTestCase
 from tests.dataset import DatasetMixin
 from yellowbrick.features.radviz import *
 
+try:
+    import pandas
```
Kautumn06: I noticed that in some of the tests, such as [test_importances.py](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_features/test_importances.py), that we use ``import pandas as pd`` instead of ``import pandas``. Of course, they both work; however, I thought this might be another issue we could add to the discussion about establishing a uniform style.

ndanielsen: @Kautumn06 thanks for pointing this out.

I've created an issue around this style and convention issue : https://github.com/DistrictDataLabs/yellowbrick/issues/340

In this particular case, I'm not as concerned with uniform conventions in the test suite as it is a little deeper in the library, but the point still is well taken.

PR #342, opened 2018-03-16
ClassPredictionError Tests

lwgray:
This is a PR to adds potential tests for ClassPredictionError #321

Below you will find a detailed outline of the test that I have already created.  Writing test is not my strong suit.

For the following group of tests, I mainly wrote test for image similarity and assert raised errors.

#### Test Name:  test_class_type
    Background/tested: 
    Not all data types are appropriate for this classifier.
    Our classifier handles both multiclass and binary classed data
    Handle: 
    We test this fact by supplying our Classifier with Multilabel data
    which should raise a YellowbrickValueError

#### Test Names: test_classes_less_than_indices & test_classes_greater_than_indices

    Background/tested:
    There can be a mismatch between the number of classes and number of indices in the data.
    There are two sides to the mismatch.  The first is when there are greater number of
    classes than indices and secondly the opposite, when there are less classes than indices.

        * Let's look at when there are more classes than indices
            * This can occur 2 different ways
                * the first is when a user specifies less classes than indices.

                * The second is when y and y_pred contain zero values for one of the 
                   specified classes
        * Now let's look at when there less classes than indices
            * This occurs when the user is trying to filter classes

#### Test Names: test_integration_class_prediction_error & test_class_prediction_error_quickmethod
    Simply the Visualizer must complete without error
        * Assert produced images are similar to baseline images
    Again, simply the visualizers quick method must work without error
        * Assert produced images are similar to baseline images


lwgray:
Don't review yet, I need to push another commit - I found several errors

lwgray:
@bbengfort 
In latest commit, the following files are important
tests/test_classifier/test_class_prediction_error.py
yellowbrick/classifier/class_balance.py

tests/test_classifier/test_class_prediction_error.py
```diff
@@ -20,68 +21,97 @@
...
-        model.fit(X,y)
+        model.fit(X, y)
         visualizer = ClassPredictionError(model, classes=["A", "B"])
-        visualizer.score(X,y)
+        visualizer.score(X, y)
         self.assert_images_similar(visualizer)
 
-    @pytest.mark.skip(reason="not implemented yet")
-    def test_no_classes_provided(self):
+    def test_class_prediction_error_quickmethod(self):
```
bbengfort: Just a quick glance at the quick method baseline image only shows one bar, was this the desired output?

tests/test_classifier/test_class_prediction_error.py
```diff
@@ -20,68 +21,97 @@
...
         """
-        Assert no errors when no classes are provided
+        Test the ClassPreditionError quickmethod
         """
-        pass
+        fig = plt.figure()
+        ax = fig.add_subplot()
```
bbengfort: I think it's very good practice to ensure that the axes for the test being drawn on was created by the user. I used to use this methodology but recently I've noticed that `ax == None` -- I'm not sure what's going on there so I've switched over to `fig, ax = plt.subplots()`.

tests/test_classifier/test_class_prediction_error.py
```diff
@@ -20,68 +21,97 @@
...
 
-    @pytest.mark.skip(reason="not implemented yet")
-    def test_class_mismatch(self):
+        clf = LinearSVC(random_state=42)
+        g = class_prediction_error(clf, X, y, ax)
+
+        self.assert_images_similar(ax=g)
```
bbengfort: Nice! Glad you could make use of this new feature of `assert_images_similar`

tests/test_classifier/test_class_prediction_error.py
lwgray: no it wasn't the desired outcome... I will take a look at it

tests/test_classifier/test_class_prediction_error.py
lwgray: Just realized that the Dummy dataset that we used to test ClassPredictionError(CPE) isn't sufficient... 
1.  Because you can perfectly predict the classes thus CPE is pointless.  I think maybe we should do sklearn make_classification with a random state

bbengfort: I absolutely agree, I've also started to move away from using these hardcoded arrays to using `make_classifier`. You can see a recent example of mine here:

https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/test_classification_report.py#L51

This is using a pytest fixture based approach since we're also moving away from unittest.

Issue #341, opened 2018-03-16
Create yellowbrick.contrib and deprecate non-core visualizers


ndanielsen:
I that we should create a `yellowbrick.contrib` module to store visualizers, utilities and other functionalities  that are prototypes, works in progress or just plain experimental.

For the core of Yellowbrick, only highly stable, fully tested, high value added and well documented visualizers and etc should be allowed to pass in.

I think that the first step should be to identify visualizers that are be suited for `contrib` and add a `depreciated` warning to them with a note that they are moving to contrib.

I propose using the deprecation library to do this. It seems stable and well tested. 
http://deprecation.readthedocs.io/en/latest/ 

For initial candidate visualizers to move, I recommend:

- features.scatter.ScatterVisualizer
- classifier.boundaries.DecisionBoundariesVisualizer

For experimental utilities, I think that we can also immediately add a basic working implementation of StatsModelsWrapper(BaseEstimator) that was in https://github.com/DistrictDataLabs/yellowbrick/issues/306

ndanielsen:
@bbengfort @rebeccabilbro @NealHumphrey any thoughts on this proposal?

bbengfort:
Yes, I think this is the best way to move forward; we can have new contributors add to contrib, then as they become documented move them into the main library. 

I do have two questions:

- Tests? 
- Contrib documentation? 

For the first, I'm thinking we just have test files next to the actual files in contrib; but we can have pytest ignore them so they're not part of CI (pytest will run them on demand). 

For the second, I suggest we put a contrib section in the main docs that just does an API read of the docstrings rather than having full blown documentation. 

Thoughts?

bbengfort:
Also for us, we can add visualizers directly to the main code base, but it is expected we will test and document them; and if we want to do something prototype-y we can put it in contrib; only brand new contributors are expected to add to contrib until after their first PR?

ndanielsen:
I agree with the contrib documentation suggestion. For testing of contrib, I think that it can take one or two options - they can live in test/test_contrib or be right next to the actual files. I'd be fine with either and potentially both as long as we can pytest ignore what we want.

For what is in main vs contrib, I think that we can easily agree on the quality of tests, value and docs around each of them. For the substance of included in the main package, I think that it should be on a case by case basis. I agree that all first time contributors should plan to add to `contrib` and we can elevate it to the main package if it is makes sense. 

[...]

PR #346, opened 2018-03-17
Add deprecation warnings on scatterviz and boundariesvisualizer

ndanielsen:
related to https://github.com/DistrictDataLabs/yellowbrick/issues/341

Depreciating ScatterViz and DecisionBoundaries as they will be moving to `contrib` until they are more polished.

Here's a sample of the deprecation warning... open to suggestions on the wording.

screen shot 2018-03-17 at 11 57 53 am


ndanielsen:
@bbengfort thanks for taking a look at this. I like the sckit-learn implementation and it is likely better that we go with that approach. I'll take a look at it later tonight and push something up

ndanielsen:
@bbengfort this is ready for you to take a look and merge in if you are okay with it

tests/sklearn_test_utils.py
```diff
@@ -0,0 +1,88 @@
+# tests.sklearn_test_utils
```
ndanielsen: Had to copy these functions over because sklearn uses nose as part of their testing framework. As we no longer have that as a dependency, it was breaking on import

bbengfort: Nice - I think this was definitely the right decision in the face of nose dependencies. I looked through the code in `assert_warns_message` and I don't necessarily see anything that is different from [`pytest.warns`](https://docs.pytest.org/en/latest/warnings.html#warns). 

Would it just be easier to use [`pytest.deprecated_call`](https://docs.pytest.org/en/latest/warnings.html#ensuring-function-triggers)?

tests/test_features/test_scatter.py
```diff
@@ -65,6 +66,10 @@ def test_init_alias(self):
         visualizer = ScatterVisualizer(features=features, markers=['*'])
         self.assertIsNotNone(visualizer.markers)
 
+    def test_depreciated(self):
+        features = ["temperature", "relative_humidity"]
+        assert_warns_message(DeprecationWarning, "Will be moved to yellowbrick.contrib in v0.7", ScatterVisualizer, features=features)
```
bbengfort: Awesome! I think deprecation warning tests are vital for future us to remove them!

yellowbrick/classifier/boundaries.py
```diff
@@ -121,7 +123,7 @@ def decisionviz(model,
 ##########################################################################
 # Static ScatterVisualizer Visualizer
 ##########################################################################
-
+@deprecated("Will be moved to yellowbrick.contrib in v0.7")
```
bbengfort: Perfect!

PR #347, opened 2018-03-17
If applied this commit: Creates docs for ClassPredictError Visualizer

lwgray:
This commit is related to #321 and #342

ClassPredictionError is a new visualizer and lacks documentation

This commits creates both the python file and rst file demonstrating
how to create a ClassPredictError visualizer.  A image was also
produced from this demonstration, for its inclusion in the rst file.


bbengfort:
@lwgray very timely! I'll get it into v0.6

PR #348, opened 2018-03-18
Updates ClassificationReport to use pcolormesh

### LINKED ISSUE #267: Lines in Classification Report
First, this is an awesome module!!
I created a classification report (see attached) and the white lines make reading the numbers impossible.
unknown
I wanted to ask if there is a way to either lighten or remove these lines so that you can read the various scores?
Thanks!

bbengfort:
Fixes #267

Updates the classification report to use the `pcolormesh` instead of
`imshow` so that the grid lines are on the outside rather than through
the figure numbers. This also makes it more robust to non-square CR
shapes and odd grids.


bbengfort:
From:

classification_report


To:

classification_report

PR #355, opened 2018-03-20
Make sklearn classification dataset for Class Prediction Error Tests #342

lwgray:
In several places throughout the code, you will find hardcoded test data.
This is the X/y data that is used by our models.  

``` python
X = np.array(
    [[2.318, 2.727, 4.260, 7.212, 4.792],
     [2.315, 2.726, 4.295, 7.140, 4.783],
     [2.315, 2.724, 4.260, 7.135, 4.779],
     [2.110, 3.609, 4.330, 7.985, 5.595],
     [2.110, 3.626, 4.330, 8.203, 5.621],
     [2.110, 3.620, 4.470, 8.210, 5.612]]
)

y = np.array([1, 1, 0, 1, 0, 0])
```

I originally used this data to test my visualizer.  
However, I soon found out that the data isn't suitable for all
applications.  In my case, using this data produced a bar graph with only one bar.

onebar

This says that the model only predicts a single class and the visualize accurately shows this.  
However, We built this visualizer to examine different degrees of prediction error thus we need more complex data to showcase this functionally.  Additionally, we would like to move away from hard-coded X/y values.

My solution is simple. I use sci-kit learn 'make_classification' function.
This allows for more complex data sets for which I am able to specify the number of classes
and select how many are informative(plus many more parameters).  Most importantly 'make_classification'  function allows the random state to be set, thus allowing for completely reproducible test images.

complex


bbengfort:
Very nice PR description!

PR #358, opened 2018-03-22
Modifies Scoring Mechanism for Regressor Visualizers

### LINKED ISSUE #213: Add R2 score to Residuals and PE Plots
Add the score as a label to the regressor plots.

bbengfort:
This PR is meant to start changing the score visualizer methodology. Besides simply being the entry point for drawing, the score methods should also _return the output of the scoring function of the underlying model_. This commit repairs the regressor models and also updates them to include the R2 value in the plot itself with the legend.

- [x] Still to do are the classifier visualizers (create an issue for this).

This commit also updates the PredictionError and ResidualPlot integration tests to use better random data and the pytest assertions as well as includes support for pandas integration testing. I've also updated the documentation with the new visualizers.

Fixes #213


bbengfort:
Examples of annotations in the legend:

prediction_error

residuals

bbengfort:
@lwgray yes, that has to do with test discovery, `nose` wanted `BlahBlahTests` and pytest wants `TestBlahBlah` -- which is horribly annoying; but see [pytest naming convention](https://pytest.readthedocs.io/en/reorganize-docs/new-docs/user/naming_conventions.html) for more on that. So long as the tests use the old style `unittest.TestCase` they'll still be discovered, but as we're moving toward pytest, I have to update these things as we go. 

Note that we could change the discovery config to look for `BlahBlahTests` - but doing it this way helps me track what is changed to pytest and what isn't. 

There is a big danger though - tests that are not renamed will simply not be run, and we want to make sure we don't accidentally skip tests!

Issue #363, opened 2018-03-23
Visualizer for Missing Values Patterns


ndanielsen:
It would be great to visualize missing data in a meaningful format that is easy to understand the patterns around it.

### Proposal/Issue

A basic implementation of the visualizer would include on the x axis, the labels / features and the y access would include a bar chart of the number of instances of a missing values.

A more advanced implementation of this would still have the x axis be the labels / features, but the range would be some type of order - either the index location or perhaps a datetime. Each missing value would be shown in a meaningful way potentially similar to this [visualization in this blog post](https://blogs.sas.com/content/iml/2017/11/29/visualize-patterns-missing-values.html):

![missinggraphics3-1](https://user-images.githubusercontent.com/2944777/37850815-fafb1812-2eb2-11e8-8069-ab86ed66490b.png)

bbengfort:
I really like this idea and I think it's fairly requested feature.

ndanielsen:
Others idea for related missing values visualizers:

Missing Values by Target (dispersion and bar), might be able to extend the currently in progress visualizers to do this.

RaulPL:
This library has some nice visualizations of missing data:
https://github.com/ResidentMario/missingno

ndanielsen:
closing PR until I can finish it up https://github.com/DistrictDataLabs/yellowbrick/pull/366

ndanielsen:
Follow up work to elevate this `contrib` module into the main package, located in:
https://github.com/DistrictDataLabs/yellowbrick/pull/519

PR #366, opened 2018-03-24
Missing Values BarChart and Dispersion Visualizers (wip)

ndanielsen:
Related issue : https://github.com/DistrictDataLabs/yellowbrick/issues/363

A work in progress and start of a series of visualizers for missing values. 

This PR has work in progress for missing values bar and dispersion visualizer. 

[Description]
Here's an initial plot done in an example notebook. 
missing_values_bar

One thought is that not sure if this fits into the feature visualizer category or if it should be another class of visualizer for that lives in yellowbrick.missingvalues. 

To Items Before Merging In

- [x] Put code in proper visualizer format
- [x] Add Basic Tests 
- [x] Handling for pandas dataframes, etc
- [ ] Add documentation


yellowbrick/features/missing.py
```diff
@@ -0,0 +1,129 @@
...
+# from yellowbrick.style.colors import resolve_colors
+
+
+
+##########################################################################
+## Feature Visualizer
+##########################################################################
+
+
+class MissingValuesBarVisualizer(DataVisualizer):
```
bbengfort: I've been thinking that naming things with a `Visualizer` or `Viz` suffix stutters a bit, I think that `MissingValuesBar` is just as informative and is a bit easier to handle for users.

yellowbrick/features/missing.py
```diff
@@ -0,0 +1,129 @@
...
+# For license information, see LICENSE.txt
+#
+# ID: missing.py [] nathan.danielsen@gmail.com.com $
+
+"""
+Implementation of missing values visualizers
+
+To Include:
+- Bar
+- Density Matrix (by time, specifiable index)
```
bbengfort: I'm super excited about this! 

The density matrix =seems similar to [Lexial Dispersion Plots](http://www.nltk.org/book/ch01.html#fig-inaugural) -- and honestly, I'm most excited about this idea. I could see a workflow that uses these visualizers as follows:

- Bar chart - determine if a feature should be eliminated because so many outliers 
- Dispersion plot - determine if instances should be removed or if there is an ordering effect in missing values 

Not sure what the heatmap vision is - but very cool!

yellowbrick/features/missing.py
```diff
@@ -0,0 +1,129 @@
...
+
+        if self.features_ is None:
+            self.features_ = range(nrows)
+
+        if self.classes_ is None:
+            # TODO: Is this the most efficient method?
+            self.classes_ = [str(label) for label in np.unique(y)]
+
+        nan_matrix = X.astype(float)
+        self.nan_col_counts = [np.count_nonzero(np.isnan(col)) for col in nan_matrix.T]
```
bbengfort: For my proposed usage above, probably % of instances would be a better metric than raw count; perhaps this could be an option supplied to the user?

yellowbrick/features/missing.py
```diff
@@ -0,0 +1,129 @@
...
+            X = X_selected.copy().view((np.float64, len(X_selected.dtype.names)))
+
+        else:
+            pass
+
+        if self.features_ is None:
+            self.features_ = range(nrows)
+
+        if self.classes_ is None:
+            # TODO: Is this the most efficient method?
```
bbengfort: Yes- I did some benchmarking and np.unique is the best method -- particularly because it returns sorted classes. I'd further recommend turning this into a np.array instead of a list.

yellowbrick/features/missing.py
```diff
@@ -0,0 +1,129 @@
...
+        # handle numpy named/ structured array
+        elif self.features_ is not None and is_structured_array(X):
+            X_selected = X[self.features_]
+            X = X_selected.copy().view((np.float64, len(X_selected.dtype.names)))
+
+        else:
+            pass
+
+        if self.features_ is None:
+            self.features_ = range(nrows)
```
bbengfort: Would prefer np.arange

yellowbrick/features/missing.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def fit(self, X, y=None, **kwargs):
+        """
+        TODO if y, then color code the missing values in the chart?
+
+
+        """
+        nrows, ncols = df.shape
+
+        # Handle the feature names if they're None.
```
bbengfort: I've actually investigated deeper how sklearn handles dataframes/structured arrays etc. It might not be relevant here, but I'm happy to go over it with you.

yellowbrick/features/missing.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        # Set the title
+        self.set_title(
+            'Missing Values by Column'
+        )
+        ind = np.arange(len(self.features_))  # the x locations for the groups
+        # Remove the ticks from the graph
+        self.ax.set_ylabel('Count')
+        self.ax.set_xticks(ind)
+        self.ax.set_xticklabels(self.features_, rotation='vertical')
```
bbengfort: Might this be better as a vertical bar chart instead of a column chart? Particularly for datasets with lots of features, it might give the visualizer a bit more breathing room.

yellowbrick/features/missing.py
bbengfort: https://bbengfort.github.io/observations/2017/05/02/python-unique-benchmark.html

bbengfort:
Let me know when you'd like me to review.

ndanielsen:
closing until I can finish it up

Issue #368, opened 2018-03-24
CompareClassifiers visualizer


NealHumphrey:
When doing classification, I often want to compare the results of different scenarios to each other. Here a scenario is the full process of the model selection triple (features, model, hyperparameters) - if I make a change in any one of them, is it an improvement over previous scenarios? 

In the [housing risk project](https://github.com/georgetown-analytics/housing-risk) I approached this by making a bar chart of the `classification_report` parameters, shown below, for various models. I propose replicating this in yellowbrick as a starting point for scenario/model comparison visualizers. 
![for_presentation_under_sampling_confusion_comparisons](https://user-images.githubusercontent.com/5642566/37867405-3f81ce70-2f6e-11e8-84d6-badbb03276a4.png)

# Implementation
As (I believe?) the first multi-model visualizer, this will require some new additions to the yellowbrick API. As such it should definitely go into the [proposed contrib section](https://github.com/DistrictDataLabs/yellowbrick/issues/341), and I expect the first version won't necessarily get it quite right. 

The implementation approach is based on some discussions with @bbengfort about how multi-model visualizers might be implemented, and building on the new [VisualizerGrid](https://github.com/DistrictDataLabs/yellowbrick/pull/176#issuecomment-372522465) to implement the subplots. 

Also, the visualizer needs a better name. 

# Enhancements
A future version could implement the 12-fold `cross_val_predict` method as was sketched out in the MultiModelMixin.

bbengfort:
@NealHumphrey - great idea and certainly a good start to the multi-model visualizer API! As per my last note on `contrib`, it doesn't necessarily have to go in `contrib` but we could prototype it in an edge branch. 

In your demo, are in/out the names of the classes?

ndanielsen:
Just wanted to jump in and say this is awesome

NealHumphrey:
@bbengfort yep in/out are class names. You can see my work in progress in the PR I just opened, which has class names that are more clearly classes.

bbengfort:
@NealHumphrey checkout: 

https://github.com/HDI-Project/ATM

We've been talking to one of the folks on that project and we're thinking about adding some classification comparison visualizers that use ATM to the contrib package.

rebeccabilbro:
Going to archive this one for now since the conversation has gone a bit stale; happy to reopen if someone has bandwidth and interest!

PR #372, opened 2018-03-27
WIP: 368 compare classifiers

NealHumphrey:
Adds feature in #368

This is very much a work in progress still, but opening a pull request to demo functionality so far and get some layout advice on matplotlib which can be super annoying. 

You can see a demo in /examples/nealhumphrey/CompareClassifiers.ipynb. Here's what the visualizer looks like right now: 
image

Things TODO:
- I want to reduce `plt.subplots_adjust(hspace=0.25)` in line 94 to more like 0.1 so the subplots are closer together, but I also want to put a title on the first subplot of each category. So, I want to adjust only some of the subplots to have more space above them to accomodate the title. Any ideas on how to do this? I could hack it by using a dummy visualizer that only displays a title, but seems like there should be a better way. 
- Properly re-draw the chart when appending new charts after running poof() - need to update and re-output the chart (right now it doesn't return anything if you run poof again)
- Properly handle colors. Maybe apply a heatmap scale to the bars color that mimics the bar values to better emphasize the best performing of each subchart?
- Smarter dimension defaults and methods
- Unit tests


bbengfort:
@NealHumphrey I think the answer to this might be to use a [GridSpec](https://matplotlib.org/users/gridspec.html) -- this allows for complex nested axes that each have their own padding and titles. In this case, that might forestall the use of the `VisualizerGrid` but it would give you more fine-tuned control of the axes objects.

bbengfort:
Let me know when you want me to do a review.

rebeccabilbro:
@NealHumphrey - we're going to close this one out since YB has through 2 new releases in the interim. @bbengfort and I still love this idea, so let us know when you're ready to start working on this one again!

PR #376, opened 2018-03-29
Adding yellowbrick.contrib to main package, tests and docs

### LINKED ISSUE #341: Create yellowbrick.contrib and deprecate non-core visualizers

ndanielsen:
Fixes #341 

[Description]
Adding in the contrib module to the main package and updated other associated codebase and documentation. 

I moved the decision boundaries visualizer there as an initial starter. For the docs, to keep it simple - I kept the `contrib/index.rst` in a flat structure that will be easier to maintain than in a nested structure as we do in the main docs.

If this looks good, I can migrate the ScatterViz other in another PR. 


ndanielsen:
@bbengfort good for you to take a look at

PR #377, opened 2018-03-30
Model Selection Package

bbengfort:
An initial proposal for a model selection package that mirrors `sklearn.model_selection`. Since most of what YB does is about model selection I was unsure if this is the correct name. If it is, then grid search should be placed under this library.

Alternatively, we could call this package `cross_validation` and place CV visualizers such as the one @pdamodaran is working on and the learning curve visualizer in this package and keep this separate from the grid search module. However, this might be confusing because of scikit learn's deprecation of the `cross_validation` package.

@DistrictDataLabs/team-oz-maintainers -- would appreciate feedback! 

This PR will go on to implement the learning curve visualizer more generally for both classification and regression and stick it into the model selection or cross validation package (whichever we choose).


bbengfort:
@ndanielsen I know this is a lot, but mind taking a quick glance through? I tried to break it up into meaningful commits so it would be easier to take it piece by piece. 

@jkeung I moved learning curve from `yellowbrick.classifier` to `yellowbrick.model_selection` and added documentation and rewrote the image comparison tests. I also added the validation curve, which is very similar to learning curve and added docs for that as well.

docs/api/index.rst
```diff
@@ -14,6 +14,7 @@ Welcome to the API documentation for Yellowbrick! This section contains a comple
    regressor/index
    classifier/index
    cluster/index
+   model_selection/index
```
ndanielsen: I agree and think that this is the appropriate name for the package

yellowbrick/model_selection/validation_curve.py
```diff
@@ -0,0 +1,364 @@
...
+        figure to.
+
+    Returns
+    -------
+    ax : matplotlib.Axes
+        The axes object that the validation curves were drawn on.
+    """
+
+    # Initialize the visualizer
+    oz = ValidationCurve(
```
ndanielsen: is `oz` the new convention for `viz`? I do like this

yellowbrick/classifier/learning_curve.py
```diff
@@ -1,233 +0,0 @@
-# yellowbrick.classifier.learning_curve
```
ndanielsen: agree with the move to `model_selection` for this

tests/test_classifier/test_learning_curve.py
```diff
@@ -1,133 +0,0 @@
-# tests.test_classifier.test_learning_curve
```
ndanielsen: I was going to say that we need to put a deprecated warning on the learning curve visualizer, but since it wasn't in the documentation. We are probably okay with moving it and then documenting it.

yellowbrick/model_selection/validation_curve.py
bbengfort: I'm not sure for documentation purposes, but I'm certainly going to start using it in my code -- I hope you do too!

tests/test_classifier/test_learning_curve.py
bbengfort: Yeh, pretty much all my previous PRs and upcoming PRs are going to be about documenting the undocumented classifiers and ensuring they have image comparison tests. I thought this would be simple, but as you can see it does require a fair amount of refactoring.

yellowbrick/model_selection/validation_curve.py
ndanielsen: agreed! unfortunately, there is no github emoji for:


8-1-9-5cm-flying-monkey-vinyl-car-stickers-motorcycle-decals-window-car-accessories-c2-0607 jpg_640x640

tests/test_classifier/test_learning_curve.py
ndanielsen: flag anything that needs to go into `contrib` and I can migrate it over

yellowbrick/model_selection/validation_curve.py
rebeccabilbro: Hey there @Kautumn06 -- any updates on your progress with establishing a naming convention for instantiated visualizers in our code base and in the yb docs? Maybe you could open a new pull request so we can keep track of the discussion?

tests/test_classifier/test_learning_curve.py
bbengfort: Will do

PR #382, opened 2018-04-04
Adds RFECV visualizer for feature selection

### LINKED ISSUE #268: Recursive Feature Elimination
This [recursive feature elimination heatmap](http://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.RFE.html#sklearn.feature_selection.RFE) would be a cool feature selection visualization to implement in yb - possibly we can hoist some of #164 into a base class (see #58) for feature importance visualizers?

bbengfort:
Adds a custom recursive feature elimination with cross-validation based visualizer that doesn't directly use `sklearn.feature_selection.RFECV` but instead wraps `sklearn.feature_selection.RFE` more directly. It's a bit slower but gives a better visualization overall as well as direct access to the underlying RFE model.

Special thanks to Cortex - part of this visualizer was developed to assist with model selection at my work!

Fixes #268

rfecv_credit


bbengfort:
Oh, also I updated the Wrapper tests to test for "rewrapping" -- this visualizer generates a new RFE model under the hood, and the `ModelVisualizer` wraps that instead of the original estimator. During this update I also updated the tests to use pytest instead of unittest

ndanielsen:
@bbengfort I'll take a look in the next 24 hours or so

bbengfort:
:-/ Travis CI fails on a visualizer that's been working for a long time, I simply restart the job and pass ...

yellowbrick/features/rfecv.py
```diff
@@ -0,0 +1,331 @@
...
+
+        if step < 0:
+            raise YellowbrickValueError("step must be >0")
+
+        # Create the RFE model
+        rfe = RFE(self.estimator, step=step)
+        n_feature_subsets = np.arange(1, n_features+1)
+
+        # Create the cross validation params
+        # TODO: handle random state
```
ndanielsen: this seems like a TODO that probably be in here, could you add this into this PR?

tests/test_utils/test_wrapper.py
```diff
@@ -116,3 +118,21 @@ def test_wrapped_methods(self):
...
+    def test_rewrap_object(self):
+        """
+        Test the ability to "rewrap" an object on demand
+        """
+        obj = WrappedEstimator()
+        old = obj._wrapped
+        new = mock.MagicMock(spec=GaussianNB())
+
+        obj.predict()
+        old.predict.assert_called_once()
```
ndanielsen: magic mock to the rescue for hard to test things

yellowbrick/features/rfecv.py
bbengfort: I can take a look at it, the issue is that the mechanism I'm using, [`cross_val_score`](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_val_score.html), doesn't take `random_state` as a parameter. 

In the `DescriminationThreshold` visualizer, you implemented `random_state` which was passed to `train_test_split`, later I updated it to take a `cv` argument, which in order to use the random state, I had to implement a [`_check_cv`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/threshold.py#L416) method that mirrored your use of `train_test_split`. This was loosely based off of [`sklearn.model_selection.check_cv`](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/model_selection/_split.py#L1862), though the sklearn mechanism is much more robust than what I did and I'm not sure if I like how I went about things in threshold.py.

So, the reason I added it as a TODO is because I'm not sure what to do here. The only option to generically handle the `cv` argument with a random state is to create our own splitter, it seems like a shame to implement this just for that purpose since it will be rigid with respect to skearn's CV system. I assume the reason `cross_val_score` doesn't take a random_state parameter is that you can simply pass it a splitter with the random state already set (which is what I'm doing in the tests). 

Any thoughts on this?

bbengfort: So on further investigation, it actually looks like no random state is needed:

https://gist.github.com/bbengfort/ae9f9a52ebb55033845553722175c405

The cv method is deterministic unless `shuffle=True`, which is always false when you pass in `None` or an integer. Does this look right?

PR #385, opened 2018-04-06
Create visualizer to give reference for binning samples into balanced classes

### LINKED ISSUE #312: Demonstration/walkthrough proposal on how to use ClassBalance (and variations) to bin continuous values for classification
I. Title 
Better Binning for Classification Problems: 
Creating Categorical Values from Continuous Values

II. Premise
- A lot of machine learning problems in the real world suffer from the curse of dimensionality; you have fewer training instances than you’d like, and predictive signal is distributed (often unpredictably!) across many different features. 
- One example is when your target is continuously-valued, but there aren’t enough instances to predict these values to the precision of regression.
- What if we transform the regression problem into a classification problem? We can try to do this by binning the continuous values into buckets for classification. But how do we pick the bins?

III. Dataset Intro
- About the Pitchfork album reviews corpus - funny! snarky! sentiment analysis??
- Download the data from https://www.kaggle.com/nolanbconaway/pitchfork-data/data
- Custom CorpusReader to access the text and scores
- Custom TextNormalizer to lemmatize and remove stop words 
- use Numpy digitize method to naively bin the continuous target values

IV. Preliminary Text Analytics Pipeline
- Build using Scikit-Learn Pipeline
- use ConfusionMatrix to visually evaluate
- use ClassBalance to visualize imbalance
- talk through selection bias - why initial bins didn’t work

V. Tuning Bins with ClassBalance
- redo with better distributed bins for target values
- (hopefully) show better results

VI. Conclusion/Teaser for New ClassBalanceHeatmap Visualizer
- how to combine insight from ConfusionMatrix with interpretability of ClassBalance?

Reviewers:
@marskar
@lwgray
@yzyzy

Juan0001:
The BalancedBinningReference is created to give references for binning samples into balanced classes.

Fixes #312 

[Description]
@rebeccabilbro was working on dataset from Pitchfork, a music review website and trying to predict the reviewer's score on a piece of music based on the review content. You can check this link to follow her story on music review score prediction: https://rebeccabilbro.github.io/better-binning/

The original scores of the music reviews are numerical, numbers between 0 and 10. Because the high dimention of the text data, she was not able to predict the floating score by using regression. So she decided to bin the scores into 4 categories (“terrible”, “okay”, “great”, and “amazing”). However, the result didn't come back ideal. She found that the unbalanced samples in each bin might be the reason the performance of the model was not as good.

To solve this problem, I decided to create a visualizer to give a reference for binning balanced samples into each bin.

### DIFF
--- a/yellowbrick/features/histogram.py
+++ b/yellowbrick/features/histogram.py
@@ -1,0 +1,191 @@
+
+# yellowbrick.features.histogram
+# Implementations of histogram with vertical lines to help with balanced binning.
+#
+# Author:   Juan L. Kehoe (juanluo2008@gmail.com)
+# Created:  Tue Mar 13 19:50:54 2018 -0400
+#
+# Copyright (C) 2018 District Data Labs
+# For license information, see LICENSE.txt
+#
+# ID: histogram.py
+
+"""
+Implements histogram with vertical lines to help with balanced binning.
+"""
+
+##########################################################################
+## Imports
+##########################################################################
+import warnings
+import matplotlib.pyplot as plt
+import numpy as np
+
+from yellowbrick.features.base import FeatureVisualizer
+from yellowbrick.exceptions import YellowbrickValueError
+from yellowbrick.utils import is_dataframe
+
+##########################################################################
+## Balanced Binning Reference
+##########################################################################
+
+class BalancedBinningReference(FeatureVisualizer):
+    """
+    BalancedBinningReference allows to generate a histogram with vertical lines
+    showing the recommended value point to bin your data so they can be evenly
+    distributed in each bin.
+
+    Parameters
+    ----------
+    ax: matplotlib Axes, default: None
+        This is inherited from FeatureVisualizer and is defined within
+        BalancedBinningReference.
+    feature: string, default: None
+        The name of the X variable
+        If a DataFrame is passed to fit and feature is None, feature
+        is selected as the column of the DataFrame.  There must be only
+        one column in the DataFrame.
+    target: string, default: None
+        The name of the Y variable
+        If target is None and a y value is passed to fit then the target
+        is selected from the target vector.
+    size: float, default: 600
+        Size of each side of the figure in pixels
+    ratio: float, default: 5
+        Ratio of joint axis size to the x and y axes height
+    space: float, default: 0.2
+        Space between the joint axis and the x and y axes
+    kwargs : dict
+        Keyword arguments that are passed to the base class and may influence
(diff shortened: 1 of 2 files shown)


yellowbrick/features/histogram.py
```diff
@@ -0,0 +1,191 @@
...
+        kwargs: dict
+            keyword arguments passed to Scikit-Learn API.
+        """
+
+        #throw an error if X has more than 1 column
+        if is_dataframe(X):
+            nrows, ncols = X.shape
+
+            if ncols > 1:
+                raise YellowbrickValueError((
```
bbengfort: If you have the feature name, and you have a DataFrame, can you grab the series from the DataFrame?

yellowbrick/features/histogram.py
```diff
@@ -0,0 +1,191 @@
...
+        return self
+
+
+    def poof(self, **kwargs):
+        """
+        Creates the labels for the feature and target variables
+        """
+
+        self.ax.set_xlabel(self.feature)
+        self.ax.set_ylabel(self.target)
```
bbengfort: Generally speaking, we don't override `poof` since the superclass does work for showing the visualization in notebooks or in a script, my advice is to add the label setting to the finalize method.

yellowbrick/features/histogram.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+        Finalize executes any subclass-specific axes finalization steps.
+        The user calls poof and poof calls finalize.
+        Parameters
+        ----------
+        kwargs: generic keyword arguments.
+        """
+
+        plt.setp(self.ax.get_xticklabels(), visible=True)
+        plt.setp(self.ax.get_yticklabels(), visible=True)
```
bbengfort: Because visualizers may be working on a specific axes object in a multi-axes figure, I usually prefer to contain all drawing to `self.ax` -- in this case I believe it's:

```python
for tk in self.ax.get_xticklabels():
    tk.set_visible(True) 
```

I know that's a bit more work, but it makes the visualizer much more robust and safer to use in a notebook environment.

Juan0001:
@bbengfort Thank you so much for your comments. I feel so lucky that I could get all these recommendations from you. Since I'm very new to this project I know there are a lot I need to improve. I think the target package is a very good idea. I will modify my codes according to your suggestions.

pdamodaran:
@Juan0001 - fyi - I have your codebase for this visualizer - i will make the changes as suggested by @bbengfort and reopen the PR and tag you as a contributor to this visualizer.

PR #415, opened 2018-05-14
added colorization to negative bar chart

### LINKED ISSUE #322: Improvements for FeatureImportances visualizer

kim81682:
Fixes #322 

[Description]
addressed just the first item

-the previous bar graph was commented out. 2 additional lines were added: one for color, the second for graphing the color if the item is negative.
-
-

### DIFF
--- a/yellowbrick/features/importances.py
+++ b/yellowbrick/features/importances.py
@@ -86,7 +86,7 @@ class FeatureImportances(ModelVisualizer):
     >>> visualizer.poof()
     """
 
-    def __init__(self, model, ax=None, labels=None, relative=True,
+    def __init__(self, model, ax=None, labels=None, relative=True, color_neg='b', color_pos='g',
                  absolute=False, xlabel=None, **kwargs):
         super(FeatureImportances, self).__init__(model, ax, **kwargs)
 
@@ -95,6 +95,9 @@ class FeatureImportances(ModelVisualizer):
             labels=labels, relative=relative, absolute=absolute,
             xlabel=xlabel,
         )
+        self.color_neg = color_neg
+        self.color_pos =color_pos
+
 
     def fit(self, X, y=None, **kwargs):
         """
@@ -168,9 +171,12 @@ class FeatureImportances(ModelVisualizer):
         pos = np.arange(self.features_.shape[0]) + 0.5
 
         # Plot the bar chart
-        colors = np.array(['b' if v > 0 else 'g' for v in self.feature_importances_])
+        #raise Exception(self.feature_importances_)
+        # Plot the bar chart
+
+
+        colors = np.array([self.color_pos if v > 0 else self.color_neg for v in self.feature_importances_])
         self.ax.barh(pos, self.feature_importances_, color=colors, align='center')
-        #self.ax.barh(pos, self.feature_importances_, align='center')
 
         # Set the labels for the bars
         self.ax.set_yticks(pos)
--- a/yellowbrick/features/importances.py
+++ b/yellowbrick/features/importances.py
@@ -168,7 +168,9 @@ class FeatureImportances(ModelVisualizer):
         pos = np.arange(self.features_.shape[0]) + 0.5
 
         # Plot the bar chart
-        self.ax.barh(pos, self.feature_importances_, align='center')
+        colors = np.array(['b' if v > 0 else 'g' for v in self.feature_importances_])
+        self.ax.barh(pos, self.feature_importances_, color=colors, align='center')
+        #self.ax.barh(pos, self.feature_importances_, align='center')
 
         # Set the labels for the bars
         self.ax.set_yticks(pos)


yellowbrick/features/importances.py
```diff
@@ -168,7 +171,12 @@ def draw(self, **kwargs):
         pos = np.arange(self.features_.shape[0]) + 0.5
 
         # Plot the bar chart
-        self.ax.barh(pos, self.feature_importances_, align='center')
+        #raise Exception(self.feature_importances_)
+        # Plot the bar chart
+
+
+        colors = np.array([self.color_pos if v > 0 else self.color_neg for v in self.feature_importances_])
```
kim81682: Fixes "color negative coefs" item

lwgray:
Thank you for joining us for the Yellowbrick Sprint at PyCon Cleveland. And also, thank you for submitting your first PullRequest - oh yeah! 

We will review this asap. We have been quite busy with all the great contributions from sprinters like yourself.  In the meantime, feel free to work on another issue.  We hope you continue to contribute.

yellowbrick/features/importances.py
```diff
@@ -168,7 +171,12 @@ def draw(self, **kwargs):
         pos = np.arange(self.features_.shape[0]) + 0.5
 
         # Plot the bar chart
-        self.ax.barh(pos, self.feature_importances_, align='center')
+        #raise Exception(self.feature_importances_)
```
ndanielsen: Thank you so much for working with on during sprints! I really enjoyed meeting you! One item to fix here, can you remove these comments?

yellowbrick/features/importances.py
```diff
@@ -86,7 +86,7 @@ class FeatureImportances(ModelVisualizer):
     >>> visualizer.poof()
     """
 
-    def __init__(self, model, ax=None, labels=None, relative=True,
+    def __init__(self, model, ax=None, labels=None, relative=True, color_neg='b', color_pos='g',
```
ndanielsen: One minor change, can you please update the documentation for this visualizer class? It should be a pretty straightforward change located in this section of the code:

https://github.com/kim81682/yellowbrick/blob/3489dd1a7b97901919e1bc2fcbafbaa8a5a20cc9/yellowbrick/features/importances.py#L55-L70

Again, thanks for contributing this!

kim81682:
Absolutely! Thank you for the experience. I still have a great deal to learn but would love to continue contributing/learning with your team :)
~kin


Sent with BlackBerry Work
(www.blackberry.com)

yellowbrick/features/importances.py
lwgray: Could this line be changed to reflect that we consider Zero to be a positive number and should be colored as such.  

`if v >= 0`

lwgray:
@kim81682  Is there anything we can do to help you complete this PR?

@ndanielsen

Issue #416, opened 2018-05-14
Unable to parse meta.json files from  the mushroom and bikeshare datasets


RaulPL:
Hi,

I downloaded all the data defined in this script https://github.com/DistrictDataLabs/yellowbrick/blob/develop/examples/download.py. Most of them have a meta.json file inside their folder but for two of them (mushroom and bikeshare datasets) I have a JSONDecodeError exception when I try to parse it with the json module of the standard library. 

### Code

This is an example code that can reproduce the error.
 ```python
import json

with open('data/bikeshare/meta.json') as file:
    meta_data = json.load(file)
meta_data
```

navarretedaniel:
I took a look at this parsing issue and think I may have an idea on how one might approach this.

I warn that it likely won't be an elegant solution, but if anything, it might be a start for more advanced Python coders.

lwgray:
@navarretdaniel we’re glad you’re  going to work on this and we’re looking forward to seeing what you put together.

bbengfort:
@navarretedaniel our new datasets mechanism resolves these issues, but thank you so much for your help!

bbengfort:
@RaulPL the datasets have been completely overhauled and the meta.json is now well formed for both of these datasets; thank you for contributing this issue to yellowbrick!

PR #417, opened 2018-05-14
Adding functions to load the available datasets. Similar to sklearn

### LINKED ISSUE #219: Data Loading
See: http://scikit-learn.org/stable/datasets/index.html

And #203 and (other issue we don't remember right now)

RaulPL:
Fixes #219 

Description
Adds a function for each dataset used in the examples (https://github.com/DistrictDataLabs/yellowbrick/blob/develop/examples/examples.ipynb). Example:
```python
data = load_concrete()
```

- Moving the download.py script to a new package called datasets.
- These functions are defined in the download.py module.
- In order to work properly these functions need to have the three files that are mostly downloaded: meta.json, datset_name.csv and README.md. At this moment load_bikeshare and load_mushroom wont work because the python json module is unable to parse their the meta.json files (issue #416) 
- Also the load_hobbies wont work because the dataset does not have a meta.json file


yellowbrick/datasets/download.py
```diff
@@ -0,0 +1,274 @@
+#!/usr/bin/env python
+# download
+# Downloads the example datasets for running the examples.
+#
+# Author:   Rebecca Bilbro <rbilbro@districtdatalabs.com>
```
ndanielsen: @RaulPL can you add your name here? you made some big changes to this file

ndanielsen:
@RaulPL looks like there is still a flake error, you're very close on this one!

screen shot 2018-05-15 at 12 38 08 pm

Issue #230, opened 2017-05-21
Parallel Coordinates is getting slower


bbengfort:
There seems to be a dependency or something that has increased the amount of time that parallel coordinates takes (already a significant amount of time). I ran the tests on my local 2.7 and 3.5 environments and got approximately 24s-1m run times. However on Travis it took > 10m. 

pip freeze:

```
alabaster==0.7.9
Babel==2.3.4
coverage==4.2
cycler==0.10.0
docutils==0.12
funcsigs==1.0.2
imagesize==0.7.1
Jinja2==2.8
MarkupSafe==0.23
matplotlib==1.5.3
mock==2.0.0
nose==1.3.7
numpy==1.11.2
pbr==1.10.0
Pygments==2.1.3
pyparsing==2.1.10
python-dateutil==2.5.3
pytz==2016.7
scikit-learn==0.18
scipy==0.18.1
six==1.10.0
snowballstemmer==1.2.1
Sphinx==1.4.8
sphinx-rtd-theme==0.1.9
```

rebeccabilbro:
Need to test each combination of dependency versions since the last to identify the deprecation warning. Most likely candidates are, in order: numpy, scipy, matplotlib.

bbengfort:
Looks like it's matplotlib 2.0.2 -- I was on matplotlib 1.5.3 locally and travis is installing the latest version of mpl.

bbengfort:
Oh, actually numpy also upgraded ...

bbengfort:
Fixed in #420 and #447

Issue #73, opened 2016-10-08
Detect continuous/discrete target vals for color util


bbengfort:
Right now the ParallelCoordinates and RadViz do not work for regressions since it supports a class based visualization. 

Give these classes the opportunity to do a continuous space by using a sequential colormap and coloring the instance lines according to target value.

bbengfort:
Now that we have the `color_sequence`  function, we can create a binning method to assign continuous valued points to the number of classes specified by the colormap.

bbengfort:
Finding a solution to this has been tricky because discrete classes has been embedded into the logic of both RadViz and ParallelCoordinates. There are two primary features that continuous data requires:

1. A colormap that will bin values into a consistent space. 
2. Instead of a label legend, a colorbar that shows the ranges of the instances.

To handle this for now, I'm going to do the following:

- I'm going to create subclasses of ParallelCoordinates and RadViz called "ContinuousParallelCoordinates" and "ContinuousRadViz" respectively. These will be temporary implementations (hopefully) that allow us to use sequential color maps by default. 

I will then create two utility functions: 

- `is_continuous(y)`
- `is_discrete(y)`

Which will accept a vector and use rules to decide if the data is continuous or discrete. Not exactly how to implement this. I will then use these in the `radviz()` and `parallel_coordinates()` helper functions. 

This should solve the issue for now, but we should revisit it when we develop RadViz and ParallelCoordinates with more features and functionality.

bbengfort:
Ok, this ticket has been extremely difficult to work on -- the RadViz and Parallel Coordinates classes are entirely set up for class based visualization only (thanks to their being pulled from matplotlib. 

This needs to be looked at a bit more closely, and I stashed my changes (which touched nearly every module in yellowbrick). We'll have to push this back to another sprint.

bbengfort:
Some notes here:
http://bbengfort.github.io/snippets/2017/01/17/resolving-matplotlib-colors.html

bbengfort:
Sklearn apparently does have a type of target checker that returns continuous or classification as one of it's outputs:

https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/multiclass.py#L175

bbengfort:
A very lightweight version of this has been implemented in #399

bbengfort:
See #334 for package where this should be placed.

PR #420, opened 2018-05-14
Add random sampling and improve performance of Parallel Components viz

### LINKED ISSUE #59: Improve Parallel Coordinates

thekylesaurus:
Fixes #59 (partially)

Allows for random sampling via "shuffle" parameter, consistent with sklearn train/test split, for instance.

Improves performance by at least an order of magnitude for several test data sets.


thekylesaurus:
Performance analysis shown below is based on iris data set, where replicates were used to increase the data set size (as indicated by the factor on the x-axis)

perf_comp
perf_comp_ratio

thekylesaurus:
Unfortunately, the approach used results in a subtly modified style.  There is no longer alpha addition intra-class.  The argument could be made that the new style reduces intra-class information; or, inversely, the it reduces unnecessary business.  Is this style change going to be acceptable?  If not, I will abandon this request and simply submit version with only the random subsampling added.

current_style
new_style

bbengfort:
@thekylesaurus awesome PR, thank you so much for taking a look at this! I'm very impressed by the speed up graphs and the thorough presentation, you definitely addressed questions I may have had right off the bat.  

I'm going to start my review now, and I'll dive into the alpha issue as I review. In the meantime, the tests aren't passing due to image closeness errors. It appears that your PR does not include updated images for the sampling method -- my guess was that you were omitting them until the style question was resolved. However, it is fairly simple to synchronize actual and images, if you'd like to follow these steps:

1. Run the tests: `pytest tests/test_features/test_pcoords.py` 
2. In the root directory run the sync: `python -m tests.images tests/test_features/test_pcoords.py` 

This will move the test case images generated by the tests in `tests/actual_images` to the baseline comparison directory, `tests/baseline_images`. You can then add the baseline images (the actual images are git ignored), commit and push them.

bbengfort:
Oh, make sure you have the test dependencies installed; `pip install -r tests/requirements.txt`!

yellowbrick/features/pcoords.py
```diff
@@ -268,32 +295,28 @@ def draw(self, X, y, **kwargs):
...
-                self.ax.plot(x, row, color=colors[label], alpha=0.25, **kwargs)
+        x_separated = hstack([X, ones((nrows, 1))])
+        increments_separated = increments.copy()
+        increments_separated.append(None)
+        for label, color in colors.items():
+            x_in_class = x_separated[y == label, :]
+            increments_in_class = increments_separated * len(x_in_class)
+            if len(x_in_class) > 0:
+                self.ax.plot(increments_in_class, x_in_class.flatten(), label=label, color=colors[label],
+                             alpha=0.5, linewidth=1, **kwargs)
```
bbengfort: Would you mind walking me through this new drawing method? I can see that the overall goal here was to call the plot method once per class instead of once per row - a laudable goal and one that I can see dramatically improving the performance of the visualizer. It might help to have a few comments to guide other contributors to this method, and if I could get a more detailed sense of what you're doing I can work to apply it to other visualizers. 

The alpha is set here to 0.5 -- so I'm not sure I understand exactly the change in the opacity style. Given the two figures you presented, I believe the top figure is preferable because it shows darker regions of greater instance density on a per-class basis. Was the top figure generated with this code or the old code? 

Frankly, even if the bottom figure was generated with your revisions, I think these revisions are worth it to keep the opacity changes and we can work on detailed styling in another PR.

yellowbrick/utils/types.py
```diff
@@ -202,6 +202,27 @@ def is_dataframe(obj):
 isdataframe = is_dataframe
 
 
+def is_series(obj):
```
bbengfort: Thanks for adding this! It would also be helpful if you wouldn't mind creating some tests for the type checking; the tests are pretty well defined in `tests/test_utils/test_types.py` -- should be a matter of copy and paste.

thekylesaurus:
@bbengfort - I can also ssh to my work's linux box, but without certain windows conveniences, my abilities become a little bit handcuffed.  I'll try to make it happen.

Thanks for your support!

yellowbrick/features/pcoords.py
```diff
@@ -239,26 +257,35 @@ def draw(self, X, y, **kwargs):
         # Convert from dataframe
         if is_dataframe(X):
             X = X.as_matrix()
+        if is_series(y):
+            y = y.as_matrix()
 
-        # Choose a subset of samples
-        # TODO: allow selection of a random subset of samples instead of head
+        # Get the total number of observations
```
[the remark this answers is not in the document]
thekylesaurus: Good call.  Will do.

yellowbrick/features/pcoords.py
```diff
@@ -239,26 +257,35 @@ def draw(self, X, y, **kwargs):
...
+        nobs = len(X)
```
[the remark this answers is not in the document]
thekylesaurus: Interesting.  Sounds good.

yellowbrick/features/pcoords.py
```diff
@@ -239,26 +257,35 @@ def draw(self, X, y, **kwargs):
...
 
+        # Choose a subset of samples
         if isinstance(self.sample, int):
-            self.n_samples = min([self.sample, len(X)])
+            self.n_samples = min([self.sample, nobs])
```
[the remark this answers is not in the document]
thekylesaurus: I will add that test case.

yellowbrick/features/pcoords.py
```diff
@@ -239,26 +257,35 @@ def draw(self, X, y, **kwargs):
...
         elif isinstance(self.sample, float):
-            self.n_samples = int(len(X) * self.sample)
-        X = X[:self.n_samples, :]
+            self.n_samples = int(nobs * self.sample)
+
+        if (self.n_samples < nobs) and self.shuffle:
+            indices = choice(nobs, self.n_samples, replace=False)
+        else:
+            indices = slice(self.n_samples)
```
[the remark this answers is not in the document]
thekylesaurus: Just learned about it at PyCon2018!

yellowbrick/features/pcoords.py
```diff
@@ -239,26 +257,35 @@ def draw(self, X, y, **kwargs):
...
         if isinstance(self.sample, int):
-            self.n_samples = min([self.sample, len(X)])
+            self.n_samples = min([self.sample, nobs])
         elif isinstance(self.sample, float):
-            self.n_samples = int(len(X) * self.sample)
-        X = X[:self.n_samples, :]
+            self.n_samples = int(nobs * self.sample)
+
+        if (self.n_samples < nobs) and self.shuffle:
+            indices = choice(nobs, self.n_samples, replace=False)
```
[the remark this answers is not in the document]
thekylesaurus: I think it makes sense to add the sklearn idiomatic "statify" option to this class (and probably others).  As this PR is already getting a little heavy, I'll create a new issue.

yellowbrick/features/pcoords.py
```diff
@@ -239,26 +257,35 @@ def draw(self, X, y, **kwargs):
...
+            indices = slice(self.n_samples)
+        X = X[indices, :]
+        y = y[indices]
+
+        # Get shape of the sampled data
+        nrows, ncols = X.shape
 
         # Normalize
         if self.normalize is not None:
             X = self.normalizers[self.normalize].fit_transform(X)
```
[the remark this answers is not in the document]
thekylesaurus: Will do.

yellowbrick/features/pcoords.py
thekylesaurus: Sounds good.  I will add comments to the code and some documentation to this PR to clarify.

The top image was the "current" method.  The bottom image was the "new" method.

I'll revert back to 0.25 alpha.  The 0.5 seemed to give a better visual agreement with the former style.  Are you OK with the linewidth specification?  Again, this gives better stylistic agreement for the samples I investigated.

yellowbrick/utils/types.py
thekylesaurus: No problem!

yellowbrick/features/pcoords.py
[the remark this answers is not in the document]
thekylesaurus: @bbengfort - do you want this new ParallelCoordinates.fit() to return X and y?  Maybe only in some non-default verbose mode used for testing only?

thekylesaurus:
@bbengfort - the test_parallel_coords() test was failing due to my decision to have shuffle=True by default, consistent with sklearn test/train split subsampling.  I have now set shuffle=False by default, and it rectifies the problem.  

However, I would think the default behavior should be to randomly subsample.  Thoughts?

yellowbrick/features/base.py
```diff
@@ -218,7 +218,7 @@ def fit(self, X, y=None, **kwargs):
         # Store the classes for the legend if they're None.
         if self.classes_ is None:
             # TODO: Is this the most efficient method?
-            self.classes_ = [str(label) for label in set(y)]
+            self.classes_ = [label for label in set(y)]
```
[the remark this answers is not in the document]
bbengfort: I appreciate that - these are class labels, not classes and are generally used for display on the image. We do have a lot to figure out with this and is a fairly large topic that includes sequential vs. discrete target values #73 using label encoders, etc. Scikit-learn also has some `check_target` utilities that we'd also like to employ more meaningfully. 

I think if you handle this inside of `ParallelCoordinates.draw()` you'll be setting a pretty good example.

yellowbrick/features/pcoords.py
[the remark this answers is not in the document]
bbengfort: We do want to ensure the signatures for sklearn interface methods remain the same, for fit it is:

```python
class Estimator(object):
    
    def fit(self, X, y=None):
        return self     
```

This will ensure that helper methods such as `fit_transform` in pipelines continue to function as expected. 

To test, I recommend we mock `ParallelCoordinates.draw`, which `X` and `y` will be passed to, we can then inspect the output of `fit` there.

yellowbrick/features/pcoords.py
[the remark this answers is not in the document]
bbengfort: Awesome, thanks - I think that would be an excellent follow-on PR!

yellowbrick/features/pcoords.py
bbengfort: I'm ok with the `linewidth` argument -- in the past, we've had issues with fixed settings like this (e.g. in a "poster" mode of matplotlib where rcParams are set, this fixed value might cause some issues) but I think it will be ok for now.

bbengfort:
I agree that the default behavior should be to randomly subsample, perhaps we can tackle that in another PR so that the tests pass now?

tests/test_features/test_pcoords.py
```diff
@@ -78,6 +78,19 @@ def test_pcoords_sample_int(self):
...
+        Assert no errors occur using integer 'sample' argument and shuffle, with different random_state args
+        """
+        visualizer = ParallelCoordinates(sample=3, shuffle=True)
+        visualizer.fit_transform(self.X, self.y)
+
+        visualizer = ParallelCoordinates(sample=3, shuffle=True, random_state=444)
+        visualizer.fit_transform(self.X, self.y)
+
+        visualizer = ParallelCoordinates(sample=3, shuffle=True, random_state=np.random.RandomState())
+        visualizer.fit_transform(self.X, self.y)
```
bbengfort: Should we also add a `shuffle=False`?

tests/test_features/test_pcoords.py
```diff
@@ -92,6 +105,19 @@ def test_pcoords_sample_float(self):
...
+        Assert no errors occur using float 'sample' argument and shuffle, with different random_state args
+        """
+        visualizer = ParallelCoordinates(sample=0.5, shuffle=True)
+        visualizer.fit_transform(self.X, self.y)
+
+        visualizer = ParallelCoordinates(sample=0.5, shuffle=True, random_state=444)
+        visualizer.fit_transform(self.X, self.y)
+
+        visualizer = ParallelCoordinates(sample=0.5, shuffle=True, random_state=np.random.RandomState())
+        visualizer.fit_transform(self.X, self.y)
```
bbengfort: Same comments as above in the int sample. 

Also, do we want to ensure no shuffling occurs when there is no sampling? Or is this satisfied in `test_static_subsample`?

tests/test_features/test_pcoords.py
```diff
@@ -131,3 +157,65 @@ def test_integrated_pcoords(self):
         visualizer.fit_transform(X, y)
         visualizer.poof()
         self.assert_images_similar(visualizer)
+
+    @staticmethod
```
bbengfort: What does making this a static method do for the test case?

tests/test_features/test_pcoords.py
```diff
@@ -131,3 +157,65 @@ def test_integrated_pcoords(self):
...
+    def test_static_subsample():
+        """
+        Assert output of subsampling method against expectations
+        """
+
+        ntotal = 100
+        ncols = 50
+
+        y = np.arange(ntotal).reshape(ntotal, 1)
+        X = np.ones((ntotal, ncols)) * y
```
bbengfort: I very much like this idea for a test case, it definitely allows you to determine the ordering with ease! Note, however, that `y` should have shape `(100,)` and instead, it has shape `(100,1)`. Because you need this shape in order to compute `X`, I recommend the following:

```python
y = np.arange(ntotal)
X = np.ones((ntotal, ncols)) * y.reshape(ntotal,1) 
```

tests/test_utils/test_types.py
```diff
@@ -522,6 +522,40 @@ def test_not_is_dataframe(self, obj):
...
+        ),
+        np.array([1,2,3]),
+        [1, 2, 3],
+    ],
+    ids=["structured array", "array", "list"])
+    def test_not_is_series(self, obj):
+        """
+        Test that is_series does not match non-dataframes
+        """
+        assert not is_series(obj)
```
bbengfort: Thank you for adding these tests!

yellowbrick/features/pcoords.py
```diff
@@ -149,6 +152,15 @@ class ParallelCoordinates(DataVisualizer):
         If int, specifies the maximum number of samples to display.
         If float, specifies a fraction between 0 and 1 to display.
 
+    random_state : int, RandomState instance or None
```
bbengfort: Awesome +100 for this - thank you so much for thinking to add this in!

yellowbrick/features/pcoords.py
```diff
@@ -169,6 +181,15 @@ class ParallelCoordinates(DataVisualizer):
         Keyword arguments that are passed to the base class and may influence
         the visualization as defined in other Visualizers.
 
+    Attributes
+    --------
+
+    n_samples_ : int
+        number of samples included in the visualization object
+
+    rng_ : np.random.RandomState
```
bbengfort: This is a complete nit, and I apologize for it, but since I made a big deal of `n_samples_` I feel like I should point out that this is not a learned parameter in fit, but a private attribute mirroring `random_state` and should probably be `_rng`. If I didn't have any other comments I would have just ignored this and perhaps made a quick change after the PR, but perhaps you could quickly update it?

yellowbrick/features/pcoords.py
```diff
@@ -225,40 +246,96 @@ def __init__(self, ax=None, features=None, classes=None, normalize=None,
...
-        # Choose a subset of samples
-        # TODO: allow selection of a random subset of samples instead of head
-
-        if isinstance(self.sample, int):
-            self.n_samples = min([self.sample, len(X)])
-        elif isinstance(self.sample, float):
-            self.n_samples = int(len(X) * self.sample)
-        X = X[:self.n_samples, :]
+        # Subsample
+        X, y = self._subsample(X, y)
```
bbengfort: Making the sampling a separate function was a very good idea - nice!

yellowbrick/features/pcoords.py
```diff
@@ -225,40 +246,96 @@ def __init__(self, ax=None, features=None, classes=None, normalize=None,
...
 
         # Normalize
         if self.normalize is not None:
             X = self.normalizers[self.normalize].fit_transform(X)
 
-        # Get the shape of the data
+        super(ParallelCoordinates, self).fit(X, y, **kwargs)
+
```
bbengfort: Perhaps I'm missing this in the code review pane, but `fit()` should call `draw()` -- am I missing where this is happening?

bbengfort: Hehe, I think I got it, but I'm leaving my original comment above in - the super method is calling `draw()` right? Perhaps it would be more clear if we did something like:

```python
        # the super method calls draw and returns self 
        return super(ParallelCoordinates, self).fit(X, y, **kwargs)
```

bbengfort:
Your illustration of the current and new incrementation method is very well done, thank you. I understand in principle what you're doing - I guess I'm not sure why you had to add the null column at the end to actually create the darkening effect, I would have thought that if you put two semi-transparent lines on top of each other, they would have darkened rather than blending as the did. Anyway - great work on figuring out a solution, the performance benefits are definitely worth it!

tests/test_features/test_pcoords.py
thekylesaurus: Added test_pcoords_sample_shuffle_false()

tests/test_features/test_pcoords.py
thekylesaurus: Added test_pcoords_sample_float_shuffle_false()

Tested in test_static_subsample.  I would argue this is the best place because X and y are exposed on input and output.

tests/test_features/test_pcoords.py
thekylesaurus: PEP8-compliance.  That is about it.

tests/test_features/test_pcoords.py
thekylesaurus: Good call. Now done.

tests/test_utils/test_types.py
thekylesaurus: :+1:

yellowbrick/features/pcoords.py
thekylesaurus: :+1:

yellowbrick/features/pcoords.py
thekylesaurus: Done.  Removed reference to _rng from docstring.

yellowbrick/features/pcoords.py
thekylesaurus: :+1:

yellowbrick/features/pcoords.py
thekylesaurus: Comment added.  I should have had this in there.

thekylesaurus:
@bbengfort - comments addressed in most recent commits.

Pertaining to failed test, can you direct me to the test that is failing?  I am able to pass all tests in test_pcoords, including the image comparison test. 

Pertaining to the nans, they are in there to prevent the line plot from drawing from right to left.  If you didn't have the nans, there are be lines crossing from the right hand side of the image to the left-hand side of the window.

The nans don't address the alpha effect.  The lack of alpha darkening is the unfortunate result of matplotlib not accounting for alpha darkening for intersections of a single trace with itself.  I spent a solid hour trying to find a work-around, with no luck.

bbengfort:
Ah, sorry I forgot about the crossing lines thing - that was part of the reason we did the one line per instance implementation; the nans are a good solution! As for the alpha darkening, perhaps we can open an issue for that? I can't help feeling that there is something in matplotlib that might make it work, but if you've already spent hours on it, we should punt, the visualizer is working very well right now. 

The tests are passing for you? Well, that's good news sort of - what system are you on? 

I think potentially the easiest thing is for me to merge your PR and then try to fix the Travis tests myself, it's notoriously tricky and annoying. Are you ok with that? 

For reference, here are the tests that are failing:

```
_______________ ParallelCoordinatesTests.test_integrated_pcoords _______________
self = <tests.test_features.test_pcoords.ParallelCoordinatesTests testMethod=test_integrated_pcoords>
    @pytest.mark.xfail(
        sys.platform == 'win32', reason="images not close on windows"
    )
    def test_integrated_pcoords(self):
        """
            Test parallel coordinates on a real data set (downsampled for speed)
            """
        occupancy = self.load_data('occupancy')
    
        X = occupancy[[
            "temperature", "relative_humidity", "light", "C02", "humidity"
        ]]
    
        y = occupancy['occupancy'].astype(int)
    
        # Convert X to an ndarray
        X = X.copy().view((float, len(X.dtype.names)))
    
        # Test the visualizer
        visualizer = ParallelCoordinates(sample=200)
        visualizer.poof()
>       self.assert_images_similar(visualizer)
E       yellowbrick.exceptions.ImageComparisonFailure: images not close (RMS 26.357)
E       tests/actual_images/test_features/test_pcoords/test_integrated_pcoords.png
E       	vs
E       tests/baseline_images/test_features/test_pcoords/test_integrated_pcoords.png
tests/test_features/test_pcoords.py:185: ImageComparisonFailure
_______________ ParallelCoordinatesTests.test_normalized_pcoords _______________
self = <tests.test_features.test_pcoords.ParallelCoordinatesTests testMethod=test_normalized_pcoords>
    @pytest.mark.xfail(
        sys.platform == 'win32', reason="images not close on windows"
    )
    def test_normalized_pcoords(self):
        """
            Assert no errors occur using 'normalize' argument
            """
        visualizer = ParallelCoordinates(normalize='l2')
        visualizer.fit_transform(self.X, self.y)
        visualizer.poof()
>       self.assert_images_similar(visualizer)
E       yellowbrick.exceptions.ImageComparisonFailure: images not close (RMS 0.074)
E       tests/actual_images/test_features/test_pcoords/test_normalized_pcoords.png
E       	vs
E       tests/baseline_images/test_features/test_pcoords/test_normalized_pcoords.png
tests/test_features/test_pcoords.py:65: ImageComparisonFailure
```

thekylesaurus:
My machine is a Windows 7 Enterprise on Dell Latitude E5570 with Intel HD Graphics 530 and AMD Radeon R7 M370.

I did more digging.  In the current code base, both of the failing tests indicated above had a wrapper applied to them:

    @pytest.mark.xfail(
        sys.platform == 'win32', reason="images not close on windows"
    )

This wrapper evidently skips these tests when run on a windows machine.

When I removed these wrappers, I actually passed the test_normalized_pcoords() test.

I failed the test_integrated_pcoords() test.  However, not because the image match failed, but because my computers firewall is blocking the load_data('occupancy') step: "SSLError: ("bad handshake: SysCallError(10054, 'WSAECONNRESET')",)"

I am in favor of your merging, but I also would express caution.  Some would argument that this PR modifies the look and feel of the ParallelCoordinates plot (due to the removal of intraclass alpha darkening).  If you think the speed up increase it worth it, please take the lead on merging and fixing the tests.

Thanks @bbengfort

bbengfort:
@DistrictDataLabs/team-oz-maintainers @thekylesaurus suggests that we consider carefully merging this parallel coordinates implementation because it changes the style of the plots:

The current, poorly performing style plots each instance as it's own line one at a time, allowing us to achieve a darkening effect where more instances are:

current_style

The new implementation draws each class as a single line, breaking them up with nan values so there is no cross back and forth from the end of the parallel coordinates to the beginning. However, because these lines are the "same line", there is no darkening effect:

new_style

The performance boost from the proposed implementation is significant, especially as more instances are added to the plot (almost a 30x improvement for 100x the iris data set). 

Any thoughts on how to proceed?

rebeccabilbro:
@bbengfort - I recommend we make this an alternate visualizer, `FastParallelCoordinates`, and keep the currently (slow) visualizer as `ParallelCoordinates`; we can use the docstrings and documentation for each to explain the tradeoffs to users and leave it up to them to decide which is best for their use case.

lwgray:
Is Key information being lost by not showing the density?

thekylesaurus:
What about a flag in the existing visualizer?

bbengfort:
@rebeccabilbro @thekylesaurus I think leaving it up to the user is probably the best thing to do; I'd say it's a tough call between overriding the draw method in a subclass and using a flag since the implications of the fast method are fairly subtle. For simplicity, I'm leaning towards using a flag in the constructor `fast=False` with a short explanation in the docstring about what's happening. 

My feeling is that the usage will be that the user tries `ParallelCoordinates` without the `fast` flag, gets aggravated it's taking so long, sets the fast flag and tries again. 

We should also create a section in the documentation with the graphs @thekylesaurus created and the speed up to show the implications in detail.

bbengfort:
I will take the following steps when I get the chance:

1. [ ] create an issue to fix tests/add fast and docs 
2. [ ] merge this PR 
3. [ ] update #230 and #59 
4. [ ] update tests and add fast submit PR 

Hopefully tomorrow or the day after. I would just merge this in now but am reluctant to because the tests are failing, and I'd like to get them sorted ASAP so other contributors don't have failing tests related to parallel coordinates.

bbengfort:
See #447

bbengfort:
@thekylesaurus thank you so much for this PR and your hard work; we're glad to have you as a contributor, and hope you'll consider working with us more in the future!

Issue #380, opened 2018-04-03
In a Jupyter notebook, no display when poof is in a separate cell


fdion:
if visualizer.poof() is called in a separate cell, no visualization appears, as the empty plot has already been rendered in the call for fit(). Workaround is to access visualizer.ax.get_figure().

### Issue

For extra detail, see demo notebook linked in Background section

### Code Snippet

Tested using the bike share data set from the YB intro

In one cell:
```
visualizer = Rank1D()
visualizer.fit(X,y)
visualizer.transform(X)
```

In a separate cell:
`visualizer.poof()`

### Background

Demo notebook
https://gist.github.com/fdion/d645e164f6969d269aff681efa31ef96

fdion:
#375 would fix the display when poof is called (I tested it by adding return ax to poof), but it wont prevent the display of the empty figure when fit is called.

bbengfort:
This is a tough one; poof is a wrapper for `visualizer.finalize()` + `visualizer.show()` or `visualizer.savefig()`. Perhaps one work around is to use `visualizer.finalize()` instead of `visualizer.poof()` so the call to show doesn't execute?

Naba7:
As in https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/base.py
the fit method returns self,the draw method returns nothing and checks a Not Implemented Error,the finalize method returns self.ax.While poof method wraps up finalize method at first which returns self.ax and a follow up if-else statement wraps either plt.show method or plt.savefig method.
    def poof(self, outpath=None, clear_figure=False, **kwargs):
        """
        Poof makes the magic happen and a visualizer appear! You can pass in
        a path to save the figure to disk with various backends, or you can
        call it with no arguments to show the figure either in a notebook or
        in a GUI window that pops up on screen.
        Parameters
        ----------
        outpath: string, default: None
            path or None. Save figure to disk or if None show in window
        clear_figure: boolean, default: False
            When True, this flag clears the figure after saving to file or
            showing on screen. This is useful when making consecutive plots.
        kwargs: dict
            generic keyword arguments.
        Notes
        -----
        Developers of visualizers don't usually override poof, as it is
        primarily called by the user to render the visualization.
        """
        # Ensure that draw has been called
        if self._ax is None:
            warn_message = (
                "{} does not have a reference to a matplotlib.Axes "
                "the figure may not render as expected!"
            )
            warnings.warn(
[...]

PR #422, opened 2018-05-15
[WIP] Make poof() return self.ax

### LINKED ISSUE #375: Poof() should return ax
Make sure that ll visualizers return the `self.ax` when calling poof(). This is so that users can always get access to the ax, for example when working inside a function in a notebook. In addition it will encourage the behavior pattern of tweaking the plot if desired after poof(). We will be adjusting our documentation to use this form to also encourage this behavior:

```
ax = viz.poof()
```

This why this behavior pattern is useful:
https://stackoverflow.com/questions/47450804/yellowbrick-increasing-font-size-on-yellowbrick-generated-charts

jtpio:
Initial work to make the `poof()` method return `self.ax`.

Fixes #375 


NealHumphrey:
@jtpio looks simple and straightforward. Were you able to search the codebase to see if any visualizers override the `poof()` method though? I can approve if you confirm there was at least a quick check for this. 

I checked on the failed appveyor integration and this was the reason, so should be safe to ignore: `Build execution time has reached the maximum allowed time for your plan (60 minutes).`. Python 3 test completed, Python 2 did not. 

This is very likely overkill for this issue, but for enforcing API uniformity for situations like this it would be ideal to include a test case that tests all subclasses of Visualizer, so if anyone overrides it'll catch it in the tests. @bbengfort or @ndanielsen I think we talked about this in the past but don't remember if we found a solution or just decided to keep watching out for things manually? If there's a working example of other unit tests that do API enforcement like this we could imitate, otherwise we can leave alone.

jtpio:
Hi @NealHumphrey and @bbengfort,

Thanks for taking the time to review this change.
I think a proper pass on the over the codebase is still needed. Also this change alone doesn't seem to have any impact on https://github.com/DistrictDataLabs/yellowbrick/issues/380.

Will allocate time for a second look later this month.

bbengfort:
@jtpio that would be great; thanks! let us know when you have a chance to look at it and if you have any questions!

rebeccabilbro:
Hey there @jtpio - since the conversation has gone a bit stale here, I'm going to go ahead and close this one. Please feel free to open up a new PR as soon as you're ready to get back into this. We'd love to have you contribute!

PR #428, opened 2018-05-15
added test for postag

### LINKED ISSUE #395: xfail if nltk data has not been downloaded (move postag to contrib)
Right now we have tests for `yellowbrick.text` that have an nltk depedency (the pos tag visualizer I believe, I don't think the others rely on it). We skip the tests if nltk is not installed, but we do not test if nltk data has been downloaded or is available. We should test this or wrap in a try/except and xfail to ensure that this doesn't fail tests for new users. 

Further, we could move nltk dependencies to the contrib directory, where we could more easily manage the test cases.

sayali-sonawane:
Fixes #395

Added a test case for checking the nltk postag data availability on user's machine. Xfail the test if dataset is not downloaded else Xpass the test. 

- Checked only for 'treebank' postag dataset from nltk

Question:

- Should other postag datasets need to be checked?


lwgray:
Thank you for joining us for the Yellowbrick Sprint at PyCon Cleveland. And also, thank you for submitting your first PullRequest - oh yeah!

We will review this asap. We have been quite busy with all the great contributions from sprinters like yourself.  However, feel free to work on another issue.  We hope you continue to contribute.

PR #444, opened 2018-05-19
Issue #374 - Add CVScores.py under model_selection

### LINKED ISSUE #374: CVScores
Implement a visualizer that shows cross-validation scores as a bar chart along with the final score as an annotated horizontal line. 

We want to start moving toward better cross-validation and model selection. Create `yb.model_selection.CVScores` visualizer that extends `ModelVisualizer` and wraps an estimator. Accepts a `cv` and `scoring` params, similar to the ones exposed in [`sklearn.model_selection.cross_val_score`](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_val_score.html). 

Output is a bar chart with the score of each split and a horizontal dotted annotation with the average score.

pdamodaran:
Fixes #374 

[Added CVScores Visualizer]

-
-
-


yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,145 @@
...
+
+from yellowbrick.base import ModelVisualizer
+from sklearn.model_selection import cross_val_score
+
+##########################################################################
+## CVScores Visualizer
+##########################################################################
+
+class CVScores(ModelVisualizer):
+		"""
```
lwgray: It appears that the spacing/indentation if incorrect (ie greater than 4 spaces)  ????

yellowbrick/classifier/class_balance.py
```diff
@@ -25,6 +25,9 @@
 
 from .base import ClassificationScoreVisualizer
 
+from .base import ClassificationScoreVisualizer
```
lwgray: What is the Class Balance visualizer being altered?  Why is line 28 a duplicate of line 26?

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    def __init__(self, model, ax=None, cv=None, scoring=None, **kwargs):    
+        """
+        Initialize the visualization with many of the options required
+        in order to make most visualizations work.
+
+        These parameters can be influenced later on in the visualization
+        process, but can and should be set as early as possible.
+
+        Parameters
```
lwgray: I think the parameters docstring should be moved out of the __init__ function and placed just above... See how @bbengfort set his up in the validation curve visualizer. 

https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/model_selection/validation_curve.py

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,151 @@
...
+
+        """
+
+        self.cv_scores_ = cross_val_score(self.estimator, X, y, cv=self.cv, scoring=self.scoring)
+        self.cv_scores_mean_ = np.mean(self.cv_scores_)
+
+        self.draw()
+        return self
+
+    def draw(self, **kwargs):
```
lwgray: I really like how you got this to work. great job!  The code just needs a bit better commenting.

yellowbrick/classifier/class_balance.py
```diff
@@ -25,6 +25,8 @@
 
 from .base import ClassificationScoreVisualizer
 
+from ..style.palettes import color_palette
```
bbengfort: I think that lines 28 and 29 are remainders from previous work; but are causing the tests to fail:

```
________________________________ pyflakes-check ________________________________
/home/travis/build/DistrictDataLabs/yellowbrick/yellowbrick/classifier/class_balance.py:28: UnusedImport
'..style.palettes.color_palette' imported but unused
```

Would you mind removing these lines, I recommend the following:

```
$ git checkout yellowbrick/classifier/class_balance.py
```

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,174 @@
...
+#
+# ID: CVScores.py [7f47800] pdamo24@gmail.com $
+
+##########################################################################
+## Imports
+##########################################################################
+
+import warnings
+import numpy as np
+import matplotlib as mpl
```
bbengfort: These unused imports are the second place the test failed:

```
/home/travis/build/DistrictDataLabs/yellowbrick/yellowbrick/model_selection/CVScores.py:21: UnusedImport
'warnings' imported but unused
/home/travis/build/DistrictDataLabs/yellowbrick/yellowbrick/model_selection/CVScores.py:23: UnusedImport
'matplotlib as mpl' imported but unused
```

Would you mind removing these two imports (`import warnings` and `import matplotlib as mpl`)?

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,174 @@
...
+# yellowbrick.model_selection.CVScores
+#
+#
+# Author:   Prema Damodaran Roman
+
+#
+# Copyright (C) 2018 District Data Labs
+# For license information, see LICENSE.txt
+#
+# ID: CVScores.py [7f47800] pdamo24@gmail.com $
```
bbengfort: I've been thinking about the name of this file; in the scikit-learn repo, this module is simply called `sklearn.model_selection._validation`; what would you think about the more pythonic `cv.py` for this file? The `CVScores` visualizer would be imported in `__init__.py` -- another thing I'd recommend adding ;-)

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,174 @@
...
+    model : a scikit-learn estimator
+            An object that implements ``fit`` and ``predict``, can be a
+            classifier, regressor, or clusterer so long as there is also a valid
+            associated scoring metric.
+            Note that the object is cloned for each validation.
+
+    ax : matplotlib.Axes object, optional
+            The axes object to plot the figure on.
+
+    cv : int, cross-validation generator or an iterable, optional
```
bbengfort: Very nice!

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,174 @@
...
+
+            see the scikit-learn
+            `cross-validation guide <http://scikit-learn.org/stable/modules/cross_validation.html>`_
+            for more information on the possible strategies that can be used here.
+
+    scoring : string, callable or None, optional, default: None
+            A string or scorer callable object / function with signature
+            ``scorer(estimator, X, y)``. 
+    
+            See scikit-learn model evaluation documentation for names of possible metrics.
```
bbengfort: As in the `cv` param; could we include a link to the scikit-learn documentation?

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,174 @@
...
+    
+    Notes
+    -----
+    
+    This visualizer is a wrapper around for the ``sklearn.model_selection.cross_val_score``
+    <<http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_val_score.html>>
+    
+    Refer to the scikit-learn
+    `cross-validation guide <http://scikit-learn.org/stable/modules/cross_validation.html>`
+    for more details
```
bbengfort: Very good note!

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,174 @@
...
+                None for unsupervised learning.
+
+        Returns
+        -------
+        self : instance
+        
+        """
+
+        self.cv_scores_ = cross_val_score(self.estimator, X, y, cv=self.cv, scoring=self.scoring)
+        self.cv_scores_mean_ = np.mean(self.cv_scores_)
```
bbengfort: Thank you for storing these properties on the visualizer for user access!

bbengfort: Also note, you can use `self.cv_scores_.mean()` here if you'd like.

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,174 @@
...
+
+        self.draw()
+        return self
+
+    def draw(self, **kwargs):
+        """
+        creates the bar chart of the CV scores generated from the fit method and places 
+        a dashed horizontal line that represents the average value of the CV scores
+        """
+        minimum = min(self.cv_scores_)
```
bbengfort: This is completely minor, but just a suggestion as it pertains to our use of NumPy in general; `self.cv_scores_` is a `numpy.ndarray` (and we certainly want to guarantee it is in the future). When dealing with these arrays, it is preferable for a variety of reasons to use the NumPy methods, e.g. `self.cs_scores_.min()`. 

For me, it was mostly about getting into the mindset of using NumPy first and foremost whenever possible; it probably doesn't matter a lot for this exact line of code, but the more Yellowbrick uses NumPy, the faster it is!

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,174 @@
...
+        if minimum > 0.05:
```
bbengfort: A comment here about why you're doing this would be helpful; CV scores are generally between `[0.0, 1.0]`, but note that R2 scores can be big negative numbers.

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,174 @@
...
+            minimum = minimum - 0.05 
+        plt.ylim(minimum, 1)
```
bbengfort: Would you mind changing this to `self.ax.set_ylim(minium, 1)`? -- using the axes object on the visualizer ensures that weirdness doesn't occur. Also, if you remove this reference to `plt` you might also have to remove the import at the top to get the tests to pass.

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,174 @@
...
+        xvals = range(1, len(self.cv_scores_) + 1, 1)
```
bbengfort: Again, completely minor, but here you could also use `np.arange` to get a numpy array of xvals.

yellowbrick/model_selection/CVScores.py
lwgray: I agree.  This is what I was trying to say about adding comments on the other PR you posted.  I'm sorry that I didn't convey it clearly

yellowbrick/model_selection/CVScores.py
```diff
@@ -0,0 +1,174 @@
...
+        self.ax.legend(frameon=True, loc=loc, edgecolor=edgecolor)
+
+        #set spacing between the x ticks
+        self.ax.xaxis.set_major_locator(ticker.MultipleLocator(1))
+
+        # Set the axis labels
+        self.ax.set_xlabel('Training Instances')
+        self.ax.set_ylabel('Score')
+
+
```
lwgray: Very nice work on the visualizer. I caught a couple of things. Let me know if you need me to clarify any of this.

1. CVScores.py is missing a "quick method". Look at the method found at end of the validationcurve.py code for an example of what I mean

2. I cloned your repo and tried to run the example code however I get an error that basically said that the CVScore is not callable. I found out why and the solution is to edit the__init__.py file that is in the same directory as CVScores.py. You need to add an import statement similar to the ones already present in __init__.py.  This is the reason why @bbengfort suggested the same in his prior comment, " The CVScores visualizer would be imported in __init__.py   -- another thing I'd recommend adding ;-)"

pdamodaran: @bbengfort and @lwgray - thanks for your feedback.  I will make the changes you suggested.

yellowbrick/classifier/class_balance.py
pdamodaran: @bbengfort line 29 is empty - did you mean just remove line 28?

yellowbrick/model_selection/CVScores.py
pdamodaran: @bbengfort cv.py sounds good to me.  Could you explain what I need to do in the __init__.py code?

pdamodaran: this is what I added to __init__.py - is this correct (after making the change to call the file cv.py)?

from .cv import CVScores, cv

lwgray: that looks right to me!  @bbengfort

pdamodaran:
This is what I put under the quick method section - @bbengfort and @lwgray - could you look at this to see if this correct?

## Quick Method
        
    def cv(model, X, y, ax=None, cv=None, scoring=None, **kwargs):
        
        """
        Displays cross validation scores as a bar chart and the 
        average of the scores as a horizontal line
        
        This helper function is a quick wrapper to utilize the
        CVScores visualizer for one-off analysis.

        Parameters
        ---------- 

        model : a scikit-learn estimator
                An object that implements ``fit`` and ``predict``, can be a
                classifier, regressor, or clusterer so long as there is also a valid
                associated scoring metric.
                Note that the object is cloned for each validation.

        ax : matplotlib.Axes object, optional
                The axes object to plot the figure on.

        cv : int, cross-validation generator or an iterable, optional
                Determines the cross-validation splitting strategy.
                Possible inputs for cv are:
                    - None, to use the default 3-fold cross-validation,
                    - integer, to specify the number of folds.
                    - An object to be used as a cross-validation generator.
                    - An iterable yielding train/test splits.

                see the scikit-learn
                `cross-validation guide <http://scikit-learn.org/stable/modules/cross_validation.html>`_
                for more information on the possible strategies that can be used here.

        scoring : string, callable or None, optional, default: None
                     A string or scorer callable object / function with signature
                     ``scorer(estimator, X, y)``. 
        
                     See scikit-learn `cross-validation guide <http://scikit-learn.org/stable/modules/cross_validation.html>`_
                     for more information on the possible metrics that can be used.

        kwargs : dict
        
        Returns
        -------
        ax : matplotlib.Axes
            The axes object that the validation curves were drawn on.
                    
        """
        # Initialize the visualizer
        oz = cv(model, X, y, ax=ax, cv=cv, scoring=scoring)
        
        # Fit and poof the visualizer
        oz.fit(X, y)
        oz.poof(**kwargs)
        return oz.ax

lwgray:
@pdamodaran The quick method looks amazing.  Just a very small thing... I don't know how @bbengfort feels about this but instead of 'oz' you might want to use 'visualizer' for continuity with other visualizers.  This is extremely minor and just my opinion.

yellowbrick/classifier/class_balance.py
bbengfort: Only line 28 is causing the tests to fail, but I do recommend the removal of both lines so that no changes are made to this file, which has no impact on your PR at all. Again, you can reset this file with:

```
$ git checkout yellowbrick/classifier/class_balance.py 
```

yellowbrick/model_selection/CVScores.py
bbengfort: Agreed, that's correct - but I don't see any changes to `__init__.py` -- did you push it?

bbengfort:
@pdamodaran the quick method looks great, though I'd recommend calling it `cv_scores()` instead of `cv()`. 

@lwgray you're probably correct, I'm the cowboy that does `oz` over `visualizer` though... I guess I should probably stop.

yellowbrick/model_selection/CVScores.py
pdamodaran: not yet - i wanted to confirm before i push

bbengfort:
@pdamodaran looking great! I'm going to merge this with the latest version of the `develop` branch to make sure the tests pass -- you should probably do a pull on your side to get the latest updates.

yellowbrick/model_selection/__init__.py
```diff
@@ -16,3 +16,4 @@
 
 from .learning_curve import LearningCurve, learning_curve
 from .validation_curve import ValidationCurve, validation_curve
+from .cv import CVScores, cv
```
bbengfort: Thanks for including these imports at the top level! I think this is where the tests are failing, this should be:

```python
from .cv import CVScores, cv_scores
```

Issue #353, opened 2018-03-20
Script to autogenerate images in docs


ndanielsen:
Currently in our `docs` folder, we have code that generates png files to be used in the generated docs. It would be helpful to have a script to generate images and save them in the appropriate folder location. 

related to: https://github.com/DistrictDataLabs/yellowbrick/issues/38

rebeccabilbro:
Closing in favor of #687 &mdash; which is very close to done! 😄

PR #446, opened 2018-05-23
Auto-generate Images for Documentation

### LINKED ISSUE #354: Scripts to Regenerate Tutorial and Quickstart Documentation Images
In the documentation, the tutorial and quickstart need a script that can autogenerate the images similar to how they are done in the `doc/api/` section. The code to generate the images is located in the rst files and can be simple extracted and moved into a py file.

These scripts can live right alongside their .rst compatriots.

rwhitt2049:
[EDIT]: [Plot directive documentation](https://matplotlib.org/devel/plot_directive.html)

Fixes #354, #353 

Here's a wall of text! I'd like to get your feedback on my proposal of using matplotlib's sphinx extension. I only did a few of the visualizations for now.

Proposal for adding documentation image auto-update functionality. Currently, the images are provided via a py script that saves an image to an images folder. That code is then duplicated into the documentation and the image is embedded. 

I propose using matplotlib's sphinx extension to embed the requisite plotting code, which will then directly embed the rendered images after the code blocks. See elbow.rst and importances.rst for an example implementation.

@ndanielsen and I discussed a different solution at PyCon that included making the docs folder a module and then developing an image rebuild script. After a quick proof of concept, I decided to go this route as the module route would be hard to maintain and missed the mark for automating the documentation.

Some pros:
- Code is deduplicated. No more scripts to build a viz, then having to remember to port the script to the docs or remembering to rerun the scripts to rebuild images.
- Visualization code lives with the visualization
- Reduces burden of anyone contributing to documentation. They won't have to build the script, remember to add to docs, and remember to embed the png. With this approach, they only need to make sure to use `.. plot::` instead of `.. code::`

Some cons:
- The `.. plot::` directive can't be split up like the `.. code::`. In practice, this means that you won't be able to add commentary between code blocks. See `importances.rst` for example of how this will look (screenshot below is the after picture of this)
- The `..plot::` directive doesn't support as many features as the `.. code::` directives: line numbers, line emphasis, etc (This could be mitigated by doing both a plot and a code directive, then suppressing the code output for the plot. You'll have to duplicate the code in this case, but may be a reasonable compromise)
- It looks like it may override your css? Note the yellow background to the code. This may be fixable, but I haven't researched it yet.

Screenshot to show consequence of not being able to split code directives. (Here's your current [docs](http://www.scikit-yb.org/en/latest/api/features/importances.html) for reference.
image

- PoC for using matplotlib sphinx extension for embedding plotting code with plots
  - Removes code directives and replaces them with plot directives
  - Embeds plotting code with plots
  - Can remove<image_creation>.py scripts as they're redundant
- Allows code and visualization code to be centralized into rst files
- Add make.bat for windows dev


rwhitt2049:
Cool, looking forward to seeing this through! There are a quite a few plots, so I'll sit down tonight and make a checklist of sections to tackle so I can keep track of what I have and haven't completed.

And to answer your question, you can turn off plotting on a directive basis via the `:nofig:` option. This will show only the code. You can also suppress the source code for everything via `conf.py/plot_include_source=True\False`, or you control it at the directive level via the `:include-source:` option. This will show or hide the source code. So I think we can make it work.

I'll do my best to intuit the intention of the documentation and then document the areas where I'd like some direct feedback.

For reference, here's the plot directive's [documentation](https://matplotlib.org/devel/plot_directive.html).

bbengfort:
Awesome, this all sounds good! Once you get the checklist, would you mind creating an issue for it? Rather than making this one big, drawn-out PR, I would suggest that we do it in chunks; e.g. first the feature visualizers, then the regressor visualizers etc. (actually, we should probably start with a small chunk first). This will also help ensure new contributors use the `plot::` directive rather than create additional technical debt.

rwhitt2049:
@bbengfort I've got a checklist made. Should I make one issue with a checklist of each section to complete? Or do you think I should make a separate issue for each section? The former would be less clutter on your issues list, but would probably promote a really big PR. The latter would make PR's smaller and make it possible to create some more novice issues. 

Thoughts? I don't have a preference.

bbengfort:
We don't necessarily have a one-to-one relationship between issues and PRs; I'm hoping that with one big issue, broken into sections, marked as novice, we can have multiple associated PRs, only closing the issue when all of the documentation is updated. We can use the comments to assign each section to individual contributors Is that ok?

rebeccabilbro:
@bbengfort - per our conversation during the new YB release earlier this evening, let's discuss what we need to do to get this extension ready, e.g. how we'll deal with the images that take a long time to `poof`, and how to address the datasets robustness.

rwhitt2049:
Circling back to this 6 months later (sorry, new baby, new house, new job).

@bbengfort , @rebeccabilbro ... Is there still a need here? I can dedicate around 16 hours to this in December if there is.

ndanielsen:
@rwhitt2049 thanks for getting back to this! We're definitely still in need of this.

Any time you can dedicate to this would be greatly appreciated. Also congrats on all of the new events and blessings in life!

bbengfort:
@rwhitt2049 @DistrictDataLabs/team-oz-maintainers - with the merge of #535 and our new datasets mechanism, I think we should more actively pursue this change in the documentation for version 1.0. 

@rwhitt2049 I think potentially we could merge this PR with the new plot directive, and then start adapting documentation as needed; do you feel the same way? Do you happen to have that checklist around still? 

@lwgray Should updating the documentation to use the plot directive be part of the checklist?

@Kautumn06 any thoughts about the code show/repetitive code thing from above?

lwgray:
@bbengfort I have added a column for "docs uses plot directive" in the survey.

Kautumn06:
This is a great idea @rwhitt2049 and thank you for writing up such a detailed and thoughtful PR! 

My only concern is about not being able to split up the code blocks, since that will make several of our documentation pages, such as [Rank Features](http://www.scikit-yb.org/en/latest/api/features/rankd.html), even longer. It may not be an issue for every visualizer, but I think it may become more noticeable as we add additional customization options to the documentation. A recent example is the `KElbowVisualizer`, where we've expanded the documentation by showing how to hide the `timings` parameter and use the optional scoring metrics, such as `silhouette` and `calinski_harabaz`. However, this is just my perspective and I'd be happy to hear everyone else's thoughts!

rwhitt2049:
@bbengfort I added the checklist as issue [687](https://github.com/DistrictDataLabs/yellowbrick/issues/687). It looks like I only made it through the Features, so there may be more to add. I just started at your index.rst and when down the list adding any plot directive candidates. I believe that you should be able to merge this PR as is, but you'll have to decide how to handle the code split.

@Kautumn06 Yes, I agree that this isn't ideal. You can suppress code in the plot directive via the `include-source` option. So you could replace the png import and hide the code in the plot directive, and then keep your code directives as is. You'll have have two places to maintain the same code, but as I see it, that's your current situation anyways. This way, you at least gain the programmatic creation of your plots.

rebeccabilbro:
@rwhitt2049 Thank you very much for adding the checklist in #687! I know @bbengfort was diving into this late last week and is enthusiastic about moving forward -- he's traveling the next few days, in and out of reliable wifi, but back at the end of this week! 
In the meantime, @wagner2010, @Kautumn06 & @lwgray -- would you three mind building the version of the docs from @rwhitt2049's PR so that we can get a few oz maintainers to weigh in on the impacts of the change?

docs/api/features/importances.rst
```diff
@@ -17,28 +17,32 @@ Random Forest, Gradient Boosting, and Ada Boost provide a
...
+optional, if an ``Axes`` isn't specified, Yellowbrick will use the current
+figure or create one). We can then fit a ``FeatureImportances`` visualizer
+with a ``GradientBoostingClassifier`` to visualize the ranked features.
+
+.. plot::
 
     # Load the classification data set
-    data = load_data("occupancy")
+    from yellowbrick.datasets import load_occupancy
+    occupancy = load_occupancy()
```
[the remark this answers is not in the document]
bbengfort: Will do, good catch!

Kautumn06:
Hi @rebeccabilbro! When building the docs, I found that the `KElbowVisualizer` page wasn't rendering the second image. I suspect this occurred because when @rwhitt2049 first created this pull request the documentation page only had a single example. However, now that it contains two examples (and therefore, two images), I needed to add a `make_blobs` call in the second code block, which is the change I just pushed in the latest commit. Here is the updated code block:
elbow_updated

And here is how it was originally. Since the images were generated by the accompanying scripts in each folder, the second `make_blobs` call wasn't necessary to create it in the docs.
original_elbow

I hope this is helpful! And @DistrictDataLabs/team-oz-maintainers just let me know if you have any questions or if there is something I missed!

bbengfort:
@Kautumn06 @lwgray I appreciate your concerns about separating the code blocks from the plot blocks. I think that in the common case the `.. plot::` will serve us well, but in the case where we do want the code and the images to be kept separate we will simply have multiple code blocks in the documentation and not include the source of `.. plot::`.

If you feel this is too messy, we can still keep Python scripts in the documentation under a folder called `pyplots` and the extension can load them and insert the image automatically without including the source. See [this documentation](https://matplotlib.org/sampledoc/extensions.html#inserting-matplotlib-plots) for more detail.

bbengfort:
@DistrictDataLabs/team-oz-maintainers here is where we are with autodoc. I've updated #446 for matplotlib 3.0.2 directives and converted `api/features/importances.rst` and `api/cluster/elbow.rst` to use the new plot directives. 

Generally the steps are as follows:

1. Replace `.. code::` blocks with `.. plot::` blocks in doc.rst
2. Remove all `.. image::` references in doc.rst
3. Delete doc.py
4. Delete old images in images/

Some questions:

1. How do we make the size of the generated images bigger?
2. What about the images required for the gallery? 

I'm not sure why CI is failing, I'll take a look at that and as soon as it's fixed we can add the above questions to #687 and merge this PR in.

docs/conf.py
```diff
@@ -126,11 +137,61 @@
...
+# See: https://github.com/phn/pytpm/issues/3#issuecomment-12133978
+numpydoc_show_class_members = False
+
+# Locations of objects.inv files for intersphinx extension that auto-links
+# to external api docs.
+intersphinx_mapping = {
+    'python': ('https://docs.python.org/3', None),
+    'matplotlib': ('http://matplotlib.org/', None),
+    'scipy': ('http://docs.scipy.org/doc/scipy/reference', None),
+    'numpy': ('https://docs.scipy.org/doc/numpy/', None),
```
bbengfort: This fixed the error we were getting during build.

docs/conf.py
```diff
@@ -126,11 +137,61 @@
...
+
+# Whether to show a link to the source in HTML.
+plot_html_show_source_link = True
+
+# Code that should be executed before each plot.
+plot_pre_code = (
+    "import numpy as np\n"
+    "import matplotlib.pyplot as plt\n"
+    "from yellowbrick.datasets import *\n"
+)
```
bbengfort: Note that this directive ensures that `plt`, `np`, and all of the dataset loaders are available during the `.. plot::` directive, so they don't have to be specifically imported. This might also help with the code splits thing, which were generally about loading datasets.

docs/conf.py
```diff
@@ -126,11 +137,61 @@
...
+
+# Whether to show links to the files in HTML.
+plot_html_show_formats = True
 
-# -- Options for HTML output ----------------------------------------------
+# A dictionary containing any non-standard rcParams that should be applied before each plot.
+plot_rcparams = {
+    "figure.figsize": (9,6),
+    "figure.dpi": 128,
+}
```
bbengfort: This solved the figure size problem in the docs.

docs/conf.py
```diff
@@ -1,41 +1,74 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
+# conf
+# Yellowbrick documentation build config file, created by sphinx-quickstart
```
bbengfort: I reorganized the configuration to make it a bit easier to manage. Note though that our configuration file is behind the current version so at some point we may want to do a more thorough review.

docs/conf.py
```diff
@@ -1,41 +1,74 @@
...
 
 # If your documentation needs a minimal Sphinx version, state it here.
-#
-# needs_sphinx = '1.0'
+# needs_sphinx = '1.8'
+
+# General information about the project.
+project = 'Yellowbrick'
+copyright = '2016-2019, The scikit-yb developers.'
+author = 'The scikit-yb developers'
```
bbengfort: First change in copyright and organization identification.

bbengfort:
@DistrictDataLabs/team-oz-maintainers this is ready for review, I'll fix the tests in another PR and merge them in when they're ready.

bbengfort:
Sorry I realize the purpose of this PR wasn't very clear, so here is what I envision the final commit after squash and merge to be:

---

Adds plot directives to generate images on build.

This PR adds the `matplotlib.sphixext.plot_directive` to our documentation configuration, allowing us to use plot directives that automatically generate related images with documented code snippets when the documentation is built. Additionally, the docs/api/features/importances.rst and docs/api/cluster/elbow.rst documentation was updated to include the new directive as an example.

This just a start toward converting our documentation to use this new directive, but there are still questions to answer. For example, how do we auto-generate the gallery images, r how do we use narrative code blocks along with the plot directive? For now, we will address these in a case-by-case basis as we convert the documentation described in #687. 

Closes #353 #354

docs/conf.py
lwgray: right on!

docs/conf.py
bbengfort: good question - this is just for the documentation and I set the figure width to be about the same size as the body of the web page. I think that these are just defaults and can be configured in the `.. plot::` directive, e.g. with 

```python
_, ax = plt.subplots(figsize=(12,4))
```

But we should test that out!

docs/api/features/importances.rst
```diff
@@ -17,44 +17,27 @@ Random Forest, Gradient Boosting, and Ada Boost provide a
...
+    from yellowbrick.features import FeatureImportances
 
-    # Create a new matplotlib figure
-    fig = plt.figure()
-    ax = fig.add_subplot()
+    # Load the classification data set
+    X, y = load_occupancy()
 
-    viz = FeatureImportances(GradientBoostingClassifier(), ax=ax)
+    viz = FeatureImportances(RandomForestClassifier())
```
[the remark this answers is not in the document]
bbengfort: I was trying to get a better feature importances plot - gradient boosting was just one bar; Random Forest didn't help much though ...

docs/conf.py
rwhitt2049: It looks like you can also specify plot sizes in the directive as options.

From docs:

"Additionally, this directive supports all of the options of the image directive, except for target (since plot will add its own target). These include alt, height, width, scale, align and class."

bbengfort: That will be very helpful for the gallery plots!

PR #413, opened 2018-05-14
Classification Report (summary) #359

ariley1472:
Fixes #359 

Initial Commit

Changes proposed in this pull request:

I want to make a pull request prior to making any significant changes. 
I'm trying to add another column to the Classification Report that shows the support. It looks like I just need to add the 'support' key value to the SCORE_KEYS and the functions within classification_report.py (including score, draw, and finalize). I still need to implement the actual drawing of the figure.

@DistrictDataLabs/team-oz-maintainers

PR #463, opened 2018-06-02
feat: If applied, this commit will make 'support' optional for classification report

lwgray:
This change is being made because the user had no control over when and how to
display support on the classification report visualizer.

It's being made to allow the user to select if the support will be showcased
as a raw count or a percentage.

This change is also being made because we want the new feature to be backward-compatible

Expands PR #413 
Examples when support is set to [None, 'percent', 'count']

image

image

image


tests/test_classifier/test_classification_report.py
```diff
@@ -178,3 +170,46 @@ def test_isclassifier(self):
 
         with pytest.raises(yb.exceptions.YellowbrickError, match=message):
             ClassificationReport(LassoCV())
+
+    def test_support_count_class_report(self):
```
bbengfort: Thank you for adding both of these tests - very nice work! I reviewed the baseline images as well, and they're also solid!

yellowbrick/classifier/classification_report.py
```diff
@@ -79,14 +83,17 @@ class ClassificationReport(ClassificationScoreVisualizer):
...
+                 support=None, **kwargs):
         super(ClassificationReport, self).__init__(
             model, ax=ax, classes=classes, **kwargs
         )
 
         self.cmap = color_sequence(cmap)
         self.cmap.set_under(color=CMAP_UNDERCOLOR)
         self.cmap.set_over(color=CMAP_OVERCOLOR)
+        self.support = support
+        self.measurements = ['precision', 'recall', 'f1', 'support']
```
bbengfort: Because we have a `SCORE_KEYS` constant, I recommend we change this to:

```python
self._displayed_scores = [key for key in SCORE_KEYS] 
if not support: 
    self._displayed_scores.remove("support") 
```

We can then use this attribute rather than the constant in the `score` method. This will also give us flexibility in the future to change what is displayed and what isn't.

yellowbrick/classifier/classification_report.py
```diff
@@ -79,14 +83,17 @@ class ClassificationReport(ClassificationScoreVisualizer):
...
+    def __init__(self, model, ax=None, classes=None, cmap='YlOrRd',
+                 support=None, **kwargs):
         super(ClassificationReport, self).__init__(
             model, ax=ax, classes=classes, **kwargs
         )
 
         self.cmap = color_sequence(cmap)
         self.cmap.set_under(color=CMAP_UNDERCOLOR)
         self.cmap.set_over(color=CMAP_OVERCOLOR)
+        self.support = support
```
bbengfort: I would recommend some validation of the support argument at this point, in case a user passes `support="foo"` or something like that. It can be as simple as:

```python
if support not in {None, True, False, "percent", "count"}:
    raise YellowbrickValueError()  
```

This would be a good candidate for a test case as well. 

```python
def test_invalid_support(self):
    """
    Ensure that bad support arguments raise exception 
    """
    with pytest.raises(YellowbrickValueError, match="bad support argument"):
        ClassificationReport(model, support="foo") 
```

yellowbrick/classifier/classification_report.py
```diff
@@ -106,14 +113,21 @@ def score(self, X, y=None, **kwargs):
 
         # Calculate the percentage for the support metric
         self.support_score = scores[-1]
-        support_percent = self.support_score / (sum(self.support_score))
+        self.support_percent = self.support_score / (sum(self.support_score))
```
bbengfort: I do very much like the idea of storing the support scores on the visualizer for access, even if they're not being displayed in the final visualization. However, I don't see any reason why we have to store it in a property that is separate from `scores_`. I propose the following, which might make things a bit simpler. 

```python 
if self.support == PERCENT:
    # scores is a tuple, convert to a list to modify the support score to a percentage
    # note that the support metric is the last item in the scores list 
    scores = list(scores) 
    scores[-1] = scores[-1] / scores[-1].sum()
  
# Create a mapping of class to score for each of the metrics in the scores list 
# Then map the score type (f1, precision, recall, and support) to the dict of class to score. 
scores = map(lambda s: dict(zip(self.classes_, s)), scores) 
self.scores_ = dict(zip(SCORE_KEYS, scores))
```

A copule of points here:

- We should have string constants for `PERCENT` and `COUNT` rather than using `"percent"` and `"count"`. 
- This method uses the np function of sum in `scores[-1].sum()` 
- This version ensures that `self.scores_` contains _all_ score metrics, we'll do the display filtering in draw. 
- Note that this code replaces lines 114-130

yellowbrick/classifier/classification_report.py
```diff
@@ -122,17 +136,17 @@ def draw(self):
         Renders the classification report across each axis.
         """
         # Create display grid
-        cr_display = np.zeros((len(self.classes_), 4))
+        cr_display = np.zeros((len(self.classes_), len(self.measurements)))
```
bbengfort: Awesome, thank you! This definitely will make things a bit more dynamic if we want to allow users to display different metrics.

yellowbrick/classifier/classification_report.py
```diff
@@ -122,17 +136,17 @@ def draw(self):
...
 
 
         # For each class row, append columns for precision, recall, f1, and support
         for idx, cls in enumerate(self.classes_):
-            for jdx, metric in enumerate(('precision', 'recall', 'f1', 'support')):
+            for jdx, metric in enumerate(self.measurements):
```
bbengfort: No change required, but by having `_displayed_scores` pre-filtered, this line of code will ensure `support` is displayed or not.

yellowbrick/classifier/classification_report.py
bbengfort: NOTE: the above is not a complete solution; just a suggestion for keeping all the scores in the same place; we still have to figure out how/where to compute the percent if `not self.support == PERCENT`.

yellowbrick/classifier/classification_report.py
```diff
@@ -122,17 +136,17 @@ def draw(self):
...
                 cr_display[idx, jdx] = self.scores_[metric][cls]
```
bbengfort: Slight bug injected here from my above code proposals - for `support`, this value must always be the percentage,  and not the count if that's the values stored, this was guaranteed before since support was always a percentage in the `scores_` dictionaries. So depending on how you proceed with this, I just wanted to highlight this minor issue.

yellowbrick/classifier/classification_report.py
lwgray: I knew you were going to comment on this; I wasn't sure what to do about it.  Thanks !

yellowbrick/classifier/classification_report.py
lwgray: A big question for me...Where do I store the support counts?  Since, percent needs to be store in ```scores_```

yellowbrick/classifier/classification_report.py
lwgray: Completed! 👍

yellowbrick/classifier/classification_report.py
lwgray: @bbengfort 
Still working through this... 😢 
Here's what I understand:
1. The support percentage is needed for all support conditions [True, 'count', 'percent']
2. The only difference between these support conditions [True, 'count', 'percent']  is when you need to draw the heatmap
3.  To draw [True or 'count'], you need a reference to the original support count
4.  To draw 'percent' nothing is needed
My question is how do you store the original support count?

lwgray:
@bbengfort By the way... Thanks for the thorough review 😺

yellowbrick/classifier/classification_report.py
bbengfort: I've been thinking about this a bit more, what if we do this:

- `scores_` contains whatever the _display text_ is going to be (either count or percentage) - because this is what users will look at and we want it to be consistent with what they see. 
- Create a property called `_support_percent` (as you did originally) that will hold the percentage for the color display. 

I realize that this is close to what you we're doing before, though exactly opposite to the original method. The alternative though is to have two duplicated data structures, e.g. `scores_` and `scores_intensity_` -- which would be identical except for the case where support is a count. 

What do you think?

```python 
# note that the support metric is the last item in the scores list 
self._support_percent  = scores[-1] / scores[-1].sum()

if self.support == PERCENT:
    # scores is a tuple, convert to a list to modify the support score to a percentage
    scores = list(scores) 
    scores[-1] = self._support_percent
  
# Create a mapping of class to score for each of the metrics in the scores list 
# Then map the score type (f1, precision, recall, and support) to the dict of class to score. 
scores = map(lambda s: dict(zip(self.classes_, s)), scores) 
self.scores_ = dict(zip(SCORE_KEYS, scores))
```

bbengfort:
@lwgray of course! thank you so much for working on this!

yellowbrick/classifier/classification_report.py
lwgray: We might need to go back and forth on this for a bit.

I thought ```scores_``` contains percentages for the heatmap while the draw method contains the text labels.

bbengfort: As it's currently implemented that's true - but this never became an issue until we added `support` -- before that `scores_` was both the label and the heatmap intensity.

tests/test_classifier/test_classification_report.py
```diff
@@ -178,3 +170,53 @@ def test_isclassifier(self):
...
+
+        assert viz.scores_ == {
+            'precision': {0: approx(0.7446808), 1: approx(0.8490566)},
+            'recall': {0: approx(0.8139534), 1: approx(0.7894736)},
+            'f1': {0: approx(0.7777777), 1: approx(0.8181818)},
+            'support': {0: approx(0.42999999999999999),
+                        1: approx(0.56999999999999995)}
+            }
+
+    def test_invalid_support(self):
```
bbengfort: Thanks for adding this!

yellowbrick/classifier/classification_report.py
```diff
@@ -79,14 +85,23 @@ class ClassificationReport(ClassificationScoreVisualizer):
...
         )
 
         self.cmap = color_sequence(cmap)
         self.cmap.set_under(color=CMAP_UNDERCOLOR)
         self.cmap.set_over(color=CMAP_OVERCOLOR)
+        self._displayed_scores = [key for key in SCORES_KEYS]
+        self.support = support
+
+        if support not in {None, True, False, "percent", "count"}:
+            raise YellowbrickValueError("bad support argument")
```
bbengfort: not a big deal, but I would prefer something a bit more informative:

```python
raise YellowbrickValueError(
    "'{}' is an invalid argument for support, use None, True, False, 'percent', or 'count'".format(support)
)
```

yellowbrick/classifier/classification_report.py
```diff
@@ -99,21 +114,26 @@ def score(self, X, y=None, **kwargs):
...
 
+        # Create a mapping composed of precision,recall, presion, and support
+        # to their respective values
         scores = map(lambda s: dict(zip(self.classes_, s)), scores)
         self.scores_ = dict(zip(SCORES_KEYS, scores))
 
-        # Change the support score from the actual support value to the percent
-        # value to be used in the Classification Report.
-        self.scores_['support'] = dict(zip(self.classes_, support_percent))
+        # Remove support scores
```
bbengfort: Maybe `# Remove support scores if not required`?

yellowbrick/classifier/classification_report.py
```diff
@@ -99,21 +114,26 @@ def score(self, X, y=None, **kwargs):
...
         scores = precision_recall_fscore_support(y, y_pred)
 
         # Calculate the percentage for the support metric
-        self.support_score = scores[-1]
-        support_percent = self.support_score / (sum(self.support_score))
+        # and store the percent in place of raw support counts
+        self._support_score = scores[-1]
+
+        scores = list(scores)
+        scores[-1] = scores[-1] / scores[-1].sum()
```
bbengfort: Ok, so you've chosen to store the original support scores and use `scores_` to only store intensity values. I think that's perfectly reasonable. In this case, however, I would change the variable name to `self.support_score_` -- `self._support_score` indicates something we'd rather not let the user take a look at (it's internal, protected), whereas `self.support_score_` indicates a learned parameter that might be important to the user, particularly if they specify percentage mode.

lwgray:
@bbengfort the quick_method test only pass when I remove the random_state from DecisionTreeClassifier

bbengfort:
@lwgray you'll probably have to add the random state, then resync the baseline images. From the root of the repository:

```
$ pytest tests/test_classifier/test_classification_report.py 
$ python -m tests.images tests/test_classifier/test_classification_report.py 
```

The first running of the test will regenerate the actual image, then the `tests.images` module will copy it to the correct directory; when you run the tests again they should pass.

Issue #406, opened 2018-05-14
docs quick start walkthrough is missing r-squared explanation


leah-kstra:
In the quick start walkthrough, at the end of the ResidualsPlot explanation, one sentence starts "Because our coefficient of determination for this model is 0.328..." but there is no explanation of where this value comes from. 

### Proposal/Issue

It'd be helpful to include another section here which explains how to find the r-squared value before moving onto the next section. (For instance, walking the reader through the exercise creating the PredictionError graph at this point in the walkthrough would help, though it might still make the most sense to keep it at the end.)  Additionally, a useful sentence to add would explain to the reader why their r-squared value might not match the value in the walkthrough due to the randomization in the train/test split. 

### Background

http://www.scikit-yb.org/en/latest/quickstart.html#using-yellowbrick

mitevpi:
I think the scoring is happening in the earlier code block, but I agree that it can be clearer where the value is coming from, and why we're calculating it/making decisions based off of it. 

R-squared origin: https://github.com/DistrictDataLabs/yellowbrick/blob/develop/docs/quickstart.rst/L#176

I wouldn't mind trying to revise the section for a little more clarity - if anyone has other suggestions, please let me know.

bbengfort:
@mitevpi this might be as simple as adding a comment to the code block you identified, but feel free to take this one on!

PR #473, opened 2018-06-09
Added details to quickstart documentation on missing r-squared value

mitevpi:
This PR is in response to #406

Added some more details and explanation on where the value was coming from based on which we are determining which model to use. Linked to the source sklearn .score() documentation rather than diving into details on the quickstart page.


docs/quickstart.rst
```diff
@@ -178,13 +178,13 @@ At this point, we can train our model; let's fit a linear regression to our mode
...
 
-Because our coefficient of determination for this model is 0.328, let's see if we can fit a better model using *regularization*, and explore another visualizer at the same time.
+Along with generating the residuals plot, we also measured the peformance, or "scored" our model on the test data above: ``visualizer.score(X_test, y_test)``. Because we used a Linear Regression model, the `scoring consists of finding the R-squared value of the data <http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html#sklearn.linear_model.LinearRegression.score>`_, which is a statistical measure of how close the data are to the fitted regression line. The R-squared value of any model may vary slightly between prediction/test runs, however it should generally be comparable. In our case, the R-squared value for this model was only 0.328, suggestion that linear correlation may not be the most appropriate to use for fitting this data. Let's see if we can fit a better model using *regularization*, and explore another visualizer at the same time.
```
bbengfort: minor typos:

peformance --> performance 
Linear Regression --> ``LinearRegression`` or "linear regression". 
was only 0.328, suggestion --> suggesting 

also, how about "we also measured the performance by "scoring" our model on the test data, e.g. the code snippet ``visualizer.score(X_test, y_test)``."?

mitevpi: Thanks for the review! I'll make the changes and re-commit.

PR #477, opened 2018-06-14
Fixes #433 - Correct spelling of "Principal" in PCA rst

### LINKED ISSUE #433: Doc error: on Docs » Visualizers and API » Feature Analysis Visualizers » PCA Projection "principle component analysis" should be "principal component analysis"
In http://www.scikit-yb.org/en/latest/api/features/pca.html
the phrase "principle component analysis" should be "principal component analysis"

lwgray:
In the rst file of principal component analysis, which is located
in docs/api/features/pca.rst, "principal" was misspelled as "principle"

This was corrected because we want to keep our documentation in
tip-top shape.  Our documentation sets us apart from other projects

### DIFF
--- a/docs/api/features/pca.rst
+++ b/docs/api/features/pca.rst
@@ -3,7 +3,7 @@
 PCA Projection
 ==============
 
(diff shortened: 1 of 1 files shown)


bbengfort:
👍 commit!

PR #480, opened 2018-06-15
Implements histogram alongside ResidualsPlot

### LINKED ISSUE #264: [Feature Suggestion] Add histogram plot of residual errors to ResidualsPlot
The current `ResidualsPlot` shows training and testing residuals as a scatter plot, by eye we can get an idea of whether more errors are above or below the 0 line. By adding a histogram of testing errors we might more clearly be able to tell if errors have a Normal  distribution.

In the following examples I have some large positive and negative errors, from the histogram it looks as though I have a negatively skewed distribution which might tell me something about my training examples:
image

```
from yellowbrick.regressor import ResidualsPlot
fig, ax = plt.subplots(figsize=(8,6)); 
model = ResidualsPlot(clone_estimator(clf), ax=ax)
model.fit(X_train, y_train)
model.score(X_test, y_test)

# add histogram of residual errors
left, bottom, width, height = [0.65, 0.17, 0.2, 0.2]
ax2 = fig.add_axes([left, bottom, width, height])

testing_residuals = pd.Series(model.predict(X_test) - y_test)
testing_residuals.plot(kind="hist", bins=50, title="Residuals on Predicted", ax=ax2);
ax2.vlines(0, ymin=0, ymax=ax2.get_ylim()[1] ) # add x==0 line

model.poof()
```

It isn't obvious where the best location would be for the histogram. Annoyingly I cannot get an `alpha` value for `ax2` either (I'd hoped to make this semi-transparent so location was less of an issue).

bbengfort:
Adds a hist=True option to the ResidualsPlot visualizer, which plots the
histogram of the residuals as a barh on a vertical axes with sharey=True
to the right of the scatter plot axes. This allows the user an easier
method to identify the distribution of the errors (e.g. more positive or
more negative).

Note that matplotlib 2.0.2 or greater is required for this method.

Still to do:

- [x] update documentation
- [x] write matplotlib version tests

Fixes #264


bbengfort:
Example of the new visualizer:

new_residuals

yellowbrick/regressor/residuals.py
```diff
@@ -479,42 +533,96 @@ def finalize(self, **kwargs):
...
+    splits determined by test_size) then plots the training and test residuals
+    agains the predicted value for the given model.
 
     This helper function is a quick wrapper to utilize the ResidualsPlot
     ScoreVisualizer for one-off analysis.
 
     Parameters
     ----------
+    model : a Scikit-Learn regressor
+        Should be an instance of a regressor, otherwise a will raise a
```
lwgray: Line is not grammatically correct - "otherwise a will raise a" 
Same mistake also found at lines 62 & 330

yellowbrick/regressor/residuals.py
```diff
@@ -479,42 +533,96 @@ def finalize(self, **kwargs):
...
+    train_color : color, default: 'b'
+        Residuals for training data are ploted with this color but also
+        given an opacity of 0.5 to ensure that the test data residuals
+        are more visible. Can be any matplotlib color.
+
+    test_color : color, default: 'g'
+        Residuals for test data are plotted with this color. In order to
+        create generalizable models, reserved test data residuals are of
+        the most analytical interest, so these points are highlighted by
+        hvaing full opacity. Can be any matplotlib color.
```
lwgray: Mispelling of 'hvaing' - should be having
mistake also made on line 350

PR #483, opened 2018-06-18
PCADecomposition test enhancements

bbengfort:
Added image similarity tests to all facets of PCADecomposition testing
as well as image similarity tests for biplot PCA visualizations.


bbengfort:
Here are the two new biplot test images:

test_biplot_2d

test_biplot_3d

PR #484, opened 2018-06-20
Fixes #257 - Adds datasaurus dataset to show importance of visualizing data

### LINKED ISSUE #257: Add Datasaurus, a version of Anscombe's Quartet
As an addition to Anscombe's Quartet, but starting from a plot that looks like a dinosaur.

https://www.autodeskresearch.com/publications/samestats 
http://www.thefunctionalart.com/2016/08/download-datasaurus-never-trust-summary.html
data: https://www.autodeskresearch.com/sites/default/files/The%20Datasaurus%20Dozen.zip

lwgray:
This dataset is added for fun.  The whole point is to demonstrate that
simple descriptive statistics can't replace looking at the data

The dataset was originally comprised of 13 images but I felt showcasing only 4 would suffice and get our point across.

image

Feedback needed:
- [ ] I have hardcoded the data points but not sure if this is the best approach.  Obviously this can be changed if needed.


yellowbrick/datasaurus.py
```diff
@@ -0,0 +1,283 @@
+# yellowbrick.datasaurus
+# Plots a Datasaurus Quartet as an illustration of the importance of visualization.
+#
+# Author:   Benjamin Bengfort <bbengfort@districtdatalabs.com>
```
bbengfort: Hehe, I'm not really an author of this file - I'm guessing just copy and paste?

PR #486, opened 2018-06-21
Fix #147 Implements a lexical dispersion plot

### LINKED ISSUE #147: DispersionPlot
Implement a `DispersionPlot` similar to that in [NLTK](http://www.nltk.org/_modules/nltk/draw/dispersion.html) that generates a lexical dispersion plot for a corpus. A lexical dispersion plot will plot occurrences of words in a text, creating a kind of mini-map that can allow a user to compare highlighting across documents.

lwgray:
Builds DispersionPlotVisualizer that shows the occurence of a word in a
corpus and the distance each word is from the beginning of the corpus.

Still to do!
- [x] Add Quick Method
- [ ] Tests
- [ ] Documentation
- [x] Better Docstrings

Below is an example.

```from nltk.corpus import gutenberg
words = ['Elinor', 'Marianne', 'Edward', 'Willoughby']
text = gutenberg.words('austen-sense.txt')
viz = DispersionPlotVisualizer(words, color='b')
viz.fit(text)
viz.poof()
```

image

### DIFF
--- a/yellowbrick/text/dispersion.py
+++ b/yellowbrick/text/dispersion.py
@@ -1,0 +1,119 @@
+# yellowbrick.text.dispersion
+# Implementations of lexical dispersions for text visualization.
+#
+# Author:   Larry Gray
+# Created:  2018-06-21 10:06
+#
+# Copyright (C) 2018 District Data Labs
+# For license information, see LICENSE.txt
+#
+# ID: dispersion.py [67b2740] lwgray@gmail.com $
+
+"""
+Implementation of lexical dispersion for text visualization
+"""
+
+
+##########################################################################
+## Imports
+##########################################################################
+
+import numpy as np
+from yellowbrick.text.base import TextVisualizer
+from yellowbrick.exceptions import YellowbrickValueError
+
+##########################################################################
+## Dispersion Plot Visualizer
+##########################################################################
+
+
+class DispersionPlotVisualizer(TextVisualizer):
+    """
+    DispersionPlotVisualizer allows for visualization of the lexical dispersion
+    of words in a corpus.  Lexical dispersion is a measure of a word's
+    homeogeneity across the parts of a corpus.  This plot notes the occurences
+    of a word and how many words from the beginning it appears.
+    
+    Parameters
+    ----------
+    words : list, default: None
+        The list of words to examine their dispersion
+    
+    ax : matplotlib axes, default: None
+        The axes to plot the figure on.
+    
+    color : list or tuple of colors
+        Specify color for bars
+        
+    kwargs : dict
+        Pass any additional keyword arguments to the super class.
+    
+    These parameters can be influenced later on in the visualization
+    process, but can and should be set as early as possible.
+    """
+    
+    def __init__(self, words, ax=None, color=None, ignore_case=False, **kwargs):
+        super(DispersionPlotVisualizer, self).__init__(ax=ax, **kwargs)
+        
+        self.color = color
+        self.words = words
+        self.ignore_case = ignore_case
+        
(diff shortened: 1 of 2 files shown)


yellowbrick/text/dispersion.py
```diff
@@ -0,0 +1,162 @@
...
+        if self.ignore_case:
+            words_to_comp = list(map(str.lower, self.words))
+            text_to_comp = list(map(str.lower, text))
+        else:
+            words_to_comp = self.words
+            text_to_comp = text
+
+        points = [(x,y) for x in range(len(text_to_comp))
+                        for y in range(len(words_to_comp))
+                        if text_to_comp[x] == words_to_comp[y]]
```
bbengfort: I do want to complement you on your use of `map` and `range` -- both of these functions return memory safe generators, which is awesome! Also the list comprehension to generate the points is absolutely stellar! In fact, my first comment was going to be to simply instantiate points as a np.array and job done! 

The issue is that in order for this exeedingly brilliant comprehension to work, you need to know the length of the words in advance, which causes you to `list(X)` on line 70, and if ignore case is on, to `list(map(str.lower, text))` on line 75. This is a bit scary because `X` could be potentially very large (especially in the case of text) or it could be a generator that might be exhausted without line 70 bringing everything into memory. 

I would propose the following code which implements only a single iteration on the data, and does so with a generator ensuring that only one element is in memory at a time. It requires a second function to create the generator via `yield` statements:

```python
    def _compute_dispersion(self, text):
        # Enumerate is a generator, doesn't matter if text is not 
        for x, word in enumerate(text):
            if self.ignore_case: 
                word = word.lower()

            # NOTE: this will find all indices if duplicate words are supplied 
            # In the case that word is not in target words, any empty list is returned 
            # and no data will be yielded 
            for y in (self.target_words_ == word).nonzero()[0]:
                yield (x, y)

    def fit(self, text):
        # Create an index (e.g. the y position) for the target words 
        self.target_words_ = np.array(self.words) 
        if self.ignore_case:
            self.target_words_ = np.array([w.lower() for w in self.target_words_]) 
        
        # Stack is used to create a 2D array from the generator 
        points = np.stack(self._compute_dispersion(text))
        self.draw(points)
        return self 
```

In the `draw` function, you can then access the first column, x, as `points[:,0]` and the second column, y as `points[:,1]`. 

I'm sad that this code causes your comprehension to be eliminated, but it would improve the performance of the visualizer by 2-3x. What do you think?

yellowbrick/text/dispersion.py
```diff
@@ -0,0 +1,162 @@
...
+        if points:
+            self.x_, self.y_ = list(zip(*points))
+        else:
+            self.x_ = self.y_ = ()
```
bbengfort: Is it necessary to store the offsets for each word on the visualizer? My first reaction is that the user probably wouldn't want to look up this information, but I could be persuaded.

yellowbrick/text/dispersion.py
```diff
@@ -0,0 +1,162 @@
...
+    of a word and how many words from the beginning it appears.
+    
+    Parameters
+    ----------
+    words : list
+        A list of words whose dispersion will be examined within a corpus 
+    
+    ax : matplotlib axes, default: None
+        The axes to plot the figure on.
+
```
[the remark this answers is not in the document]
bbengfort: (and in the quick method)

lwgray:
Thank you for the comments... I will do my best to answer each in a timely manner.  👍

yellowbrick/text/dispersion.py
lwgray: I think your solution beautifully simple

PR #489, opened 2018-06-23
Fix #487 Create Test and docs for DispersionPlot

### LINKED ISSUE #487: Complete Tests and Documentation for DispersionPlot Visualizer
To be complete, the new Dispersion Plot Visualizer requires:

- [ ] Tests
- [ ] Documentation

lwgray:
Fix dispersion plot
- [x] Reversed ylabels

Tests to be completed
- [x] integrated test
- [x] integrated test with ignore_case = False
- [x] integrated test showing corpus text can accept generator

Docs to be completed
- [x] example
- [x] main text under Text Modeling visualizers section
- [ ] docs/index.rst
- [ ] DESCRIPTION.rst
- [ ] README.md


tests/test_text/test_dispersion.py
```diff
@@ -0,0 +1,53 @@
...
+# Created:  2017-03-22 15:27
+#
+# Copyright (C) 2018
+# For license information, see LICENSE.txt
+#
+# ID: test_freqdist.py [bd9cbb9] rbilbro@districtdatalabs.com $
+
+"""
+Tests for the frequency distribution text visualization
+"""
```
bbengfort: Just wanted to note that the header needs to be updated for this file.

tests/test_text/test_dispersion.py
```diff
@@ -0,0 +1,53 @@
...
+from yellowbrick.text.dispersion import *
+from tests.dataset import DatasetMixin
+from tests.base import VisualTestCase
+
+
+##########################################################################
+## DispersionPlot Tests
+##########################################################################
+
+class FreqDistTests(VisualTestCase, DatasetMixin):
```
bbengfort: Should also be renamed `DispersionPlotTests`

tests/test_text/test_dispersion.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @pytest.mark.xfail(
+        sys.platform == 'win32', reason="images not close on windows"
+    )
```
bbengfort: I'd advise leaving this `xfail` off until we know for certain that it won't work on windows. Note if you comment or remove these lines, you'll also need to remove `import sys` to avoid a linter error.

tests/test_text/test_dispersion.py
```diff
@@ -0,0 +1,53 @@
...
+    def test_integrated_dispersionplot(self):
+        """
+        Assert no errors occur during DispersionPlot integration
+        """
+        corpus = self.load_data('hobbies')
+	
+        docs = ' '.join(corpus.data)
+        text = docs.split()
```
bbengfort: Great solution! If you're feeling frisky you could also get it in one step as:

```python
text = [word for doc in corpus.data for word in doc.split()]
```

We might also want to test a generator in a separate test (to ensure both lists and generators work with the visualizer):

```python
from itertools import chain 

text = chain(*map(str.split, corpus.data))
```

Note the `*` before the map; that expands the map into chain arguments. Chain and map are both generators in this case.

tests/test_text/test_dispersion.py
```diff
@@ -0,0 +1,53 @@
...
+
+        target_words = ['game', 'player', 'score', 'oil', 'man']
+
+        visualizer = DispersionPlot(target_words)
+        visualizer.fit(text)
+
+        visualizer.poof()
+        self.assert_images_similar(visualizer)
```
bbengfort: Very nice!

tests/test_text/test_dispersion.py
```diff
@@ -0,0 +1,53 @@
...
+	
+        docs = ' '.join(corpus.data)
+        text = docs.split()
+
+        target_words = ['game', 'player', 'score', 'oil', 'man']
+
+        visualizer = DispersionPlot(target_words)
+        visualizer.fit(text)
+
+        visualizer.poof()
```
bbengfort: Hmm ... so the travis tests are failing with images not close, RMSE=23 ... that's high. One possible fix is not to call `poof` - text is often a problem with tests because linux has different fonts than mac, etc.

tests/test_text/test_dispersion.py
```diff
@@ -0,0 +1,76 @@
...
+from yellowbrick.text.dispersion import *
+from tests.dataset import DatasetMixin
+from tests.base import VisualTestCase
+from itertools import chain
+
+##########################################################################
+## DispersionPlot Tests
+##########################################################################
+
+class DispersionPlotTests(VisualTestCase, DatasetMixin):
```
bbengfort: It looks like AppVeyor is having a problem with the hobbies corpus, which I'll add an issue about. In the meantime, could you add the following:

```python
@pytest.mark.xfail(sys.platform == "win32", reason="Issue #491")
class DispersionPlotTests(VisualTestCase, DatasetMixin):
    ...
```

Issue #485, opened 2018-06-21
FeatureImportances not working with LogisticRegression


zfernandes93:
**Describe the bug**

FeatureImportances visualizer does not work with LogisticRegression. This is because instead of outputting coef_ as a list of numbers, LogisticRegression outputs the coef_ as a list within a list

**To Reproduce**

%%matplotlibmatplot  inline

import pandas as pd
import yellowbrick as yb
import matplotlib.pyplot as plt 

from sklearn.linear_model import LogisticRegression, Lasso
df = pd.read_csv('data/occupancy/occupancy.csv')

features = [
    "temperature", "relative humidity", "light", "C02", "humidity"
]
classes = ["unoccupied", "occupied"]

X = df[features]
y = df['occupancy']

** You'll see that like most classifiers, Lasso outputs coef_ as list
la = Lasso()
la.fit(X,y)
print(la.coef_)

** LogisticRegression outputs a list inside of a list
lr = LogisticRegression()
lr.fit(X,y)
print(lr.coef_)

** If you want to see the Error 
visualizer = FeatureImportances(LogisticRegression(), relative = False)
visualizer.fit(X, y)
visualizer.poof()

**Traceback**
IndexError: index 14 is out of bounds for axis 0 with size 1

**Desktop (please complete the following information):**
 - OS:  macOS
 - Python Version 3.6
 - Yellowbrick Version 0.7

rebeccabilbro:
Good catch @zfernandes93! 

It looks like in this case scikit-learn's [`LogisticRegression`](http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html) `coef_` returns a multidimensional array, in anticipation that there will be multiple classes and decision function coefficients for each class. In the case of a binary classification, the shape of `coef_` is (1, n_features), for multiclass it will be (n_classes, n_features), but either way `FeatureImportances` will break because it's expecting a flat array. 

[...]

PR #490, opened 2018-06-24
Feature importances fix

rebeccabilbro:
This PR addresses the bug identified in #485 by adding a check inside the `FeatureImportances` `fit` method for coefficients that may take the form of multdimensional arrays (as in the case of `LogisticRegression`), and if so, takes the mean of the arrays to artificially flatten them to enable the `FeatureImportances` visualizer to successfully plot the importances as a horizontal bar chart.

_Note: This PR isn't quite done, since I'd also like to update the [tests](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_features/test_importances.py) to ensure the fix is working as expected._


yellowbrick/features/importances.py
```diff
@@ -122,6 +124,10 @@ def fit(self, X, y=None, **kwargs):
         # Get the feature importances from the model
         self.feature_importances_ = self._find_importances_param()
 
+        # Check if feature importances is a multidim array & if so flatten
+        if self.feature_importances_.ndim > 1:
+            self.feature_importances_ = np.mean(self.feature_importances_, axis=0)
```
bbengfort: Just to record the discussion from Slack in the issue: we're expecting `LogisticRegression` to have a `coefs_` property whose shape is [`(n_classes, n_features)`](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/linear_model/logistic.py#L1100) therefore by taking the column-wise mean (e.g. axis=0) we're getting the average coefficient of all classes for each feature and using that to plot importances.

For future us,  would you mind updating the comment as follows?

```python
# If feature importances is a multidim array, we're expecting a shape of 
# (n_classes, n_features) therefore we flatten by taking the average by 
# column to get shape (n_features,)  (see LogisticRegression) 
```

Additionally, in the future I could see a very interesting feature importance visualizer that uses this shape of array to create a stacked bar chart of feature importances for each class/feature combination. Would you mind opening an issue for that or would you like me to?

yellowbrick/features/importances.py
```diff
@@ -122,6 +129,16 @@ def fit(self, X, y=None, **kwargs):
         # Get the feature importances from the model
         self.feature_importances_ = self._find_importances_param()
 
+        # If feature importances is a multidim array, we're expecting a shape of
+        # (n_classes, n_features) therefore we flatten by taking the average by
+        # column to get shape (n_features,)  (see LogisticRegression)
+        if self.feature_importances_.ndim > 1:
+            self.feature_importances_ = np.mean(self.feature_importances_, axis=0)
+
+        # TODO - as an alternative to the above flattening approach, explore an
```
bbengfort: 👍 thanks for adding the TODO and the issue!

Issue #439, opened 2018-05-18
Fit time timer utility


wagner2010:
Add timer utility functionality to the library to adapt any learned `fit_time_` parameters:

```python
import time

from functools import wraps
from dateutil.relativedelta import relativedelta


def humanizedelta(*args, **kwargs):
    """
    Wrapper around dateutil.relativedelta (same construtor args) and returns
    a humanized string representing the delta in a meaningful way.
    """
    delta = relativedelta(*args, **kwargs)
    attrs = ('years', 'months', 'days', 'hours', 'minutes', 'seconds')
    parts = [
        '%d %s' % (getattr(delta, attr), getattr(delta, attr) > 1 and attr or attr[:-1])
        for attr in attrs if getattr(delta, attr)
    ]

    return " ".join(parts)


class Timer(object):
    """
    A context object timer. Usage:
        >>> with Timer() as timer:
        ...     do_something()
        >>> print timer.interval
    """

    def __init__(self, wall_clock=True):
        """
        If wall_clock is True then use time.time() to get the number of
        actually elapsed seconds. If wall_clock is False, use time.clock to
        get the process time instead.
        """
        self.wall_clock = wall_clock
        self.time = time.time if wall_clock else time.clock

    def __enter__(self):
        self.start    = self.time()
        return self

    def __exit__(self, type, value, tb):
        self.finish   = self.time()
        self.interval = self.finish - self.start

    def __str__(self):
        return humanizedelta(seconds=self.interval)


def timeit(func, wall_clock=True):
    """
    Times a function; returns a Timer along with the result.
    """
    @wraps(func)
    def timer_wrapper(*args, **kwargs):
        """
        Inner function that uses the Timer context object
        """
        with Timer(wall_clock) as timer:
            result = func(*args, **kwargs)

        return result, timer
    return timer_wrapper
```

ZJPoh:
If no one else is working on this, I'd be happy to start working on this. This is my first time contributing so I might have some questions.

rebeccabilbro:
Hey there @ZJPoh, that would be great! Let us @DistrictDataLabs/team-oz-maintainers  know if there's anything we can help with!

bbengfort:
Closed by #493

PR #493, opened 2018-07-06
Add Timer utility and update Manifold fit time

zjpoh:
Add feature per [#439](https://github.com/DistrictDataLabs/yellowbrick/issues/439)

From initial exploration, the only `fit_time_` that I found is in `yellowbrick.features.manifold`.

A couple questions.
1. Should I be adding the Timer class to `yellowbrick.utils.helpers`?
1. Is it better to calculate the fit time using a context or by adding a decorator to `self.manifold.fit_transform`?

First time contributing~ Please let me know if this makes sense.


rebeccabilbro:
Great questions @ZJPoh - while I'm a bit fuzzy on the original vision for this feature (@wagner2010, @bbengfort feel free to jump in here to clarify!) as I'm thinking about this a bit more, it might make sense to break this into a few different parts, e.g.:

 - creating a [helper utility](https://github.com/DistrictDataLabs/yellowbrick/blob/d364facfdecaf5874b85bca3f7f0b2832e3483a2/yellowbrick/utils/helpers.py) that takes the `fit_time_` attribute to produce a human-readable string label that can be added to the plotted figure. As you said, currently the only visualizer that has `fit_time_` is Manifold, so this would also involve refactoring it's [`finalize`](https://github.com/DistrictDataLabs/yellowbrick/blob/d364facfdecaf5874b85bca3f7f0b2832e3483a2/yellowbrick/features/manifold.py#L381) to leverage the new utility.

 - creating a [decorator/timer_wrapper](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/utils/decorators.py) that could be used to time  the `fit` method of any YB visualizer and set its `fit_time_` (float) attribute, such that a user could do:

```python
slow_coord = ParallelCoordinates(fast=False)
slow_coord.fit(X, y)
print(slow_coord.fit_time_)

fast_coord = ParallelCoordinates(fast=True)
fast_coord.fit(X, y)
print(fast_coord.fit_time_)
```

@ndanielsen @lwgray @pdamodaran thoughts?

bbengfort:
@zjpoh thanks for taking this on - sorry this is a bit confusing because this is code that we had already written but just need to get included in the yellowbrick utilities! So to answer your questions:

- This code should be added to a `yellowbrick.utils.timer` module e.g. `timer.py` in `yellowbrick/utils`
- Both the decorator and the context manager would be valuable in this module, but for practical use, we envision mostly the context manager being used:

    ```python 
    def fit(self, X, y=None):
        with Timer() as self.fit_time_:
            # Do stuff 
    ```
- Note that the [KElbowVisualizer](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/cluster/elbow.py#L242) also will require the timer, though it's saving in a variable called `k_timers` (you can just create an issue to update this, though). 
- We'll probably need tests for the timer. 

Does that help answer your questions?

zjpoh:
@bbengfort Thanks for the comments. I'll 
* Add the module `yellowbrick.utils.timer`
* Use context manager for now
* Add tests to the module
* Create another issue on [KElbowVisualizer](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/cluster/elbow.py#L242) when I am done

bbengfort:
@zjpoh let us know when you're ready for a review.

zjpoh:
@bbengfort This is ready for review.
* I'm not exactly sure how to test timer. Please comment. Thanks.

ndanielsen:
@zjpoh thanks for contributing! Recommend that you take a look at the Travis results:

You are getting a linting test failure at : 

```
=================================== FAILURES ===================================
___________ pyflakes-check(ignoring ImportStarUsage ImportStarUsed) ____________
/home/travis/build/DistrictDataLabs/yellowbrick/tests/test_utils/test_timer.py:19: UnusedImport
'time' imported but unused
```

zjpoh:
@ndanielsen Thanks for pointing that out.

examples/examples.ipynb
```diff
@@ -288,9 +288,9 @@
...
     "\n",
-    "RadViz is a multivariate data visualization algorithm that plots each feature dimension uniformely around the circumference of a circle then plots points on the interior of the circle such that the point normalizes its values on the axes from the center to each arc. This meachanism allows as many dimensions as will easily fit on a circle, greatly expanding the dimensionality of the visualization. \n",
+    "RadViz is a multivariate data visualization algorithm that plots each feature dimension uniformly around the circumference of a circle then plots points on the interior of the circle such that the point normalizes its values on the axes from the center to each arc. This mechanism allows as many dimensions as will easily fit on a circle, greatly expanding the dimensionality of the visualization. \n",
     "\n",
-    "Data scientists use this method to dect separability between classes. E.g. is there an opportunity to learn from the feature set or is there just too much noise?"
+    "Data scientists use this method to detect separability between classes. E.g. is there an opportunity to learn from the feature set or is there just too much noise?"
```
bbengfort: Thank you for catching these spelling errors in the example notebook!

tests/test_utils/test_timer.py
```diff
@@ -0,0 +1,46 @@
...
+except ImportError:
+    import mock
+
+from yellowbrick.utils.timer import *
+
+##########################################################################
+## Helper Function Tests
+##########################################################################
+
+class TestTimer(unittest.TestCase):
```
bbengfort: We're trying to move over to [pytest](https://docs.pytest.org/en/latest/) and though we haven't fully deprecated our use of `unittest`, it would help us when we do, if we could have these be pytest tests by default. To do this, would you mind changing the definition of the class to `TimerTests(object)`? The tests below will then also require `assert` statements, which I'll point out below.

tests/test_utils/test_timer.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    Timer functions and utilities
+    """
+
+    @mock.patch('time.time', mock.MagicMock(return_value=1234.))
```
bbengfort: Patching time is an excellent method of testing the timer! If you change the mock to have a `side_effect` instead of a `return_value` that is a sequence, e.g. `mock.Mock(side_effect=[1234.2, 1242.8])` then the mock will return the values in order of calls. You can then use this to test the `Timer` difference of 8.6 seconds.

tests/test_utils/test_timer.py
```diff
@@ -0,0 +1,46 @@
...
+    def test_timer(self):
+        with Timer() as timer:
+            timer.time = mock.MagicMock(return_value=1236.)
+        self.assertIsInstance(timer.interval, float)
+        self.assertEqual(timer.interval, 2)
```
bbengfort: For pytest, these assertions will have to be updated to:

```python
assert isinstance(timer.interval, float) 
assert timer.interval == 2 
```

tests/test_utils/test_timer.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+    def test_human_readable_time(self):
+        self.assertEqual(human_readable_time(1.01), '00:00:01.0100')
+        self.assertEqual(human_readable_time(61.01), '00:01:01.0100')
+        self.assertEqual(human_readable_time(3661.01), '01:01:01.0100')
+        self.assertEqual(human_readable_time(360061.01), '100:01:01.0100')
```
bbengfort: This is not a problem at all, and is kind of a stylistic thing; you don't have to change it if you don't want to. One thing we've done in yellowbrick tests, is to use table driven tests, to make it easy to add new test cases if something comes up; e.g. 

```python
tests = [
    (1.01, '00:00:01.0100'), 
    (61.01, '00:01:01.0100'), 
    ... 
]

for case, expected in tests:
    assert human_readable_time(case) == expected 
```

bbengfort: Note that in these tests, you can also use something like `pytest.parameterize` -- but don't worry about that!

yellowbrick/features/manifold.py
```diff
@@ -314,9 +314,9 @@ def fit_transform(self, X, y=None):
             y = np.asarray(y)
             self.range_ = (y.min(), y.max())
 
-        start = time.time()
-        Xp = self.manifold.fit_transform(X)
-        self.fit_time_ = time.time() - start
+        with Timer() as timer:
+            Xp = self.manifold.fit_transform(X)
+        self.fit_time_ = timer.interval
```
bbengfort: This is great, thank you so much for updating this! One thing I was wondering, what would happen if we did this as follows:

```python
with Timer() as self.fit_time_:
    Xp = self.manifold.fit_transform(X) 
```

This would allow the user to `print(self.fit_time_)` to get the human readable interval, but we would probably have to update the title of the figure to ensure it worked correctly.

yellowbrick/utils/timer.py
```diff
@@ -0,0 +1,49 @@
...
+
+##########################################################################
+## Timer Class
+##########################################################################
+
+
+def human_readable_time(s):
+    h, s = divmod(s, 3600)
+    m, s = divmod(s, 60)
+    return "{:>02.0f}:{:02.0f}:{:>07.4f}".format(h, m, s)
```
bbengfort: This is a great function that doesn't require the python-dateutil library; thanks!

yellowbrick/utils/timer.py
```diff
@@ -0,0 +1,49 @@
...
+    def __enter__(self):
+        self.start = self.time()
+        return self
+
+    def __exit__(self, *exc):
+        self.finish = self.time()
+        self.interval = self.finish - self.start
+
+    def __str__(self):
+        return human_readable_time(self.interval)
```
bbengfort: Looks good!

tests/test_utils/test_timer.py
zjpoh: Sure, I will move over the pytest.

tests/test_utils/test_timer.py
zjpoh: Thanks for pointing that out!

tests/test_utils/test_timer.py
zjpoh: I will look into pytest.parametrize~

yellowbrick/features/manifold.py
zjpoh: `self.fit_time_` will then be an object instead of the string. 

We will have to change the documentation [here](https://github.com/DistrictDataLabs/yellowbrick/blob/43405974e25fa600b38aa9a7e52a6e86d0bd0e59/yellowbrick/features/manifold.py#L154) and the plotting part [here](https://github.com/DistrictDataLabs/yellowbrick/blob/43405974e25fa600b38aa9a7e52a6e86d0bd0e59/yellowbrick/features/manifold.py#L398).

Let me know if there are other changes that are needed.

tests/test_utils/test_timer.py
```diff
@@ -26,21 +26,24 @@
...
+        assert timer.interval == pytest.approx(8.6)
+
+
+@pytest.mark.parametrize('s,expected', [
+    (1.01, '00:00:01.0100'),
+    (61.01, '00:01:01.0100'),
+    (3661.01, '01:01:01.0100'),
+    (360061.01, '100:01:01.0100')
+])
+def test_human_readable_time(s, expected):
```
bbengfort: Awesome, thanks for making these changes!

yellowbrick/features/manifold.py
```diff
@@ -395,7 +394,7 @@ def finalize(self):
         """
         self.set_title(
             '{} Manifold (fit in {:0.2f} seconds)'.format(
-                self._name, self.fit_time_
+                self._name, self.fit_time_.interval
```
bbengfort: This is perfect, thank you!

Issue #502, opened 2018-07-16
Ensure ClassificationScoreVisualizers subclasses return score and add tests


bbengfort:
Extended from #361 to be more specific. In #407 we reworked the `ClassificationScoreVisualizer` to save the score in a `score_` property, then return it from the method after drawing. We still need to update the subclasses to either call `super()` or ensure that they implement the behavior described above. 

- [ ] `ClassPredictionError` should return an `self.estimator.score()` but already stores `score_`; requires test to ensure it returns a score between 0 and 1. 
- [ ] `ClassificationReport` should return `self.estimator.score()` but already stores `score_`; requires test to ensure it returns a score between 0 and 1. 
- [ ] `ConfusionMatrix` should return `self.estimator.score()` and store `score_` (it keeps its references in a `confusion_matrix_` property. Requires a test to ensure it returns a score between 0 and 1. 
- [ ] `ROCAUC` already returns a score property. Check to make sure it has a test for the score.

bbengfort:
Closed by #503

PR #407, opened 2018-05-14
WIP ClassificationScoreVisualizers should return accuracy

abatula:
Fixes #361 

Checking my understanding of the issue: ClassificationScoreVisualizer currently does not have an implemented `score()` method. It needs to be created, should return the score function, and should add the accuracy to the plotted figure if accuracy or F1 are not already plotted.

Changes proposed in this pull request:

-
-
-

@DistrictDataLabs/team-oz-maintainers

Issue #252, opened 2017-05-26
Finish up ROCAUC


bbengfort:
ROCAUC needs a few things to be finalized:

- [x] decision function bug (binary classifier) 
- [x] documentation 
- [x] add to gallery 
- [x] simplify example in docstring and switch to YB data instead of `sklearn.datasets`
- [x] add `multiclass` and `binary` fixtures from conftest.py (see [ClassificationReport](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/test_classification_report.py#L46)
- [x] remove [`X`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/test_rocauc.py#L44), [`yb`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/test_rocauc.py#L53), and [`ym`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/test_rocauc.py#L54) prefering to use above fixtures
- [x] remove [`load_binary_data`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/test_rocauc.py#L74) and [`load_multiclass_data`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/test_rocauc.py#L93) to use above fixtures 

Note the bug is as follows:

```
======================================================================
ERROR: Test ROCAUC with a binary classifier
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/benjamin/Repos/ddl/yellowbrick/tests/test_classifier/test_rocauc.py", line 110, in test_binary_rocauc
    s = visualizer.score(X_test, y_test)
  File "/Users/benjamin/Repos/ddl/yellowbrick/yellowbrick/classifier/rocauc.py", line 171, in score
    self.fpr[i], self.tpr[i], _ = roc_curve(y, y_pred[:,i], pos_label=c)
IndexError: too many indices for array
```

derekflint:
+ 1

rebeccabilbro:
@bbengfort - re: the decision function bug, I did a little research ([notebook here](https://github.com/rebeccabilbro/yellowbrick/blob/rocauc-refactor/examples/rebeccabilbro/rocauc_bug_research.ipynb)) into which classifiers and conditions actually seem to trigger the bug in the hopes that it might give us some ideas about how we should address it. 

[...]

PR #503, opened 2018-07-16
Follow-up on classification score visualizers

rebeccabilbro:
This PR is an attempt to finish up the work that @abatula started in [her recently merged PR](https://github.com/DistrictDataLabs/yellowbrick/pull/407), which now successfully ensures that `ClassificationScoreVisualizer` returns `self.score_` after the `score()` method is called! 

Here's what I've currently managed to do:

 - updated [`ROCAUC`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/rocauc.py#L201) `score` method to return `self.score_` instead of `self.estimator.score`.  _Note: ran into [this issue](https://github.com/DistrictDataLabs/yellowbrick/issues/252) when trying to add test coverage, so skipped that step for now; see below TODO_
 - updated [`ClassBalance`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/class_balance.py#L92) `score` method to call `self.draw()` and return `self.score_` and added test coverage
 - updated  [`ClassPredictionError`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/class_balance.py#L258) `score` method to call `self.draw()` and return `self.score_` and added test coverage
 - updated [`ConfusionMatrix`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/confusion_matrix.py#L192) `score` method to call `self.draw()` and return `self.score_` and added test coverage

Still to do:
- [x] update  [`ClassificationReport`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/classification_report.py#L141) `score` method to call `self.draw()` and return `self.score_` and add test coverage
- [x] figure out what to do about adding a `test_score_returns_score()` to the tests for `ROCAUC` (maybe just go ahead and add it  & mark it `@pytest.mark.skip` until we can come back to #252 ?) 

Question:
- there's now a good amount of duplication in the tests across the different classification score visualizers... possibly a signal that we should refactor; let me know if you think I should approach the   `test_score_returns_score()` tests differently


rebeccabilbro:
@bbengfort - oops! Just saw that you added [an issue for this](https://github.com/DistrictDataLabs/yellowbrick/issues/502) - lmk how you want to proceed

bbengfort:
@rebeccabilbro just reference that issue from this PR if this fixes it; I'll review in a second.

tests/test_classifier/test_rocauc.py
```diff
@@ -109,6 +109,11 @@ def test_binary_rocauc(self):
 
         # Score the visualizer
         s = visualizer.score(X_test, y_test)
+
+        # Test that score method successfully returns a value between 0 and 1
+        assert 0 <= s <= 1
```
bbengfort: A good place to put this test! Is there a similar place for multiclass? Would that be required?

tests/test_classifier/test_rocauc.py
```diff
@@ -109,6 +109,11 @@ def test_binary_rocauc(self):
...
+
+        # Test that the score method returns a value between 0 and 1
         self.assertAlmostEqual(s, 0.93230159261495249)
```
bbengfort: This probably can be removed in favor of the assertion you added.

tests/test_classifier/test_classification_report.py
```diff
@@ -221,3 +221,14 @@ def test_invalid_support(self):
...
             ClassificationReport(LinearSVC(), support="foo")
+
+    def test_score_returns_score(self):
+        """
+        Test that ClassificationReport score() returns a score between 0 and 1
+        """
+        viz = ClassificationReport(LinearSVC(random_state=42))
+        
+        viz.fit(self.binary.X.train, self.binary.y.train)
+        s = viz.score(self.binary.X.test, self.binary.y.test)
```
bbengfort: Nice work using the dataset fixture that the class was using!

tests/test_classifier/test_class_prediction_error.py
```diff
@@ -108,3 +110,26 @@ def test_class_type(self):
...
+        X = data[[
+            "temperature", "relative_humidity", "light", "C02", "humidity"
+        ]]
+
+        y = data['occupancy']
+
+        # Convert X to an ndarray
+        X = X.copy().view((float, len(X.dtype.names)))
+
+        X_train, X_test, y_train, y_test = tts(X, y, test_size=0.2, random_state=42)
```
bbengfort: For this test, it might be simpler to use the [`X` and `y` variables created on L38](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/test_class_prediction_error.py#L38) if you wanted - but this is perfectly fine. `ClassPredictionErrorTests` does need to have an integration test with a real data set + pandas added (though that's not for the PR, that's just a general comment).

yellowbrick/classifier/class_balance.py
```diff
@@ -223,10 +245,11 @@ def score(self, X, y, **kwargs):
         Returns
         -------
 
-        ax : the axis with the plotted figure
+        score_ : float
+            Global accuracy score
         """
 
-        # We're replying on predict to raise NotFitted
+        # We're relying on predict to raise NotFitted
```
bbengfort: Good catch!

yellowbrick/classifier/classification_report.py
```diff
@@ -129,16 +138,21 @@ def score(self, X, y=None, **kwargs):
         scores = list(scores)
         scores[-1] = scores[-1] / scores[-1].sum()
 
-        # Create a mapping composed of precision,recall, presion, and support
+        # Create a mapping composed of precision, recall, F1, and support
         # to their respective values
         scores = map(lambda s: dict(zip(self.classes_, s)), scores)
-        self.scores_ = dict(zip(SCORES_KEYS, scores))
+        self.scores_map_ = dict(zip(SCORES_KEYS, scores))
```
bbengfort: This is very minor, though I've probably given it more thought than I should have. I like the change in `ClassBalance` to store what used to be `scores_` into a `predictions_` table. Keeping the `score_` property more consistent as a single value metric throughout the `ClassificationScoreVisualizer` subclasses. 

However, here I would ask that we maintain this property as `self.scores_`. Although this might create a bit of confusion between `score_` and `scores_`, I believe it is important here to maintain backward compatibility with folks who were looking to fetch deeper metrics of this class (I know I have quite a bit of code that uses this attribute).

bbengfort:
Looks like I accidentally deleted my comment replying to your question about redundancy in the tests - you're right, there is a lot of code duplication, but I don't think you should approach it differently. In the future, we should make this one of the checks in `test.checks`, e.g. for things like `fit()` returns self, etc. But I think you did well with the tests - adding checks to tests that already evaluated score, and using the datasets already available on the class. 

Just for future, it may be easier to use `sklearn.datasets.make_classification` and the fixtures in `conftest.py` that are already split into train/test splits - just to reduce some of the boilerplate of setting up datasets; but like I said in the code review comments - some of the tests needed to have the "real world integration" and pandas tests added anyway.

tests/test_classifier/test_rocauc.py
rebeccabilbro: Done

tests/test_classifier/test_rocauc.py
rebeccabilbro: Done

tests/test_classifier/test_class_prediction_error.py
rebeccabilbro: Done

yellowbrick/classifier/classification_report.py
rebeccabilbro: Reverted

bbengfort:
Ready for me to merge when tests pass?

rebeccabilbro:
@bbengfort yes, thank you!

bbengfort:
Closes #502

Issue #508, opened 2018-07-19
Feature Correlation to Dependent Variable Visualizer


bbengfort:
**Describe the solution you'd like**

This issue extends #334 with details about bullet point 3: "plot feature/target correlations". 

As seen in [Model comparison using a noisy dataset -1](https://medium.com/@harsha.g1/model-comparison-using-a-noisy-dataset-1-db20f62c5126), it is useful to compare the pairwise correlation between the features and the dependent variable or target as a bar chart; similar to Rank1D and Rank2D, except that this is for the target only. 

![1_tfldsf_iujqgandsy6b39a](https://user-images.githubusercontent.com/745966/42943280-49a0c4c4-8b30-11e8-9331-d8e1d9f4d6c7.png)

Once the target package has been created, create a visualizer, `yellowbrick.target.FeatureCorrelation` that creates this bar chart by fitting `X` and `y`. Write documentation that links this visualizer to the `JointPlot` visualizer. 

Include the following correlations:

- Pearson 
- [Mutual Information](http://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.mutual_info_classif.html#sklearn.feature_selection.mutual_info_classif)

zjpoh:
@bbengfort If nobody is working on this, I would like to give it a try.

bbengfort:
@zjpoh go for it, thanks for taking this one!

PR #510, opened 2018-07-19
[MRG] Enhanced FeatureImportances for Multidimensional Coefficients

### LINKED ISSUE #492: Enhanced FeatureImportances for Multidimensional Coefficients
Create an enhanced version of the Feature Importances visualizer to allow the user to see the coefficients on a per-class basis.

Some estimators returns a multidimensional array for `coef_` , like scikit-learn's [`LogisticRegression`](http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html), in anticipation that there will be multiple classes and decision function coefficients for each class. In the case of a binary classification, the shape of `coef_` is (1, n_features), for multiclass it will be (n_classes, n_features). In #490, we addressed the bug raised in #485 by simply averaging the per-class feature importances for each feature in the case of multidimensional coefficients (as with LogisticRegression). However, we could create an enhanced version of this visualizer that would allow the user to see the coefficients on a per-class basis, depicted as a stacked histogram.

**Examples**
The `fit` method now checks to see if the `feature_importances_` are a multidimensional array. We could add a param such as `hist=True` that instead of averaging would stack the importances per class using something like [this](https://stackoverflow.com/questions/16653815/horizontal-stacked-bar-chart-in-matplotlib/16654564#16654564) in our `draw` method. If we go this route, we may want to consider adding a target labels attribute to the class à la the [`ClassificationScoreVisualizer`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/base.py#L67).

zjpoh:
Fixes #492 

Creates stack bar plot when there is multi-dimensional feature importance


zjpoh:
This is the feature importance plot from classifying the iris dataset with LogisticRegression.
ori_plot_iris

This a sample of how the stack bar plot that I initially created.
sample_new_plot

It this what we are planning to get?

rebeccabilbro:
Looking good so far @zjpoh! I think the main things we'd want to add next would be labels, titles, and a legend to assist users with interpreting their plots. Some of these things may be ones that you'd already thought of and hadn't gotten to yet, so apologies if you've already started tackling these!

- [ ] Add a legend similar to the one used in `ClassPredictionError`, where the colors correspond to the classes:

Class prediction error plot with legend

 - [ ] Ensure that the x-axis gets labeled properly, which the current `FeatureImportance` visualizer does with a [`_get_xlabel` method](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/features/importances.py#L229). 

Also:
 - [ ] [Label the bars](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/features/importances.py#L190)
 - [ ] [Set the title](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/features/importances.py#L201)
- [ ] [Remove the grid](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/features/importances.py#L207)

rebeccabilbro:
@zjpoh - one additional thought -- It might be worth it to poke around Scikit-Learn and see which (if any) other estimators might produce multidimensional coefficients. We originally identified this one via `LogisticRegression`, but different sklearn estimators sometimes have slight variations on the attribute names (e.g. ``coef_`` vs. ``feature_importances_``); it would be helpful to know these in advance so this visualizer can anticipate them!

zjpoh:
@rebeccabilbro I have added added a stacked parameter to the class. When `stack=True`, a stacked plot will be made.

A couple questions that I have are
1. I notice that when `relative=True`, the feature importance is scaled by the maximum value. What if the largest absolute value of the feature importance is a negative value?
rel_fea_imp
1. I scaled the feature importance by the maximum value of a specific class. Let me know if this is reasonable.
rel_fea_imp_stacked
1. Should we add another issue to add the option of selection top n features? This is particular useful when there are many features.
fea_imp_many

bbengfort:
@zjpoh this is looking great; thanks for the PR! To answer your questions:

1. Yes, it is a bug that we're not selecting the maximum absolute value to normalize by; you could change [Line 148](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/features/importances.py#L148) to:
    
    ```python
    maxv = np.abs(self.feature_importances_).max()
    ```

    I believe that should fix the problem, right?

2. Do you mean that you're normalizing each class to the largest value of the class, not the global maximum? It seems reasonable to me on the surface I guess, though wouldn't it be simpler to use the global maximum? 

3. Yes, we absolutely do need a top n features; if you'd like to add a new issue, please feel free, but note that this is mentioned in #322 

Is this ready for a review or are you still working on finishing it up?

zjpoh:
@bbengfort Thanks! I'm still working on it. I haven't write unit test yet.
1. I will do that.
2. I normalize by the global maximum. So the largest bar is 100. But since we are stacking multiple bars, the total value of a feature can be larger than 100.
3. I will do that.

yellowbrick/features/importances.py
```diff
@@ -129,23 +130,22 @@ def fit(self, X, y=None, **kwargs):
         # Get the feature importances from the model
         self.feature_importances_ = self._find_importances_param()
 
+        # Get the classes from the model
+        self.classes_ = self._find_classes_param(y)
```
bbengfort: So this is a bit tricky, this parameter is only used if `self.stack == True` and in the case of regression, this is simply the unique values in the regressor (which can be quite a lot of values) and does not refer to classes. I appreciate that this must come before Line 139 because `_find_classes_param` will set `stack=False`. 

Instead, I propose that you use `yellowbrick.utils.is_classifier` to check if the model is a classifier, then fetch the classes from the estimator, otherwise raise an exception that with stack=True only multi-output classifiers are appropriate. It would be best if `self.classes_` was only stored for classification targets. Would you mind also adding documentation for the `self.classes_` attribute in the docstring?

Note that alternatively you could use [`sklearn.utils.multiclass.type_of_target`](http://scikit-learn.org/dev/modules/generated/sklearn.utils.multiclass.type_of_target.html) to check the type of the target; if the result is in `{"binary", "multiclass", "mutliclass-multioutput"}` then it is a classification target, if it is in `{"continuous", "continuous-multioutput"}` then it is a regressor; but this is more important for #508 and #524 than it is for this.

yellowbrick/features/importances.py
```diff
@@ -185,7 +190,25 @@ def draw(self, **kwargs):
...
+                left = [left_arr[j, int(self.feature_importances_[idx][j]>0)]
+                        for j in range(len(self.feature_importances_[idx]))]
+
+                self.ax.barh(pos, self.feature_importances_[idx], left=left,
+                             color=colors[idx], label=self.classes_[idx])
+
+                left_arr[:, 0] += np.minimum(self.feature_importances_[idx],
+                                             zeros)
+                left_arr[:, 1] += np.maximum(self.feature_importances_[idx],
+                                             zeros)
```
bbengfort: Very good implementation for the stacked bar chart! I always find the matplotlib stacked charts tricky to implement; we probably should create a helper function do this if we're going to have other stacked bar charts in Yellowbrick! Would you mind creating an issue for this if you think that's a good idea?

yellowbrick/features/importances.py
```diff
@@ -207,9 +230,24 @@ def finalize(self, **kwargs):
...
+        Searches the wrapped model for the classes_ parameter.
+        """
+        for attr in ["classes_"]:
+            try:
+                return getattr(self.estimator, attr)
+            except AttributeError:
+                continue
+
+        self.stack = False
+        return np.unique(y)
```
bbengfort: This was a good way to solve the coloring for each of the stacked pieces, nice problem solve! 

Depending on how you decide to address my earlier comment about using `is_classifier`; it would probably be better for this method to return `None` instead of `np.unique` in the case of a regressor; otherwise, we may store arbitrarily large data; e.g. up to the size of `y` for no purpose.

yellowbrick/features/importances.py
```diff
@@ -257,7 +295,8 @@ def _is_fitted(self):
 ##########################################################################
 
 def feature_importances(model, X, y=None, ax=None, labels=None,
-                        relative=True, absolute=False, xlabel=None, **kwargs):
+                        relative=True, absolute=False, xlabel=None,
+                        stack=False, **kwargs):
```
bbengfort: Just a reminder if you wouldn't mind modifying the docstring to include the `stack` attribute; I'd appreciate it!

tests/test_features/test_importances.py
```diff
@@ -248,6 +248,28 @@ def test_multi_coefs(self):
...
+        model = MockEstimator()
+        model.make_importance_param(value=coefs)
+        model.make_classes(value=np.arange(coefs.shape[0]))
+
+        visualizer = FeatureImportances(model, relative=False, stack=True)
+        visualizer.fit(coefs, np.random.rand(100))
+        visualizer.poof()
+
+        npt.assert_equal(visualizer.feature_importances_.shape, (3, 9))
+        assert hasattr(visualizer.ax, 'legend')
```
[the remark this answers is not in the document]
zjpoh: Sure. I will add an image similarity test.

yellowbrick/features/importances.py
```diff
@@ -94,13 +95,13 @@ class FeatureImportances(ModelVisualizer):
     """
 
     def __init__(self, model, ax=None, labels=None, relative=True,
-                 absolute=False, xlabel=None, **kwargs):
+                 absolute=False, xlabel=None, stack=False, **kwargs):
```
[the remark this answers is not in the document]
zjpoh: Yeap. I think a test case for `stack=False` is redundant.

I will add `stack` to documentation.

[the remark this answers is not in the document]
bbengfort: Awesome, thanks!

tests/test_features/test_importances.py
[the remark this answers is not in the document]
bbengfort: Cheers!

yellowbrick/features/importances.py
zjpoh: Thanks for pointing this out.

I will do a `yellowbrick.utils.is_classifier` check before line 139 and add `self.classes_` to docstring.

yellowbrick/features/importances.py
zjpoh: Good point!

yellowbrick/features/importances.py
zjpoh: Yeap~ I will add it to docstring.

zjpoh: Yeap~ I will add it to docstring.

rebeccabilbro:
@zjpoh - your list of estimators with multidimensional coefs is excellent and will be SO helpful for other visualizers that we're working on (including the ROCAUC one I'm working on now, #252 ;D ). Thank you for putting it together!!

yellowbrick/features/importances.py
rebeccabilbro: +1

tests/test_features/test_importances.py
```diff
@@ -248,6 +249,24 @@ def test_multi_coefs(self):
 
         npt.assert_equal(visualizer.feature_importances_.ndim, 1)
 
+    @pytest.mark.xfail(
+        sys.platform == 'win32', reason="images not close on windows"
+    )
```
bbengfort: Thanks for checking the AppVeyor CI tests - what was the RMSE out of curiosity?

tests/test_features/test_importances.py
```diff
@@ -248,6 +249,24 @@ def test_multi_coefs(self):
...
+        """
+        X_iris, y_iris = load_iris(True)
+        X_iris_pd = pd.DataFrame(X_iris, columns=['f1', 'f2', 'f3', 'f4'])
+
+        viz = FeatureImportances(LogisticRegression(), stack=True)
+        viz.fit(X_iris_pd, y_iris)
+        viz.poof()
+
+        npt.assert_equal(viz.feature_importances_.shape, (3, 4))
+        self.assert_images_similar(viz)
```
bbengfort: Nice!

tests/test_features/test_importances.py
```diff
@@ -248,6 +249,24 @@ def test_multi_coefs(self):
...
         npt.assert_equal(visualizer.feature_importances_.ndim, 1)
 
+    @pytest.mark.xfail(
+        sys.platform == 'win32', reason="images not close on windows"
+    )
+    def test_multi_coefs_stacked(self):
+        """
+        Test stack plot with multidimensional coefficients
+        """
+        X_iris, y_iris = load_iris(True)
```
bbengfort: Very good test, thank you so much! Another thing we do in testing is use [`sklearn.datasets.make_classification`](http://scikit-learn.org/stable/modules/generated/sklearn.datasets.make_classification.html) to specify exactly how the classification behaves with some fixed random state. The baseline image you created with Iris is a good test, though, so no need to jump through any other hoops!

tests/test_features/test_importances.py
```diff
@@ -248,6 +249,24 @@ def test_multi_coefs(self):
...
+        X_iris_pd = pd.DataFrame(X_iris, columns=['f1', 'f2', 'f3', 'f4'])
```
bbengfort: Technically, pandas is not a dependency of Yellowbrick, even though it is a test dependency; still just to guard against confusion, we usually do `pytest.mark.skipif(pd is None)`; I will go ahead and make this change when I merge though - no worries!

tests/test_features/test_importances.py
```diff
@@ -350,6 +369,21 @@ def test_find_importances_param_not_found(self):
         with pytest.raises(YellowbrickTypeError):
             visualizer._find_importances_param()
 
+    def test_find_classes_param_not_found(self):
```
bbengfort: Good test, and nice work on the `MockEstimator`!

yellowbrick/features/importances.py
```diff
@@ -72,6 +73,11 @@ class FeatureImportances(ModelVisualizer):
         The label for the X-axis. If None is automatically determined by the
         underlying model and options provided.
 
+    stack : bool, default: False
```
bbengfort: Good docstring!

yellowbrick/features/importances.py
```diff
@@ -129,23 +138,28 @@ def fit(self, X, y=None, **kwargs):
...
+            self.classes_ = None
+            self.stack = False
 
-        # TODO - as an alternative to the above flattening approach, explore an
-        # alternative visualize that uses the array shape to create a stacked bar chart
-        # of feature importances for each class/feature combination
+        # If self.stack = True and feature importances is a multidim array,
+        # we're expecting a shape of (n_classes, n_features)
+        # therefore we flatten by taking the average by
+        # column to get shape (n_features,)  (see LogisticRegression)
```
bbengfort: Awesome, thank you for updating the comment!

yellowbrick/features/importances.py
```diff
@@ -207,9 +246,27 @@ def finalize(self, **kwargs):
...
+        """
+        Searches the wrapped model for the classes_ parameter.
+        """
+        for attr in ["classes_"]:
+            try:
+                return getattr(self.estimator, attr)
+            except AttributeError:
+                continue
+
+        raise YellowbrickTypeError(
```
bbengfort: Technically, at this point, you could use `np.unique` since you're using the `is_classifier` check and you know this is a classifier. However, I think this is just fine and does inform the user about what's going on.

bbengfort:
@majidaldo Thank you for your suggestion and a gridded heatmap is a very interesting idea. I get where you're coming from, particularly in classification land (and in text analysis where there are a huge number of features). I think there is a bit of nuance to this. `FeatureImportances` shows the impact of features with respect to the model. I think your question is more closely associated with "most informative features" to a specific instance or instances. Having a `MostInformativeFeatures` visualizer that utilized the `predict()` method as its primary drawing method would be incredibly useful and is a very good suggestion. Would you mind opening a feature request issue for that?

Issue #470, opened 2018-06-07
Update examples.ipynb to include new Visualizers


lwgray:
I noticed that the examples.ipynb was out-of-date.  
Here are the things that need to be changed/updated.

- [ ] Include new visualizers
    - Class Prediction Error
    - CVScores
    - Manifold
    - Feature Importance
    - Recursive Feature Elimination
    - Classification Report w/ support
    - Discrimination Threshold
    - Validation Curve
    - Learning Curve)

- [ ]  Remove deprecated Visualizers (e.g ScatterVisualizer)

- [ ]  When the visualizers are executed some of them raise warnings.  For aesthetics of the notebook, we need to turn off all warnings.  See example code below

```
 import warnings
warnings.filterwarnings('ignore')
```

- [x]  Fix issue #436 as it partially addresses some of these concerns.

@DistrictDataLabs/team-oz-maintainers

Kautumn06:
If no one else is working on this, I'd be happy to start making some of the updates that Larry suggested.

@DistrictDataLabs/team-oz-maintainers

Kautumn06:
I started to update `examples.ipynb` today and I noticed that it is already quite long, and that does not even include all of the additional visualizers that I need to add. So I think it might be worth considering breaking apart the examples into separate notebooks, such as classification, regression, clustering, feature analysis, model selection, and text modeling. 

Once these notebooks are created, we could then always create a "Greatest Hits" type of notebook that contains some of our favorites from each group (and "Greatest Hits, Vol. 2" if we have a hard time selecting our favorites).

Of course, if anyone has any suggestions or different ideas, please just let me know!
@DistrictDataLabs/team-oz

bbengfort:
@Kautumn06 that would be fine with me, just keep in mind that we have [Binder](https://mybinder.org/v2/gh/DistrictDataLabs/yellowbrick/develop?filepath=examples%2Fexamples.ipynb) set up, and we'd want to configure it to point to all of these notebooks.

lwgray:
@Kautumn06 Thanks for tackling this 😄

joelvanveluwen:
@Kautumn06 @lwgray - are you guys still working on this? @Kautumn06 if you are working on sub-notebooks I'm happy to start on a few!

Kautumn06:
@joelvanveluwen Thank you for your interest! Yes, I am still working on breaking the original into smaller, more focused notebooks.

[...]

PR #517, opened 2018-07-20
Add Examples Notebook of Regression Visualizers (#470)

Kautumn06:
Initially, I had intended to update the original `examples.ipynb`, which contains a brief description and example of the current visualizers available in Yellowbrick. However, that notebook was already quite long, even before adding our newest visualizers. So I decided to create a new series of notebooks that focus on only one of our main categories of visualizers.

This notebook contains examples of Yellowbrick's regression visualizers. To make it easier to review, I extracted the exact code and description text from the original `examples.ipynb` and tried to make only minimal changes. Some of the more notable changes are:

* Uses the new `yellowbrick.datsets` module to load the data
* Adds text from our documentation to expands on the Visualizer descriptions
* Adds an additional example of the `ResidualsPlot` to show how to turn off the optional histogram
residuals-without-histogram

* Replaces the code used to create the list of alphas for the `AlphaSelection` Visualizer with the code used in the documentation. I think this plot is preferable since it clearly shows the error dropping as it approaches the optimal value of alpha and then rising again after it passes it.
_Original `AlphaSelection` Visualizer Plot_
original-alphaselection
_New `AlphaSelection` Visualizer Plot_
new-alphaselection

* Doesn't include the `ManualAlphaSelection` Visualizer. The example in the original is entirely commented out and when returns an error when run, so I've left it out for now until I've had a chance to find what is causing the problem.
manual-alpha


Kautumn06:
@bbengfort thanks for your feedback! I'll update #470 and the `README.md` file, and I've made a note to delete `examples/download.py` once the switch to the new `load_dataset` functionality is completed. 

For the remaining notebooks, I should be able to have those finished before the next release, since I've already done a lot of work on them. At first, I started making all of the changes in a single branch, but once I realized that this was going to be a bigger project than I expected, I was worried about including all of them in a single PR. So I thought it would be more manageable to submit them individually, since I know it is a lot to review and it's difficult to compare the changes since they're new files.

bbengfort:
@Kautumn06 perfect, thank you! And you're right, I think doing them in individual PRs makes complete sense. I'll go ahead and merge this into develop now.

PR #519, opened 2018-07-23
Adding MissingValues to Contrib

ndanielsen:
## Adding Two Visualizers to `yellowbrick.contrib`for Missing Values

Adding a Barchart and Dispersion chart specialized to plot missing values. Each of these has a different behavior when the `y` targets are pasted into fit like this `viz.fit(X, y=y)`:

Docs and tests have been added. One small bug to fix later is a small warning appears when the visualizers are fit without targets:

`No handles with labels found to put in legend.`

### Without targets supplied
download-1

### With targets supplied
download-2

### Without targets supplied
download-3

### With targets supplied
download-4


ndanielsen:
@DistrictDataLabs/team-oz-admins ready for a review

yellowbrick/contrib/missing/bar.py
```diff
@@ -0,0 +1,247 @@
...
+
+    When y targets are supplied to fit, the output is a stacked bar chart where
+    each color corresponds to the total NaNs for the feature in that column.
+
+    Parameters
+    ----------
+    alpha : float, default: 0.5
+        A value for bending elments with the background.
+
+    marker : matplotlib marker, default: |
```
bbengfort: alpha and marker params left over from Dispersion copy and paste? missing bar width param.

yellowbrick/contrib/missing/bar.py
```diff
@@ -0,0 +1,247 @@
...
```
bbengfort: Docstring comments from above.

yellowbrick/contrib/missing/bar.py
```diff
@@ -0,0 +1,247 @@
...
+
+        # Set up classifier score visualization properties
+        if self.classes_ is not None:
+            n_colors = len(self.classes_)
+        else:
+            n_colors = None
+
+        self.colors = color_palette(kwargs.pop('colors', None), n_colors)
+
+    def get_nan_col_counts(self, **kwargs):
```
bbengfort: This method feels like something that should be on the `MissingDataVisualizer` -- or at least the part with the `nan_matrix` since that is the shared functionality with the Dispersion plot.

yellowbrick/contrib/missing/base.py
```diff
@@ -0,0 +1,69 @@
...
+        if is_dataframe(X):
+            self.X = X.values
+            if self.features_ is None:
+                self.features_ = X.columns
+        else:
+            self.X = X
+
+        self.y = y
+
+        super(MissingDataVisualizer, self).fit(X, y, **kwargs)
```
bbengfort: `fit()` should return self; which I think the `DataVisualizer` should do, so adding a return here would do it.

yellowbrick/contrib/missing/base.py
```diff
@@ -0,0 +1,69 @@
...
+            Returns the instance of the transformer/visualizer
+        """
+        if is_dataframe(X):
+            self.X = X.values
+            if self.features_ is None:
+                self.features_ = X.columns
+        else:
+            self.X = X
+
+        self.y = y
```
bbengfort: I'm a bit squeamish storing X and y on the visualizer - for a lot of workflows they could be fairly large arrays that aren't intended to be held in memory for longer than it takes to transform them (e.g. multiple visualizers would duplicate the arrays in memory), and they refer to data that's being passed through transformers and manipulated, so if stored by reference and not copied, these values would change as the data moved to the next stage of processing. 

It would be my preference that in the bar and dispersion plots, they override `fit()` where most of the work is conducted, including calls to `get_nan_values` that pass the data in as arguments. It would make the class a bit easier to test as well.

docs/api/contrib/index.rst
```diff
@@ -11,3 +11,4 @@ The ``yellowbrick.contrib`` package contains a variety of extra tools and experi
    boundaries
    statsmodels
    scatter
+   missing/index
```
bbengfort: Thanks for adding documentation for the visualizers; this is a very complete submission, with two very cool visualizers!

examples/ndanielsen/horse-colic.attrs
```diff
@@ -0,0 +1,29 @@
...
+packed cell volume,Numeric,46.295,10.419,29 (10%)
+total protein,Numeric,24.457,27.475,33 (11%)
+abdominocentesis appearance,Numeric,2.037,0.805,165 (55%)
+abdomcentesis total protein,Numeric,3.02,1.969,198 (66%)
+outcome,Numeric,1.552,0.737,1 (0%)
+surgical lesion,Numeric,1.363,0.482,0 (0%)
+type of lesion 1,Numeric,3657.88,5399.514,0 (0%)
+type of lesion 2,Numeric,90.227,649.569,0 (0%)
+type of lesion 3,Numeric,7.363,127.537,0 (0%)
+cp_data,Numeric,1.67,0.471,0 (0%)
```
bbengfort: Wow, that's quite the dataset!

tests/test_contrib/test_missing/test_dispersion.py
```diff
@@ -0,0 +1,124 @@
...
+        # add nan values to a range of values in the matrix
+        X[X > 1.5] = np.nan
+
+        features = [str(n) for n in range(20)]
+        classes = ['Class A', 'Class B']
+        viz = MissingValuesDispersion(features=features, classes=classes)
+        viz.fit(X, y=y)
+        viz.poof()
+
+        self.assert_images_similar(viz, tol=self.tol)
```
bbengfort: Great tests for both visualizers!

yellowbrick/contrib/missing/bar.py
```diff
@@ -0,0 +1,247 @@
...
+##########################################################################
+## MissingValues Visualizer
+##########################################################################
+
+
+class MissingValuesBar(MissingDataVisualizer):
+    """The MissingValues Bar visualizer creates a bar graph that lists the total
+    count of missing values for each selected feature column.
+
+    When y targets are supplied to fit, the output is a stacked bar chart where
```
bbengfort: Love the stacked bar charts!

yellowbrick/contrib/missing/bar.py
```diff
@@ -0,0 +1,247 @@
...
+            if self.classes_ is not None:
+                label = self.classes_[index]
+
+            color = self.colors[index]
+
+            self.ax.barh(self.ind - self.width / 2, nan_col_counts, self.width,
+                        color=color, label=label, left=bottom_chart)
+
+            # keep track of counts to build on stacked
+            bottom_chart = nan_col_counts
```
bbengfort: Nice - I've always found the stacked bar charts in matplotlib a bit finicky, but I like your implementation with the loop and the bottom_chart tracking!

yellowbrick/contrib/missing/bar.py
```diff
@@ -0,0 +1,247 @@
...
+        else:
+            nan_matrix = self.X.astype(np.float)
+
+        if self.y is None:
+            nan_col_counts = [np.count_nonzero(np.isnan(col)) for col in nan_matrix.T]
+            return nan_col_counts
+
+        else:
+            # add in counting of np.nan per target y by column
+            nan_counts = []
```
bbengfort: This array seems like something the user would be interested in, e.g. stored in `self.nan_counts_` since it's a learned parameter. It would work for both `nan_col_counts` and `nan_counts` - scikit-learn routinely changes the shape of its underlying learned parameters depending on the input to the transformer.

yellowbrick/contrib/missing/dispersion.py
```diff
@@ -0,0 +1,226 @@
...
+            nan_matrix = np.zeros(self.X.shape)
+            nan_matrix[mask] = np.nan
+
+        else:
+            nan_matrix = self.X.astype(float)
+
+        if self.y is None:
+            return np.argwhere(np.isnan(nan_matrix))
+        else:
+            nan_locs = []
```
bbengfort: I think the user would also be interested in accessing this property, e.g. `self.nan_locs_` as a learned parameter.

yellowbrick/contrib/missing/dispersion.py
```diff
@@ -0,0 +1,226 @@
...
+        different color.
+        """
+        nan_locs = self.get_nan_locs()
+        if y is None:
+            x_, y_ = list(zip(*nan_locs))
+            self.ax.scatter(x_, y_, alpha=self.alpha, marker=self.marker, label=None)
+        else:
+            self.draw_multi_dispersion_chart(nan_locs)
+
+    def draw_multi_dispersion_chart(self, nan_locs):
```
bbengfort: This is also a very cool feature - and I can see myself using it a lot for my own classification projects!

yellowbrick/features/base.py
```diff
@@ -218,7 +218,7 @@ def fit(self, X, y=None, **kwargs):
         # Store the classes for the legend if they're None.
         if self.classes_ is None:
             # TODO: Is this the most efficient method?
-            self.classes_ = [str(label) for label in set(y)]
+            self.classes_ = [str(label) for label in np.unique(y)]
```
bbengfort: I've been wanting to do this a while, but I tried it once and it broke our tests -- glad to see it's working with the tests now; guess we've made a bit of progress. I think the issue was that np.unique` orders the return value alphabetically, where as `set` does not guarantee any particular order.

ndanielsen:
@bbengfort thanks for the review, I'll squash-merge in and then begin to tackle the items in order to lift this up into the main package.

ndanielsen:
I do like the idea of the `contrib` for building up visualizers, tests and docs versus massive, difficult efforts to get it right on the first swing

PR #532, opened 2018-07-29
[MRG] Feature Correlation to Dependent Variable Visualizer

zjpoh:
Add features according to #508

Here are some sample plots.
pearson
mutual_info

Please let me know if there are any modifications you would like to have.

Note: The code is just a bare-bones version. I will add documentation and tests after you think the plot is good.

@bbengfort Can you expand on what you mean by linking this visualizer to the `JointPlot` visualizer' Also, can explain how the ID part of the docstring works?


yellowbrick/target/feature_correlation.py
```diff
@@ -0,0 +1,87 @@
+# yellowbrick.classifier.feature_correlation
+# Class balance visualizer for showing per-class support.
```
bbengfort: Thanks for adding our code header! I see what you mean by your question about the ID line now. Here, the first line should be the module path, e.g. `yellowbrick.target.feature_correlation` and the second line should be a quick description, here `"Feature Correlation to Dependent Variable Visualizer."` 

Everything else looks good, I've already mentioned the ID line, but you don't need to add my name as Author -- I can do that if I update this file. Also note that if you make changes to a file in YB, feel free to add yourself to the author line in the ID tag.

yellowbrick/target/feature_correlation.py
```diff
@@ -0,0 +1,87 @@
...
+from sklearn.feature_selection import (mutual_info_classif,
+                                       mutual_info_regression)
+from scipy.stats import pearsonr
+
+
+##########################################################################
+## Class Feature Correlation
+##########################################################################
+
+class FeatureCorrelation(TargetVisualizer):
```
bbengfort: This is a good name for this visualizer, thanks!

yellowbrick/target/feature_correlation.py
```diff
@@ -0,0 +1,87 @@
...
+
+    method_func = {
```
bbengfort: For static variables like this, I would prefer if they were all caps, e.g. `METHOD_FUNC` and `METHOD_LABEL`. Did you see an example implementation in YB for something like this where the names were lower case? If so, I can review why we did things that way, or fix it if it needs fixing!

yellowbrick/target/feature_correlation.py
```diff
@@ -0,0 +1,87 @@
...
+        'mutual_info': 'Mutual Information',
+        'mutual_info_classif': 'Mutual Information'
+    }
+
+    def __init__(self, ax=None, method='pearson', classification=False,
+                 **kwargs):
+        super(FeatureCorrelation, self).__init__(ax=None, **kwargs)
+
+        if classification and method == 'mutual_info':
+            method = 'mutual_info_classif'
```
bbengfort: I'm glad you inserted this check and thought about how to specify the difference between the two mutual information algorithms. Another possibility is to use `target_type` on fit to determine if the output is `continuous` or `binary`, `multiclass` etc. and you can choose the metric that way. 

Also, I'd recommend that you do a check for `method in self.method_func` and if not, raise an exception `YellowbrickValueError` when the visualizer is instantiated.

yellowbrick/target/feature_correlation.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # Parameters
+        self.set_params(
+            method=method
+        )
+
+    def fit(self, X, y, features_):
```
bbengfort: Are features here the feature names? If so, it would be better to specify these in `__init__` rather than `fit`, and if the feature names aren't provided or discovered off of a `DataFrame` we can just identify them by index (e.g. 0, 1, 2, etc). I think the `FeatureImportances` visualizer had some mechanism for this right?

yellowbrick/target/feature_correlation.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # Calculate Features correlation with target variable
+        self.coef_ = np.array(self.method_func[self.method](X, y))
+
+        # Sort features by correlation
+        sort_idx = np.argsort(self.coef_)
+        self.coef_ = self.coef_[sort_idx]
```
bbengfort: Awesome, sorting by correlation really makes the plot look great I think! We might also want to add an option (which defaults to `False`) to not sort the features by coef, used in the case of the user specifying features by index or something like that.

yellowbrick/target/feature_correlation.py
zjpoh: Yeap. There are three in `yellowbrick.features.rankd`.
<https://github.com/DistrictDataLabs/yellowbrick/blob/933bc47df0ce5b29792023d0023134d31fab30b9/yellowbrick/features/rankd.py#L191>
<https://github.com/DistrictDataLabs/yellowbrick/blob/933bc47df0ce5b29792023d0023134d31fab30b9/yellowbrick/features/rankd.py#L334>
<https://github.com/DistrictDataLabs/yellowbrick/blob/933bc47df0ce5b29792023d0023134d31fab30b9/yellowbrick/features/rankd.py#L448>

yellowbrick/target/feature_correlation.py
zjpoh: Hmm.... Can you explain how to determine if we are dealing with a classification or a regression problem from the target type? For example, y should be `int` in classification problem but having y as `float` will still work. Also, if `y` is age, then it can be `int` and also regression problem. So, perhaps is better for the user to specify the problem type, albeit more work and less elegant.

yellowbrick/target/feature_correlation.py
zjpoh: That's right. If X is a `DataFrame`, then we can get the name from the `DataFrame` column names. If the feature names are not specify, we can just refer them by index.

yellowbrick/target/feature_correlation.py
zjpoh: I think this is a good idea~

yellowbrick/target/feature_correlation.py
bbengfort: Gah, yeh I suspected something like this existed. Bummer - I think the reason this is this way in RankD is because of the subclassing. 

Go ahead and keep it lower case for now and I'll figure out what to do with this later.  

Also, could we name this `correlation_method` and `correlation_label`?

yellowbrick/target/feature_correlation.py
bbengfort: You're right that in general the type of model that the user selects will determine how the model treats the target. In this case, however, the difference between [`mutual_info`](http://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.mutual_info_regression.html#sklearn.feature_selection.mutual_info_regression) and [`mutual_info_classif`](http://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.mutual_info_classif.html#sklearn.feature_selection.mutual_info_classif) is the solver that relies on a continuous or discrete input and will only provide interpretable results based on the form of the target. 

Let's look at all the cases:

- y is continuous: the discrete solver will not work, classification will also not work because of the huge number of possible classes (binning or thresholding is required) 
- y is discrete, small number of classes: the continuous solver will not work, also a regression model would probably not behave sensibly 
- y is discrete, large number of classes: both solvers will perform poorly, though the discrete solver would probably perform a bit better. Both regression and classification models are possible, perhaps regression models (e.g. for age) would be a bit better without some binning into age groups 

Since we don't have a model to check, and because scikit-learn does use [`target_type`](http://scikit-learn.org/dev/modules/generated/sklearn.utils.multiclass.type_of_target.html) to raise exceptions for badly formed problems, I would err on the side of the first two common cases rather than create a possible confusion to handle the third case.

bbengfort:
@zjpoh hope I answered the questions, as always ping me when you are ready for a review! Also, let me know if you want an invite to the slack channel

zjpoh:
@bbengfort I'll let you know when it is ready for review. I gave my personal email to Jason to add me to the slack channel~ Thanks for the invitation.

zjpoh:
@bbengfort This is ready for review.

### Discussions on mutual information calculation:
Instead of automatically selection which mutual information algorithm in scikit-learn to use, I think the better approach is to let the user choose which algorithm to use.

As you pointed out if
* y is continuous, then one have to use `mutual_info_regression` because `mutual_info_classif` will treat each distinct value as a separate class. `mutual_info_regression` will work because it is not based on binning. Instead, the calculation basically rewrite mutual information in terms of entropy, then uses Kozachenko-Leonenko entropy estimator, which is based on k-nearest neighbor statistics.
* y is discrete with small number of classes, then one have to use `mutual_info_classif` instead of `mutual_info_regression`.
* y is discrete with a large number of classes, then the choice of these two algorithms depends is problem dependent.

Another point is that to correctly use these two functions, one have to explicitly use `discrete_features` parameter in `mutual_info_regression` and `mutual_info_classif` to specify if a feature in the discrete or not because similar to target, the determination of whether a feature is continuous or discrete is very important.

I tried using `sklearn.utils.multiclass.type_of_target` to automatically determine if `mutual_info_classif` or `mutual_info_regression` will be used but the output isn't very reliable because the output of any integer target is classification. For example, if target is dollar amount in int, then `mutual_info_classif` will be used. If the same target is converted to float, then `mutual_info_regression` will be used.

yellowbrick/target/feature_correlation.py
```diff
@@ -0,0 +1,336 @@
...
+        """
+        Finalize the drawing setting labels and title.
+        """
+        self.set_title('Features correlation with dependent variable')
+
+        self.ax.set_xlabel(self.correlation_label[self.method])
+
+        self.ax.grid(False, axis='y')
+
+    def _select_mutual_info_algo(self, y):
```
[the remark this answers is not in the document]
zjpoh: Good point. I will remove them~

yellowbrick/target/feature_correlation.py
```diff
@@ -0,0 +1,336 @@
...
+        -------
+        self : visualizer
+            The fit method must always return self to support pipelines.
+        """
+        self._create_labels_for_features(X)
+
+        self._select_features_to_plot(X)
+
+        # Calculate Features correlation with target variable
+        self.corr_ = np.array(
```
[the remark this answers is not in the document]
zjpoh: Good catch

yellowbrick/target/feature_correlation.py
```diff
@@ -0,0 +1,336 @@
...
+                self.features_ = np.arange(0, ncols)
+        else:
+            self.features_ = np.array(self.labels)
+
+    def _select_features_to_plot(self, X):
+        """
+        Select features to plot
+        """
+        if self.feature_index and self.feature_names:
+            raise YellowbrickWarning(
```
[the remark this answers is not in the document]
zjpoh: Good point. I will add that to the comment.

yellowbrick/target/feature_correlation.py
```diff
@@ -0,0 +1,336 @@
...
+                'Both feature_index and feature_names '
+                'are specified. feature_names is ignored'
+            )
+        elif self.feature_index:
+            if (min(self.feature_index) < 0
+                    or max(self.feature_index) >= X.shape[1]):
+                raise YellowbrickValueError('Feature index is out of range')
```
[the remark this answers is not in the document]
zjpoh: Good catch.

yellowbrick/target/feature_correlation.py
```diff
@@ -0,0 +1,336 @@
...
+    >>> viz = FeatureCorrelation()
+    >>> viz.fit(X, y)
+    >>> viz.poof()
+    """
+
+    def _pearsonr(X, y, **kwargs):
+        """
+        Utility function calculate pearson correlation coefficient
+        """
+        return [pearsonr(x, y, **kwargs)[0] for x in np.asarray(X).T]
```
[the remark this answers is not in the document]
zjpoh: Sure.

tests/test_target/test_feature_correlation.py
```diff
@@ -0,0 +1,214 @@
...
+
+from tests.base import VisualTestCase
+from tests.dataset import DatasetMixin
+
+
+##########################################################################
+## Feature Correlation Tests
+##########################################################################
+
+class TestFeatureCorrelationVisualizer(VisualTestCase, DatasetMixin):
```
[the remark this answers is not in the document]
zjpoh: Got it.

tests/test_target/test_feature_correlation.py
```diff
@@ -0,0 +1,214 @@
...
+        """
+        Test FeatureCorrelation visualizer with unknown method
+        """
+        method = 'foo'
+        e = ('Method foo not implement; choose from '
+             'mutual_info-classification, '
+             'mutual_info-regression, '
+             'pearson')
+        with pytest.raises(YellowbrickValueError, match=e):
+            FeatureCorrelation(method=method)
```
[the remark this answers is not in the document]
zjpoh: Thanks for informing. I think regex is better option especially for the list of available methods.

tests/test_target/test_feature_correlation.py
```diff
@@ -0,0 +1,214 @@
...
+            viz.fit(self.X, self.y)
+
+    def test_feature_correlation_select_feature_by_index(self):
+        """
+        Test selecting feature by index
+        """
+        viz = FeatureCorrelation(feature_index=[0, 2, 3])
+        viz.fit(self.X, self.y)
+
+        self.assertEqual(viz.corr_.shape[0], 3)
```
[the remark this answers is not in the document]
zjpoh: Ops. I'm too used to unittest framework....

zjpoh:
@bbengfort This is ready for review. Thanks.

zjpoh:
@bbengfort Sure. Full steam ahead.

Issue #379, opened 2018-04-02
Export Travis-CI images to S3


bbengfort:
Create a TravisCI `after_failure` script that uploads actual images and diffs to S3 so we can diagnose TravisCI failures. 

### Proposal/Issue

It seems like every time we push a commit to GitHub the TravisCI tests fail, even if they were working locally. Right now we're messing with the tolerances but with no real insight into what the key differences are in the CI environment vs locally. To better help us address these issues we should push the failed test images and their diffs to S3 for further inspection. 

Issues to consider:

1. Limit the number of uploads to S3: only upload failed test actual and diff images 
2. Organize by test/commit - multiple travis tests might be running at the same time 
3. Log where the images were uploaded and stored 
4. Long run clean up of these images over time

bbengfort:
This might be the easiest mechanism to do this:

https://github.com/travis-ci/artifacts

PR #533, opened 2018-07-30
Rocauc refactor

rebeccabilbro:
There is currently a [bug](https://github.com/DistrictDataLabs/yellowbrick/issues/252) in `ROCAUC` that triggers an error when a user attempts to use the visualizer for binary classification with a scikit-learn estimator that has only a `decision_function` and not also a `predict_proba`. Estimators with both are fine, and estimators with only `predict_proba` also seem to be fine. Multi-class classification also seems to work without error regardless of estimator. More documentation of the bug in this ([notebook](https://github.com/rebeccabilbro/yellowbrick/blob/rocauc-refactor/examples/rebeccabilbro/rocauc_bug_research.ipynb)). 

It seems like the bug occurs because the default params for `ROCAUC` are `micro=True`, `macro=True`, and `per_class=True`, however, in the case of an estimator with only a decision function being used for a binary classification problem, there will only be enough information to plot a single curve. _Note: estimators that also have a `predict_proba` are safe because of the order in which the `attrs` are checked in the  `_get_y_scores` function._

This PR attempts to address the bug by:
 - in `score`, checking if `y_pred` is of shape (n_samples, ), and if so, setting `self._binary_decision` to `True`
 - raising an error on `score` if `self._binary_decision` is `True` and `micro`, `macro`, and `per_class` are all also set to `True` (_Note: I also moved the `no curves will be drawn` check inside the `self._binary_decision = False` logic inside `score` &mdash; it was originally in `__init__`_)
 - in `score`, if `self._binary_decision`, computing the ROC curve and ROC area for the single class
 - in `draw`, if `self._binary_decision`, drawing the ROC curve and ROC area for the single class

Let me know what you think and if you have any suggestions!


tests/test_classifier/test_rocauc.py
```diff
@@ -88,6 +88,60 @@ def test_binary_rocauc(self):
...
+        # Test that score method successfully returns a value between 0 and 1
+        assert 0 <= s <= 1
+
+        # Check the scores
+        self.assertEqual(len(visualizer.fpr.keys()), 1)
+        self.assertEqual(len(visualizer.tpr.keys()), 1)
+        self.assertEqual(len(visualizer.roc_auc.keys()), 1)
+
+        # Compare the images
+        visualizer.poof()
```
bbengfort: The test is failing on CI with an RMSE of 6.8 - I'm wondering if it's a font issue; removing the call to `poof()` simplifies the test and may get it to pass?

tests/test_classifier/test_rocauc.py
```diff
@@ -224,10 +278,12 @@ def test_rocauc_no_curves(self):
...
         # Create and fit the visualizer
+        visualizer = ROCAUC(LogisticRegression(), per_class=False, macro=False, micro=False)
+        visualizer.fit(self.binary.X.train, self.binary.y.train)
+
+        # Attempt to score the visualizer
         with pytest.raises(YellowbrickValueError, match="no curves will be drawn"):
-            ROCAUC(
-                LogisticRegression(), per_class=False, macro=False, micro=False
-            )
+            visualizer.score(self.binary.X.test, self.binary.y.test)
```
bbengfort: Should we create a test that checks that an exception is raised when `self._binary_decision is True` and `micro`, `macro`, and `per_class` are all also set to `True`?

tests/test_classifier/test_rocauc.py
```diff
@@ -88,6 +88,60 @@ def test_binary_rocauc(self):
         visualizer.poof()
         self.assert_images_similar(visualizer)
 
+    def test_binary_probability_decision(self):
```
bbengfort: Stellar test cases for the models that have various combinations of `decision_function` and `predict_proba`! Very nice work!

yellowbrick/classifier/rocauc.py
```diff
@@ -77,15 +77,21 @@ class ROCAUC(ClassificationScoreVisualizer):
 
     micro : bool, default = True
         Plot the micro-averages ROC curve, computed from the sum of all true
-        positives and false positives across all classes.
+        positives and false positives across all classes. Micro is not defined
+        for binary classification problems with estimators with only a
+        decision_function method.
```
bbengfort: Thank you for updating the reader in the doc string!

yellowbrick/classifier/rocauc.py
```diff
@@ -168,8 +169,30 @@ def score(self, X, y=None, **kwargs):
         # Compute the predictions for the test data
         y_pred = self._get_y_scores(X)
 
-        # # Classes may be label encoded so only use what's in y to compute.
-        # # The self.classes_ attribute will be used as names for labels.
+        # Check to see if shape of predictions signals a binary decision
+        if len(y_pred.shape) == 1:
```
bbengfort: I was a bit confused about this for a while, I didn't understand how the dimensionality of `y_pred` determined if the decision was binary or not. I had to realize that `_get_y_scores` is calling either the `decision_function` or `predict_proba`, which should return a 2D array. 

Would it be possible to explain this a bit more in the comment above for future us who are also likely to forget as I did? 

Also, note that `len(y_pred.shape) == 1` is equivalent to `y_pred.ndim == 1`, which may also help explain what we're checking in this condition.

yellowbrick/classifier/rocauc.py
```diff
@@ -168,8 +169,30 @@ def score(self, X, y=None, **kwargs):
...
+            self._binary_decision = True
+
+            # Raise an error if it's a binary decision and user has set micro,
+            # macro, or per_class to True
+            if self.micro or self.macro or self.per_class:
+                raise ModelError(
```
bbengfort: I love this exception message - it's not that there is a bug in ROCAUC, it's just that there are a set of conditions when the defaults don't work - this error explains how to fix it when that case is encountered and gives our users a clear path to success!

yellowbrick/classifier/rocauc.py
```diff
@@ -168,8 +169,30 @@ def score(self, X, y=None, **kwargs):
...
+                    "binary classification for estimators with only "
+                    "decision_function methods; set micro, macro, and "
+                    "per-class params to False."
+                )
+        else:
+            self._binary_decision = False
+            # If it's not a binary decision, at least one of micro, macro, or
+            # per_class must be True
+            if not self.micro and not self.macro and not self.per_class:
+                raise YellowbrickValueError(
```
bbengfort: Absolutely the right place to move this to.

yellowbrick/classifier/rocauc.py
```diff
@@ -396,15 +435,21 @@ def roc_auc(model, X, y=None, ax=None, **kwargs):
 
     micro : bool, default = True
         Plot the micro-averages ROC curve, computed from the sum of all true
-        positives and false positives across all classes.
+        positives and false positives across all classes. Micro is not defined
+        for binary classification problems with estimators with only a
+        decision_function method.
```
bbengfort: Good catch updating the quick method docs as well!

yellowbrick/classifier/rocauc.py
```diff
@@ -219,7 +247,16 @@ def draw(self):
...
+        if self._binary_decision == True:
+            self.ax.plot(
+                self.fpr[0], self.tpr[0],
+                label='ROC for binary decision, AUC = {:0.2f}'.format(
+                        self.roc_auc[0]
+                )
+            )
+
+        # If per-class plotting is requested, plot ROC curves for each class
         if self.per_class:
```
bbengfort: Just a quick check to make sure I'm following along - if `self._binary_decision is True`, then an exception will already have been raised if `per_class`, `micro`, or `macro` is true, which is why we can short-circuit the conditionals. I can't see the entirety of the function in the review; but if `self._binary_decision is True` can we return  from there, or is there work at the bottom of the function after the conditionals that needs to be done?

tests/test_classifier/test_rocauc.py
rebeccabilbro: Great idea - I actually just added three to check for each possible combo, just in case.

tests/test_classifier/test_rocauc.py
rebeccabilbro: Thanks!

yellowbrick/classifier/rocauc.py
rebeccabilbro: Yep, it took a bit of work to figure this out for me, too -- I just added a (maybe too lengthy) description of what's happening here in a comment inside the code. Hopefully that helps others and future us! Also switched to `ndim`.

yellowbrick/classifier/rocauc.py
rebeccabilbro: There's a bit more logic that happens in the function. Here's the whole thing:

```
        if y_pred.ndim == 1:
            self._binary_decision = True

            # Raise an error if it's a binary decision and user has set micro,
            # macro, or per_class to True
            if self.micro or self.macro or self.per_class:
                raise ModelError(
                    "Micro, macro, and per-class scores are not defined for "
                    "binary classification for estimators with only "
                    "decision_function methods; set micro, macro, and "
                    "per-class params to False."
                )
        else:
            self._binary_decision = False
            # If it's not a binary decision, at least one of micro, macro, or
            # per_class must be True
            if not self.micro and not self.macro and not self.per_class:
                raise YellowbrickValueError(
                    "no curves will be drawn; specify micro, macro, or per_class"
                )
```

rebeccabilbro:
@bbengfort - ok, just pushed all those updates. One question on your suggestion to add a new issue  for us to capture the binary decision case in the YB docs; #252 does have a bullet for docs, and this current PR won't close the issue because there are still a couple of todos associated with the issue. I think instead of opening a new issue, once this is merged in, I'll just open a new PR to wrap up the remaining stuff from the issue. Sound good?

yellowbrick/classifier/rocauc.py
bbengfort: Not too lengthy, I think it was very descriptive, thank you!

bbengfort:
@rebeccabilbro yes, opening a new PR to wrap up the remaining stuff sounds good to me; I'm going to clone your branch locally to see if I can help repair the image comparison test; not sure why it's still failing. Give me a few minutes and a whiskey!

bbengfort:
@rebeccabilbro no joy - I tried all my tricks ... no idea why the tolerance is off. I just increased the tolerance, which should pass the tests. There is really nothing we can do until we have some way of seeing the diffs from Travis and AppVeyor, RE #379

bbengfort:
@rebeccabilbro all the tests pass; am going to go ahead and merge -- thanks for working on this!

PR #535, opened 2018-07-31
Implement Dataset and Corpus data loaders.

bbengfort:
This is a significant change that perhaps we should save for the 1.0 release of Yellowbrick. Datasets, instead of being simple numpy arrays, are now complex objects that provide a lot of functionality similar to the sklearn Bunch object; e.g. access to README and meta information, the ability to transform data into arrays or data frames and more.

This represents a complete overhaul of how datasets in Yellowbrick are managed and will require updates to the documentation and tests.

Still a work on progress. 

Related to #527 #512 #500 #441 #424 

Fixes #416

PR #540, opened 2018-08-02
Finish up the ROCAUC docs per issue #252

rebeccabilbro:
This pull request updates the documentation for the `ROCAUC` Visualizer and resolves the final tasks associated with #252:

 - adds documentation and examples of the multi-class ROCAUC plots to the docs 
 - adds multi-class ROCAUC to gallery 
 - simplify the ROCAUC example in the docstrings for the ROCAUC class method and quick method, and switch to YB data instead of `sklearn.datasets`
 - adds a warning to the documentation about the [decision function bug](https://github.com/DistrictDataLabs/yellowbrick/pull/533) that we found with binary classification, which has been fixed in `develop` but won't be in `master` until the next release.


docs/api/classifier/rocauc.rst
```diff
@@ -18,29 +21,74 @@ sensitivity and specificity.
...
+
+.. code::
+
+    # Load multi-class classification dataset
+    data = load_game("occupancy")
+
+    classes = ["win", "loss", "draw"]
+
+    # Encode the non-numeric columns
+    game.replace({'loss':-1, 'draw':0, 'win':1, 'x':2, 'o':3, 'b':4}, inplace=True)
```
lwgray: Where is the ``game`` variable coming from?

rebeccabilbro: Here `game` is not a variable but an dataframe that holds the win, lose or draw game dataset - hopefully that is clarified in my most recent push!

yellowbrick/classifier/rocauc.py
```diff
@@ -124,18 +124,16 @@ class classification is not defined for binary classification problems
...
-    >>> from sklearn.datasets import load_breast_cancer
     >>> from yellowbrick.classifier import ROCAUC
     >>> from sklearn.linear_model import LogisticRegression
     >>> from sklearn.model_selection import train_test_split
-    >>> data = load_breast_cancer()
-    >>> X = data['data']
-    >>> y = data['target']
+    >>> data = load_data("occupancy")
+    >>> features = ["temp", "relative humidity", "light", "C02", "humidity"]
     >>> X_train, X_test, y_train, y_test = train_test_split(X, y)
```
Kautumn06: In a lot of our documentation, even when we just use the default test size, `test_size=0.25`, we tend to explicitly state it, so we might want to add it here.

docs/api/classifier/rocauc.rst
Kautumn06: When loading the new datasets, I believe that you don't have to pass "game" to the function. You should be able to simply use: `game = load_game()`

rebeccabilbro: Thanks @Kautumn06! Just updated that.

lwgray: Exactly @Kautumn06 that is what I was trying to say

rebeccabilbro: Ok gotcha @lwgray ! Fixed now.

docs/api/classifier/rocauc.rst
```diff
@@ -3,44 +3,92 @@
...
     classes = ["unoccupied", "occupied"]
 
     # Extract the numpy arrays from the data frame
-    X = data[features].as_matrix()
-    y = data.occupancy.as_matrix()
+    X = data[features].values
+    y = data.occupancy.values
 
     # Create the train and test data
     X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2)
```
Kautumn06: Since `load_occupancy()` contains time series data, should we use `sklearn.model_selection.TimeSeriesSplit` instead of `sklearn.model_selection.train_test_split`. `train_test_split` assumes the observations are independent and shuffles the data points by default when making the split. However, this wouldn't be true for time series datasets such as `occupancy` (which includes per-minute observations from February 4, 2015 through February 18, 2015). Here is a short excerpt from `TimeSeriesSplit`'s [documentation](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.TimeSeriesSplit.html):

> Time Series cross-validator
> 
> Provides train/test indices to split time series data samples that are observed at fixed time intervals, in train/test sets. In each split, test indices must be higher than before, and thus shuffling in cross validator is inappropriate.

docs/api/classifier/rocauc.rst
```diff
@@ -3,44 +3,92 @@
...
+
+This leads to another metric, area under the curve (AUC), which is a computation of the relationship between false positives and true positives. The higher the AUC, the better the model generally is. However, it is also important to inspect the "steepness" of the curve, as this describes the maximization of the true positive rate while minimizing the false positive rate.
 
 .. code:: python
 
     from sklearn.model_selection import train_test_split
 
     # Load the classification data set
-    data = load_data("occupancy")
+    data = load_occupancy()
```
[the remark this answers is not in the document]
rebeccabilbro: Ah got it, this is clearly why we need #535 so badly! Reverting for now until we have to update all the docs again next week once #535 is merged in ;)

docs/api/classifier/rocauc.rst
```diff
@@ -3,44 +3,92 @@
...
 
     # Specify the features of interest and the classes of the target
     features = ["temperature", "relative humidity", "light", "C02", "humidity"]
     classes = ["unoccupied", "occupied"]
 
     # Extract the numpy arrays from the data frame
-    X = data[features].as_matrix()
-    y = data.occupancy.as_matrix()
+    X = data[features].values
```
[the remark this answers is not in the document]
rebeccabilbro: Got it

docs/api/classifier/rocauc.rst
rebeccabilbro: This is fascinating. Since it seems to apply more broadly to our use of the occupancy dataset throughout our docs, though, perhaps it is better suited to a new issue? Want to open one?

Kautumn06: Sure @rebeccabilbro! We do use the `occupancy` dataset in a lot of our examples, so that's a good idea.

Issue #539, opened 2018-08-01
TSNE size & title bug


rebeccabilbro:
**Describe the bug**
Looks like our `TSNEVisualizer` might have a bug that causes an error on instantiation if either the `size` or `title` parameters are used.

**To Reproduce**

```python
from yellowbrick.text import TSNEVisualizer
from sklearn.feature_extraction.text import TfidfVectorizer

corpus = load_data('hobbies')
tfidf = TfidfVectorizer()
docs = tfidf.fit_transform(corpus.data)
labels = corpus.target

tsne = TSNEVisualizer(size=(1080, 720))
```

or 

```
tsne = TSNEVisualizer(title="My Special TSNE Visualizer")
```

**Dataset**
This bug was triggered using the YB hobbies corpus.

**Expected behavior**
Users should be able to influence the size of the visualizer on instantiation using the `size` parameter and a tuple with `(width, height)` in pixels, and the title of the visualizer using the `title` parameter and a string.

**Traceback**

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-59-120fbfcec07c> in <module>()
----> 1 tsne = TSNEVisualizer(size=(1080, 720))
      2 tsne.fit(labels)
      3 tsne.poof()

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/yellowbrick/text/tsne.py in __init__(self, ax, decompose, decompose_by, labels, classes, colors, colormap, random_state, **kwargs)
    180 
    181         # TSNE Parameters
--> 182         self.transformer_ = self.make_transformer(decompose, decompose_by, kwargs)
    183 
    184     def make_transformer(self, decompose='svd', decompose_by=50, tsne_kwargs={}):

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/yellowbrick/text/tsne.py in make_transformer(self, decompose, decompose_by, tsne_kwargs)
    234         # Add the TSNE manifold
    235         steps.append(('tsne', TSNE(
--> 236             n_components=2, random_state=self.random_state, **tsne_kwargs)))
    237 
    238         # return the pipeline

TypeError: __init__() got an unexpected keyword argument 'size'

```
or for `title`:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-64-92c88e0bdd33> in <module>()
[...]

PR #542, opened 2018-08-02
Fixes TSNEVisualizer title size bug

rebeccabilbro:
The `TSNEVisualizer` currently has a bug (#539) that causes an error on instantiation if either the `size` or `title` parameters are used. The issue is that we pass the `**kwargs` (including any title and size provided by the user) into the internal sklearn `TSNE` transform pipeline, which doesn't recognize those params. 

I've fixed the bug by popping those two from the `**kwargs` in `__init__`, and using them to set the `self._size` (part of the `Visualizer` base class) and `self._title` (which I added a check for in `finalize`, where `set_title` gets called). I also added two tiny tests to make sure users will be able to specify a `size` (a tuple of pixels) or a `title` (a string).


yellowbrick/text/tsne.py
```diff
@@ -177,6 +177,8 @@ def __init__(self, ax=None, decompose='svd', decompose_by=50, labels=None,
         self.colors = colors
         self.colormap = colormap
         self.random_state = random_state
+        self._size = kwargs.pop('size', None)
```
bbengfort: Note that there is a size `@property` that is part of the base class that uses this attribute; unfortunately, it isn't as simple as simply setting `_size` to what the user passes in, a computation has to be made: 

https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/base.py#L100

I made sort of a strange choice when advising the `size` property implementation - based on my matplotlib naivete and my familiarity with pixel sizes, I requested that size be specified in pixels instead inches as is usual in matplotlib. This has had some implications for our implementation and we should probably discuss whether or not we'd like to revise it.

yellowbrick/text/tsne.py
```diff
@@ -177,6 +177,8 @@ def __init__(self, ax=None, decompose='svd', decompose_by=50, labels=None,
...
+        self._title = kwargs.pop('title', None)
```
bbengfort: Does this work if you set `self.title = kwargs.pop('title', None)` instead? The `Visualizer.title` is not a `property` but it is an attribute that's used to determine the default vs. user title if specified. (See comment below).

bbengfort:
Ok, I'm sorry for any confusion I may have caused - I just realized that I had mistakenly thought that size and title were not being modified, but actually, it is a `TypeError` being raised. Popping the kwargs off is exactly the right strategy, but I think that's all that needs to be done, the base class will handle the rest. Can I propose the following:

```python
    def __init__(self, ax=None, decompose='svd', decompose_by=50, labels=None,
                 classes=None, colors=None, colormap=None, random_state=None, **kwargs):

        # Visual Parameters
        self.labels = labels
        self.colors = colors
        self.colormap = colormap
        self.random_state = random_state

        # Fetch TSNE kwargs from kwargs by popping only the keys that belong to TSNE params 
        tsne_kwargs = {
                key: kwargs.pop(key)
                for key in TSNE().get_params() 
                if key in kwargs 
        }
        self.transformer_ = self.make_transformer(decompose, decompose_by, tsne_kwargs)
        
        # Call super at the end so that size and title are set correctly
        super(TSNEVisualizer, self).__init__(ax=ax, **kwargs)
```

This will ensure that only the correct kwargs go to TSNE based on what is in `get_params` (and will evolve as scikit-learn evolves and will be correct no matter the scikit-learn version) and will allow the base visualizer to get the `size` and `title` params.

rebeccabilbro:
Thanks @ndanielsen and @bbengfort !

I've made the updates as suggested. Let me know if there are any other tests or additions you think we might need!

bbengfort:
This looks great to me, and thank you for the additional tests! I'll go ahead and get this merged in.

Issue #475, opened 2018-06-13
Add alpha support for scatter plots


mattharrison:
**Describe the solution you'd like**
When I'm plotting scatter plots (such as PCA), oftentimes there is a bunch of data. By using the ``alpha`` parameter on matplotlib (also available through the pandas plotting interface), it allows you to discern the data and amount of overlapping.

**Is your feature request related to a problem? Please describe.**
In general, I think an ``alpha`` parameter to any scatter plot visualizations (``PCADecomposition``) would be awesome. It would also be inline with what pandas plotting provides.

**Examples**
Seaborn scatter plots come with a slight alpha applied. But you can use ``scatter_kws={'alpha':0.3}`` to set it. In pandas or matplotlib you just pass ``alpha=0.3``.

Jake's book also has alpha turned on when plotting scatter plots. See https://jakevdp.github.io/PythonDataScienceHandbook/05.09-principal-component-analysis.html

rebeccabilbro:
@bbengfort oh that's an interesting alternative to deprecating ScatterVisualizer altogether in v0.8! While making it a mixin and refactoring the other plots to inherit the scatter-draw behavior might take a bit more work, I imagine it might make things a bit easier down the line as we introduce more types of visualizers. I'm game to help if we decide to go this way.

ndanielsen:
@bbengfort I think that the ScatterVisualizer makes the most sense as a base class. I agree that it is initially more work but would likely makes things simpler down the road

lwgray:
@bbengfort Nothing more to add...I agree with @rebeccabilbro and @ndanielsen

mattharrison:
For ``alpha=`` vs ``scatter_kws``, I think it depends on whether you want to tie the implementation to matplotlib. The former is generic, the latter indicates (to me at least) that these are the parameters that will be passed into matplotlib.

bbengfort:
Ok, I'll get started turning the `ScatterVisualizer` into a mixin - thanks for weighing in everyone!

rebeccabilbro:
I was doing a bit of experimentation (see [notebook here](https://github.com/rebeccabilbro/yellowbrick/blob/scatter-bug/examples/rebeccabilbro/scatter_alpha.ipynb)) with `ScatterVisualizer`, and it would certainly be easy enough to add an `alpha` param. But, given that we're thinking of turning `ScatterVisualizer` into a mixin, it might be best to just do it there. 

[...]

PR #554, opened 2018-08-09
Adds an alpha param to Scatterplot

peterespinosa:
#475 requests that an alpha parameter be added to scatterplots. This PR adds support for user-specified alpha values for scatterplots, which affects the translucency of the plotted points. I've also added a small test to ensure the user-provided alpha properly resets the param. _Note: I plan to follow up by adding `alpha` params to the other visualizers listed in #475 as needing them. This will be implemented in a subsequent PR._


yellowbrick/contrib/scatter.py
```diff
@@ -71,6 +72,10 @@ def scatterviz(X,
     markers : iterable of strings, default: ,+o*vhd
         Matplotlib style markers for points on the scatter plot points
 
+    alpha : float, default: 1.0
```
bbengfort: Thank you so much for updating the quick method with the new `alpha` parameter as well as the base visualizer! I know that there is a lot of code to update and it can be easy to miss, but if you wouldn't mind, please also update Line 85-86 and pass alpha to the `ScatterVisualizer` instantiation to ensure that alpha works as expected.

tests/test_contrib/test_scatter.py
```diff
@@ -147,6 +147,17 @@ def test_integrated_scatter(self):
...
+    def test_alpha_param(self):
+        """
+        Test that the user can supply an alpha param on instantiation
+        """
+        # Instantiate a scatter plot and provide a custom alpha
+        visualizer = ScatterVisualizer(
+            x="temperature", y="light", alpha=0.7
+        )
+
+        assert visualizer.alpha == 0.7
```
bbengfort: Awesome - I really appreciate it when PRs also include tests, it makes the review process tons easier and ensures we don't break things in the future!  

This test is sufficient for this PR, but because you mentioned you'd be interested in updating other visualizers with alpha parameters, I was wondering if you'd like to explore our [image similarity testing](http://www.scikit-yb.org/en/latest/contributing.html#image-comparison-tests)? This adds a bit more complexity to the development process but might be worth figuring out if you're interested in working on Yellowbrick more routinely. 

Alternatively for the other PRs, you could do something a bit simpler than image similarity, by mocking the `viz.ax.scatter` method and asserting that it is called with the supplied alpha. Let me know if you'd like me to go into more detail on this or an image similarity, I'd be happy to help!

yellowbrick/contrib/scatter.py
peterespinosa: whoops! just pushed a new commit; hopefully, that sorts it, but please double-check for me (I'm pretty new to Python!)

tests/test_contrib/test_scatter.py
peterespinosa: Thanks! I'm also new to testing. This is a great learning experience. I'm maybe not quite ready for the visual test, but I'll take a look at the link you sent!

bbengfort: Ok, no worries at all! The image similarity tests aren't necessarily difficult but do require some extra process to save baseline images. When we merge this in, I can quickly write a mock test as an example of what mocking might look like, in case you'd like to use that for the other PRs.

lwgray:
@peterespinosa great job on your first contribution!

PR #555, opened 2018-08-09
Minor updates to contrib ScatterViz tests

bbengfort:
This PR implements a mock test to ensure `alpha` is passed to the `ax.scatter` as well as cleans up a couple of other tests that I spotted while I was there. 

@peterespinosa I wanted to provide this for reference in case it helps with writing other tests without having to deal with image similarity! I've added some pointers as comments in the code.

### DIFF
--- a/tests/test_contrib/test_scatter.py
+++ b/tests/test_contrib/test_scatter.py
@@ -17,7 +17,6 @@ Test the ScatterViz feature analysis visualizers
 ##########################################################################
 
 import pytest
-import unittest
 import numpy as np
 import matplotlib as mptl
 
@@ -30,9 +29,15 @@ from tests.base import VisualTestCase
 from yellowbrick.exceptions import ImageComparisonFailure
 
 try:
-    import pandas
+    import pandas as pd
 except ImportError:
-    pandas = None
+    pd = None
+
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
 
 ##########################################################################
 # ScatterViz Base Tests
@@ -62,6 +67,9 @@ class ScatterVizTests(VisualTestCase, DatasetMixin):
         super(ScatterVizTests, self).tearDown()
 
     def test_init_alias(self):
+        """
+        Test alias for ScatterViz
+        """
         features = ["temperature", "relative_humidity"]
         visualizer = ScatterVisualizer(features=features, markers=['*'])
         self.assertIsNotNone(visualizer.markers)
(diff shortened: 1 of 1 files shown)


tests/test_contrib/test_scatter.py
```diff
@@ -30,9 +29,15 @@
...
+    import pandas as pd
 except ImportError:
-    pandas = None
+    pd = None
+
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
```
bbengfort: Importing the mock library in this fashion is to preserve python 2 compatibility; if we decide to drop python 2 support, then we can simply import from unittest.

tests/test_contrib/test_scatter.py
```diff
@@ -152,12 +160,21 @@ def test_alpha_param(self):
...
-            x="temperature", y="light", alpha=0.7
-        )
+        visualizer = ScatterVisualizer(alpha=0.7, features=["a", "b"])
 
+        # Test param gets set correctly
         assert visualizer.alpha == 0.7
 
+        # Mock ax and fit the visualizer
+        visualizer.ax = mock.MagicMock(autospec=True)
+        visualizer.fit(self.X[:, :2], self.y)
```
bbengfort: These two lines of code replace the `visualizer.ax` property, which was a `matplotlib.Axes` object, with a `mock.MagicMock` object. The `MagicMock` records all calls made to it and can return mock values or raise exceptions depending on how we want it tested. The `ax.scatter` method is called on `fit`, so we call fit here with some real data.

tests/test_contrib/test_scatter.py
```diff
@@ -152,12 +160,21 @@ def test_alpha_param(self):
...
+
+        # Test that alpha was passed to the scatter plot
+        _, scatter_kwargs = visualizer.ax.scatter.call_args
+        assert "alpha" in scatter_kwargs
+        assert scatter_kwargs["alpha"] == 0.7
```
bbengfort: Generally speaking, you can use the `assert_called_with` and `assert_called_once_with` functions to ensure that `ax.scatter` was called correctly. However, in this case you'd have to know all the arguments that were passed to the function. You could use `mock.ANY` so you don't necessarily have to know what value was passed, but you'd still have to know the arguments list. This method specifically checks for `alpha`, which is what we're testing in this function.

tests/test_contrib/test_scatter.py
lwgray: I think your use of mock here is quite clever and it sets the stage for testing if alpha is sent to the scatterplot

tests/test_contrib/test_scatter.py
rebeccabilbro: Nice - this is also something we should earmark for #558

tests/test_contrib/test_scatter.py
```diff
@@ -178,8 +195,7 @@ def test_scatter_quick_method(self):
         # test that is returns a matplotlib obj with axes
         self.assertIsInstance(ax, mptl.axes.Axes)
 
-    @unittest.skipUnless(pandas is not None,
-                         "Pandas is not installed, could not run test.")
+    @pytest.mark.skipif(pd is None, reason="pandas is required for this test")
```
rebeccabilbro: Thanks for catching this unittest=>pytest update!

Issue #558, opened 2018-08-10
Add alpha support for other scatter visualizers


rebeccabilbro:
**Describe the solution you'd like**
In follow up to @peterespinosa's [recent PR](https://github.com/DistrictDataLabs/yellowbrick/pull/554), we should add an `alpha` (translucency) param to the scatter visualizers that don't currently have one.

**Is your feature request related to a problem? Please describe.**
The following visualizers need an `alpha` parameter:

- PCADecomposition 
- PredictionError 
- ResidualsPlot (implements a default alpha) 
- text.TSNE (implements a default alpha) 

see discussion in #475

bbengfort:
NOTE: When this parameter is added; please also use `manual_legend` to correctly draw the legend without opacity. See #564 and #557

PR #564, opened 2018-08-19
Adds manual legend drawing utility

### LINKED ISSUE #557: Add a non alpha legend mechanism
Add a legend mechanism such that adjusting the `alpha` (translucency) param of a visualizer does not render the colors in the legend less readable.

For visualizers that accept an `alpha` param, the `alpha` doesn't just impact the translucency of the plotted data for visualizers, it *also* affects the translucency of the color labels in the plot's legend, making them hard to read. 

image

image

To resolve this, we should add a legend function similar to the one @bbengfort has implemented in [several places](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/features/manifold.py#L403), which adds color circles with a label to the legend.

```python 
from matplotlib import patches

def add_manual_legend(colors, labels):
    """
    Create a manual scatter plot legend by specifying a list of colors and labels. 
    """
    handles = [
        patches.Patch(color=color, label=label)
        for color, label in zip(colors, labels)
    ]
    self.ax.legend(handles=handles)
```

See discussion in #475

bbengfort:
Created a module, yellowbrick.draw that is intended to store utility
functions for common matplotlib drawing procedures (e.g. adding a manual
legend or creating a stacked bar chart).

Added manual_legend that can draw a legend with names and colors by
specifying rectangular patches as the artist for the legend. Although
this does not take into account the line style or marker style of the
plot, it does provide an easy way to specify color and label pairs as
the legend to avoid issues like too many labels, or opacity making the
legend unreadable.

TODO:

- [x] discuss if we want to add simple circles and lines to this mechanism 
- [x] create a list of visualizers we'd like to update to use this utility 
- [x] update current visualizers from above. 

Visualizers to Update:

- [x] Manifold 
- [x] RadViz 
- [x] ParallelCoordinates 
- [x] ResidualsPlot 

Fixes #557


.gitignore
```diff
@@ -52,11 +52,6 @@ coverage.xml
 # Sphinx documentation
 docs/_build/
 
-# PyBuilder
-target/
-!yellowbrick/target/
-!docs/api/target/
```
bbengfort: This was causing some weirdness with my git -- I don't know anyone who uses PyBuilder so I just nixed it from the .gitignore and the issues were resolved.

rebeccabilbro: Interesting - looks like this line has been in .gitignore since the beginning, just added as part of a generic list, so I'm totally good with cutting it. Out of curiosity, what IDE/editor were you using when the odd git stuff was happening? Just asking because I'm constantly switching between my home machine set-up (atom.io) and my work machine (pycharm), and always have trouble keeping track. I think it's usually IntelliJ weirdness for me though.

yellowbrick/draw.py
```diff
@@ -0,0 +1,82 @@
+# yellowbrick.draw
+# Utilities for common matplotlib drawing procedures.
```
rebeccabilbro: Love this idea of having a special module of drawing utilities! Now I'm wondering which other things we might want to go in here - maybe methods for annotating plots with the visualizer's scores or fit times, or other metadata like that?

tests/test_draw.py
```diff
@@ -0,0 +1,56 @@
...
+
+        _, ax = plt.subplots()
+        ax.scatter(Ax, Ay, c='r', alpha=0.35, label='a')
+        ax.scatter(Bx, By, c='g', alpha=0.35, label='b')
+        ax.scatter(Cx, Cy, c='b', alpha=0.35, label='c')
+
+        # Add the manual legend
+        manual_legend(
+            ax, ('a', 'b', 'c'), ('r', 'g', 'b'), frameon=True, loc='upper left'
+        )
```
rebeccabilbro: Great test and good to see these matplotlib kwargs work as expected!

rebeccabilbro:
One other quick thing; in addition to updating Manifold, RadViz, ParallelCoordinates, and ResidualsPlot to use the `manual_legend`, note that we've also got an open issue (#558) that means to add `alpha` params to PCADecomposition, PredictionError, ResidualsPlot, and TSNE. Once those have alphas, they'll probably need a `manual_legend` too, right?

tests/test_draw.py
bbengfort: And on Windows too, how'd that happen!?

yellowbrick/draw.py
```diff
@@ -0,0 +1,82 @@
...
+        Axes object. If None, the current axes are drawn on, but this is not
+        recommended.
+
+    labels : list of str
+        The text labels to associate with the legend. Note that the labels
+        will be added to the legend in the order specified.
+
+    colors : list of colors
+        A list of any valid matplotlib color reference. The number of colors
+        specified must be equal to the number of labels.
```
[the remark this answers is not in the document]
bbengfort: Ah yes, we definitely need that check - I meant to add that and had forgotten; coming up shortly.

tests/test_draw.py
rebeccabilbro: Well, still trying to configure the Surface loaner...

.gitignore
bbengfort: I'm using Atom -- though seriously considering switching over to VS code. 

PyBuilder is a continuous build tool that is mostly directed towards applications; so I don't think we have to worry about it for library development. And yes, it was part of the default GitHub python .gitignore file that we just added by default.

yellowbrick/draw.py
bbengfort: Sure, my feeling is that anywhere in our code where we do the same n number of matplotlib steps, might be a good candidate for a drawing utility.

rebeccabilbro:
Thanks for adding the color length check and updating `Manifold`! Did you want to make the rest of the visualizer updates part of a separate PR (in which case I'll merge in now!) or do them in this one?

bbengfort:
@rebeccabilbro I updated Parallel Coordinates, RadViz, and ResidualsPlot to use manual legend just now and commented on #558 that we should add manual_legend. If the tests pass and you're satisfied with my updates to those visualizers, please feel free to merge; otherwise, let me know any changes that need to be made.

rebeccabilbro:
Excellent, thank you so much for following up and updating the rest of those visualizers (and all the baseline images, etc)! This PR feels quite complete to me, so I'll open up a new issue for the `manual_legend` upgrade, which I think can be addressed at a later time. Thanks again for your contribution!

PR #567, opened 2018-08-20
Fixes #491 unicode-decode error on Windows

### LINKED ISSUE #491: Hobbies corpus unicode decoding error on Windows x64 (AppVeyor)
**Describe the bug**

There appears to be a unicode decoding error on Python 3 in Windows when loading the hobbies corpus. This was observed in AppVeyor builds that use the hobbies corpus during testing such as #489 but is probably true for any Windows machine. 

**To Reproduce**

```python

datasets = DatasetMixin() 
hobbies = datasets.load_data('hobbies')
```

**Expected behavior**

We should either remove the unicode encoding from the corpus dataset (unless we beleive that this is a positive feature for the tests ... though currently the encoding of the data does not matter to visualizer performance) or we should fix the datasets to correctly decode the data from disk. 

**Traceback**

```
tests\dataset.py:197: in load_data
    return DatasetMixin.load_corpus(name, fixtures)
tests\dataset.py:230: in load_corpus
    data.append(f.read())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
self = <encodings.cp1252.IncrementalDecoder object at 0x0000001C91A0D828>
input = b'The Lonely City\xc2\xa0bristles with heart-piercing wisdom. Loneliness, according to Laing, feels \xe2\x80\x9clike b...an read The\xc2\xa0Rumpus\xe2\x80\x99s review of\xc2\xa0The Lonely City\xc2\xa0right\xc2\xa0here.)\n\nRelated Posts:\n'
final = True
    def decode(self, input, final=False):
>       return codecs.charmap_decode(input,self.errors,decoding_table)[0]
E       UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 164: character maps to <undefined>
C:\Python36-x64\lib\encodings\cp1252.py:23: UnicodeDecodeError
```

 - OS: `AppVeyor Winx64`
 - Python Version `PYTHON=C:\Python36-x64, PYTHON_VERSION=3.6.4, PYTHON_ARCH=64`
 - Yellowbrick Version `develop`

**Additional context**

See [AppVeyor Build 0.7.300](https://ci.appveyor.com/project/districtdatalabs/yellowbrick/build/0.7.300/job/wy9k5x9linoc6tf8) - note that tests _passed_ in `
Environment: PYTHON=C:\Python27-x64, PYTHON_VERSION=2.7.14, PYTHON_ARCH=64`

drwaterman:
Fixes #491 Unicode decode error on Windows/Python 2


bbengfort:
@drwaterman, when you get the chance, here are the fixes I propose:

1. Undo the changes made in 12fb863 by setting the encoding in `datasets.py` back to `"UTF-8"` and reverting the baseline image. (It's fine to do this in another commit rather than trying to revert or remove the last commit). 
2. In [tests/test_test/test_dispersion on Line 74](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_text/test_dispersion.py#L74) change `text = chain(*map(str.split, corpus.data))` to `text = chain(*map(lambda s: s.split(), corpus.data))`

This will hopefully allow the tests to pass everywhere!

drwaterman:
@bbengfort After making the suggested changes everything seems to check out OK. I still have the strange numpy error we ran into yesterday and per @rebeccabilbro documented it in #572.

drwaterman:
@bbengfort If you are interested in reducing the RMSE for the Windows tests, I did a little playing around and I think if you turn off text antialiasing you can get rid of some of the text shadowing that we saw in the windows diff files. I added the following to a couple of the tests:

At imports: `import yellowbrick.style.rcmod as yb_rcmod`
At the beginning of a testing method: `yb_rcmod.set_aesthetic(rc={'text.antialiased': False})`

Then comparing the output images from a windows machine and a mac machine, most of the text shadows disappear.

There may be further rcmod parameters that can do similar things for the other rendered elements, I only tested with the text.

The reduction in RMSE wasn't huge so you might not want to spend the time to implement this change unless a fix can also be found for the other elements.

Issue #458, opened 2018-05-29
Add legend option to PCA visualizer 


mattharrison:
The ``PCADecomposition`` allows us to set the color of the different entries. It would be cool to add a legend to describe the coloring.

### Proposal/Issue

The ``DataVisualizer`` class supports creating legends, but it requires that ``y`` is passed into ``.fit``. Admittedly, PCA doesn't require labels, but sometimes you have them. I would love to label the colorings in PCA. Could the constructor accept a ``legend`` parameter that is similar to the ``color`` parameter, but accepts labels?

bbengfort:
I think this is a great idea, and we can absolutely do this the same way that we do in the `DataVisualizer`, with the `y` attribute. Probably this can be done similarly to the manifold visualizers, which allow both heatmap colors for regression and discrete colors for classification and coloring.

rebeccabilbro:
Note, we now have a `manual_legend` utility in `yellowbrick.draw`, which could be useful here. See #564 for more discussion.

Naba7:
Can I be assigned this issue?

rebeccabilbro:
Hello @Naba7 &mdash; as explained in response to your notes on #738, #677, and #476, we do not "assign" issues or reserve issues for contributors. Anyone is welcome to submit a PR for a feature or bugfix they work on. 

However, given that you already have one PR open that still needs to be completed (#755), have started working on #615, and are new to working on Yellowbrick and still getting to know our API, we would *really* appreciate if you would focus on getting those first PRs across the finish line before starting anything new. 

We appreciate your enthusiasm about contributing to Yellowbrick. One of the most important lessons to learn is that open source is a marathon, not a sprint, so we hope you can be patient and enjoy the journey &mdash; we promise Yellowbrick isn't going away!

bbengfort:
Feature has been added in #874

PR #569, opened 2018-08-21
[WIP - On hold] Adds manual legend to PCADecomposition (addresses #458)

pdamodaran:
Added legend_labels parameter to show the classes corresponding to the list of colors passed to the `PCADecomposition` visualizer.  Used the new manual_legend feature under draw.py.


rebeccabilbro:
@pdamodaran - thanks for taking this one on! It looks like you might be missing the `from yellowbrick.draw import manual_legend` import statement at the top of `pca.py`, which is triggering the test failures.

yellowbrick/features/pca.py
```diff
@@ -207,6 +214,10 @@ def finalize(self, **kwargs):
         self.ax.set_ylabel('Principal Component 2')
         if self.proj_dim == 3:
             self.ax.set_zlabel('Principal Component 3')
+        if self.legend_labels is not None:
+              manual_legend(
+              ax, self.legend_labels, self.color, frameon=True, loc='best'
```
rebeccabilbro: Looks like this is where the tests are failing; this should be `self.ax` instead of `ax`, right?

rebeccabilbro:
Thanks for making that fix @pdamodaran! The automated tests are running now, but in the meantime, would you do the following:

```
cd tests
pip install -r requirements.txt
cd test_features
pytest test_pca.py
```

... and see what happens?

yellowbrick/features/pca.py
```diff
@@ -24,6 +24,7 @@
 from yellowbrick.features.base import MultiFeatureVisualizer
 from yellowbrick.style import palettes
 from yellowbrick.exceptions import YellowbrickValueError
+from yellowbrick.draw import manual_legend
```
rebeccabilbro: Thanks for using the new `manual_legend` utility here!

yellowbrick/features/pca.py
```diff
@@ -64,6 +65,9 @@ class PCADecomposition(MultiFeatureVisualizer):
 
     color : list or tuple of colors, default: None
         Specify the colors for each individual class.
+        
+    legend_labels : list or tuple of values for an optional legend that lists
```
rebeccabilbro: I'm thinking it would be best to use `classes` here instead of `legend_labels`; in addition to not introducing new params into YB, this would also be more analogous to what we do in the `DataVisualizers` (though I'm okay with keeping this as `MultiFeatureVisualizer`)

yellowbrick/features/pca.py
```diff
@@ -98,6 +102,8 @@ def __init__(self,
                  proj_dim=2,
                  proj_features=False,
                  color=None,
+                 legend_labels=None,
```
rebeccabilbro: Let's change this to `classes=None`

yellowbrick/features/pca.py
```diff
@@ -98,6 +102,8 @@ def __init__(self,
...
+                 label=None,
```
rebeccabilbro: I looked, but I can't see where this `label` param gets used -- can we remove this or did I miss something?

yellowbrick/features/pca.py
```diff
@@ -121,6 +127,7 @@ def __init__(self,
 
         # Visual Parameters
         self.color = color
+        self.legend_labels = legend_labels
```
rebeccabilbro: Let's change to `self.classes_ = classes`. Since this visualizer inherits from `MultiFeatureVisualizer` rather than `DataVisualizer`, we'll need to explicitly set this (whereas with a `DataVisualizer`, `self.classes_` would be automatically created if the user passed in something for the `classes` param)

yellowbrick/features/pca.py
```diff
@@ -207,6 +214,10 @@ def finalize(self, **kwargs):
         self.ax.set_ylabel('Principal Component 2')
         if self.proj_dim == 3:
             self.ax.set_zlabel('Principal Component 3')
+        if self.legend_labels is not None:
```
rebeccabilbro: To keep consistent with the other changes, let's change this to a check to see if `self.classes_` exists, and if so, add a manual legend.

yellowbrick/features/pca.py
```diff
@@ -207,6 +214,10 @@ def finalize(self, **kwargs):
...
+              manual_legend(
+              self.ax, self.legend_labels, self.color, frameon=True, loc='best'
```
rebeccabilbro: We can change `self.legend_label` to `self.classes_` here. We should also be careful that `self.color` will work as expected -- have you experimented with a few different inputs for the color/colormap params? It might be easiest to explore in a Jupyter notebook. 

Also, do we actually need to pass in `self.ax`?

yellowbrick/features/pca.py
```diff
@@ -215,7 +226,7 @@ def finalize(self, **kwargs):
 
 def pca_decomposition(X, y=None, ax=None, features=None, scale=True,
                       proj_dim=2, proj_features=False, color=None,
-                      colormap=palettes.DEFAULT_SEQUENCE,
+                      legend_labels=None,colormap=palettes.DEFAULT_SEQUENCE,
```
rebeccabilbro: Again, let's change to `classes=None` here for consistency with other similar visualizers.

yellowbrick/features/pca.py
```diff
@@ -251,6 +262,9 @@ def pca_decomposition(X, y=None, ax=None, features=None, scale=True,
 
     color : list or tuple of colors, default: None
         Specify the colors for each individual class.
+        
+    legend_labels : list or tuple of values for an optional legend that lists
```
rebeccabilbro: Please replace with: 
```
    classes : list, default: None
        a list of class names for the legend
        If classes is None and a y value is passed to fit then the classes
        are selected from the target vector.
```

yellowbrick/features/pca.py
```diff
@@ -279,8 +293,8 @@ def pca_decomposition(X, y=None, ax=None, features=None, scale=True,
     # Instantiate the visualizer
     visualizer = PCADecomposition(
         ax=ax, features=features, scale=scale, proj_dim=proj_dim,
-        proj_features=proj_features, color=color, colormap=colormap,
-        random_state=random_state, **kwargs
+        proj_features=proj_features, color=color, legend_labels=legend_labels,
```
rebeccabilbro: Please change `legend_labels=legend_labels` to `classes=classes`

yellowbrick/features/pca.py
pdamodaran: you are correct @rebeccabilbro - I don't see it being used and it is not defined in the parameter list.  Shall I remove it?

pdamodaran:
I have begun implementing the changes you suggested @rebeccabilbro.   I will get in touch with you if I have any questions.  Thanks for your feedback!

pdamodaran:
@rebeccabilbro I made the changes you suggested.  I took out label from the parameter list.  I changed self.ax to self.  

Regarding the color checks, I was thinking about going the Manifold route.  What I was wondering though is if the utility class "_determine_target_color_type" in the Manifold visualizer can be moved to something like the utils package so that it can be used by multiple visualizers.  Let me know your thoughts regarding this.

pdamodaran:
@rebeccabilbro - the legend only gets drawn if the classes parameter was passed to the visualizer.  Here is the code that does this:

if self.classes_ is not None:
              manual_legend(
              self, self.classes_, self.color, frameon=True, loc='best'

Do you suggest a different way to infer the classes and apply the legend?

Also, I will go ahead and create a new issue for moving the _determine_target_color_type to the utils package.

rebeccabilbro:
Gotcha @pdamodaran - would you please push a notebook to this branch inside the `yellowbrick/examples/pdamodaran` folder that demonstrates how this prototype is working on your machine? I have been trying various versions where I try to pass `classes` into `PCADecomposition` and still can't seem to make your legends show up :/

pdamodaran:
@rebeccabilbro i have added a pca example notebook

rebeccabilbro:
Ah, thanks @pdamodaran, I see what the issue is now! Below is a bit more detail about how I and our users would probably expect to use the visualizer; I hope this helps clarify how the legend implementation should work, but if not, please let me know!

# PCA Classification Scenario

I have a classification dataset with features (`X`) and a target/outcome (`y`). My goal is to predict the target/outcome given the features. Sometimes there is a clear subset of those features that are very predictive of the target, but not in my case, so I am going to do a dimensionality reduction to compress the current features into a smaller feature space that extracts and captures the axes of principal variation -- this means I'm hoping I can convert my original `X` into a `pca_X` that is equally if not more predictive of `y` than `X`, and hopefully reduces some of the noise from `X`. 

In order to determine if this strategy is viable, I use the Yellowbrick `PCADecomposition` visualizer with my training data to see if my classes (e.g. the different possible values of the target `y`) are grouped together after I do a pca decomposition/projection. 

In this case, here's what I would expect to do:

```
iris = datasets.load_iris()
X = iris.data
y = iris.target
```

Since I have classes and know the labels, I'll pass those in as a list:

```
classes = ["setosa", "versicolour", "virginica"]
oz = PCADecomposition(classes=classes) 
oz.fit_transform(X, y)
oz.poof()
```

Let's say I have classes but not labels, I would expect to be able to do something like:

```
oz = PCADecomposition() 
oz.fit_transform(X, y) 
oz.poof()
```
# Since I've passed in a 'y', `PCADecomposition` can extrapolate at least the

rebeccabilbro:
@pdamodaran - sorry for the accidental and premature close! Reopened so that we can finish this up!

See above comment for a bit of usage context that I hope will help give this visualizer a bit more of a natural feel for users who are using it for a classification problem (as described in #458). Minimally, we should decide if the `classes` param expects a tuple or a list; currently the docstring says it expects a list, but your notebook is passing in a tuple. I think a list is probably a bit more natural and more closely resembles what we do in, for example, `ClassificationReport` and `ConfusionMatrix`. 

The additional advice above provides some suggestions about how to think about the usage for `determine_target_color_type`. Please let me know if you have additional questions or if you'd like to jump on a Hangouts call  -- happy to discuss further and thanks for sticking with this!

rebeccabilbro:
Hi there @pdamodaran - given how much the scikit-yb codebase has changed since this PR started, I'm thinking it might be best to close this one out and have you open a fresh one when you're ready. Hope that's ok!!

pdamodaran:
@rebeccabilbro - that's fine - will do!

PR #574, opened 2018-08-21
Adds alpha params to other scatterplot-based visualizers

rebeccabilbro:
This PR attempts to address #558, adding `alpha` (opacity/translucency) params to the rest of the scatterplot-based visualizers.

Done:
 - adds `alpha` to `PredictionError` and test coverage to check the param is passed to the internal pyplot `scatter`
 - adds `alpha` to `ResidualsPlot` and minimal test coverage
 - adds `alpha` to `text.TSNE` and test coverage to check the param is passed to the internal pyplot `scatter`

Still to do:
- [x]  create followup issue to add `alpha` to `PCADecomposition`  (note: was waiting to do this until #569 is done, so as not to collide with @pdamodaran) and appropriate test coverage
- [x] in the `test_alpha_param` for `ResidualsPlot`, figure out how to mock the axes to check the param is passed to the internal pyplot `scatter`. My first attempts triggered the Matplotlib ValueError "Axes instance argument was not found in a figure", which I think was caused by the residuals plot having multiple axes due to the new histogram feature, but I'm not entirely sure.


yellowbrick/regressor/residuals.py
```diff
@@ -113,7 +117,7 @@ class PredictionError(RegressionScoreVisualizer):
     """
 
     def __init__(self, model, ax=None, shared_limits=True,
-                 bestfit=True, identity=True, **kwargs):
+                 bestfit=True, identity=True, alpha=1.0, **kwargs):
```
bbengfort: If possible, I'd like to keep the default `alpha=0.75` like we had originally; I believe the alpha in the residuals plots is important to be able to quickly identify clusters of similar/duplicate errors and predictions.

yellowbrick/regressor/residuals.py
```diff
@@ -240,8 +250,9 @@ def finalize(self, **kwargs):
         self.ax.legend(loc='best', frameon=True)
```
bbengfort: If we give the user the ability to change the alpha, do we want to use `manual_legend` here?

yellowbrick/regressor/residuals.py
```diff
@@ -504,25 +526,27 @@ def draw(self, y_pred, residuals, train=False, **kwargs):
 
         if train:
             color = self.colors['train_point']
-            alpha = 0.5
             label = "Train $R^2 = {:0.3f}$".format(self.train_score_)
         else:
             color = self.colors['test_point']
-            alpha = 0.9
```
bbengfort: I'm not sure what to recommend here, but I do believe it's important to ensure that train/test have different alpha values - there are usually far more train points than test points, and without the increase in opacity it is hard to see the test points. Perhaps we could do something similar to what you did for `colors` - have `alphas["train"]` and `alphas["test"]`?

yellowbrick/text/tsne.py
```diff
@@ -35,7 +35,7 @@
 ##########################################################################
 
 def tsne(X, y=None, ax=None, decompose='svd', decompose_by=50, classes=None,
-           colors=None, colormap=None, **kwargs):
+           colors=None, colormap=None, alpha=1.0, **kwargs):
```
bbengfort: If possible, I'd also like to maintain the default `alpha=0.7` that we had before as it makes it easier to identify clusters with lots of points that are very close together in the TSNE projection.

yellowbrick/text/tsne.py
```diff
@@ -335,7 +345,7 @@ def draw(self, points, target=None, **kwargs):
         for label, points in series.items():
             self.ax.scatter(
                 points['x'], points['y'], c=colors[label],
-                alpha=0.7, label=label
+                alpha=self.alpha, label=label
             )
 
     def finalize(self, **kwargs):
```
bbengfort: If we're allowing the opacity to change, do we also want to use `manual_legend` here?

yellowbrick/regressor/residuals.py
rebeccabilbro: Yep, no problem

yellowbrick/regressor/residuals.py
rebeccabilbro: Ideally yes --  unfortunately though, I think putting a `manual_legend` in `finalize()` will overwrite the legend created for the best-fit and 45 degree lines, which are currently plotted and created in the `draw()` method (since best-fit at least depends on knowing `y` and `y_pred`). Another slightly odd consequence is that instead of seeing the label for R squared as a circle, it will be a bar:

(no manual legend)
image

(with manual legend)
image
   
I'm inclined to address this as a separate issue, but let me know if you have strong feelings.

yellowbrick/regressor/residuals.py
rebeccabilbro: True, though it might not be *quite* as bad as you might think. Here are some samples with different values of alpha (applied equally to a train of 80% and a test of 20%):

(alpha = 0.9)
image

(alpha = 0.7)
image

(alpha = 0.5)
image

(alpha = 0.3)
image

yellowbrick/text/tsne.py
rebeccabilbro: Good point! Done!

yellowbrick/text/tsne.py
rebeccabilbro: Another great point! I have approached this by changing `color_values` (which is created in `draw` using the `resolve_colors` method) into `self.color_values_` so that it can be used by `manual_legend` in `finalize`.

yellowbrick/regressor/residuals.py
rebeccabilbro: Ok, will do!

yellowbrick/regressor/residuals.py
rebeccabilbro: You got it!

yellowbrick/regressor/residuals.py
bbengfort: Did this change get made? Am I seeing an older version of the code?

yellowbrick/regressor/residuals.py
bbengfort: Good comment!

yellowbrick/regressor/residuals.py
```diff
@@ -354,6 +370,10 @@ class ResidualsPlot(RegressionScoreVisualizer):
     line_color : color, default: dark grey
         Defines the color of the zero error line, can be any matplotlib color.
 
+    alpha : float, default: 0.75
```
bbengfort: I think this is a good compromise for a default alpha between 0.5 and 0.9 (the original train and test alphas, respectively). Good choice!

yellowbrick/regressor/residuals.py
rebeccabilbro: Oops, I thought I had, but have now!

tests/test_regressor/test_residuals.py
```diff
@@ -323,3 +351,19 @@ def test_score(self):
...
+        visualizer = ResidualsPlot(
+            Ridge(random_state=8893), alpha=0.3
+        )
+
+        # Test param gets set correctly
+        assert visualizer.alpha == 0.3
+
+        # TODO: mock ax and test alpha is passed to pyplot scatterplot
+        # not sure how to do this without triggering the Matplotlib error:
+        # "ValueError: Axes instance argument was not found in a figure"
```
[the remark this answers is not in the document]
rebeccabilbro: Unfortunately setting `hist=False` does not correct the error. Here are the test results:

```
    def test_alpha_param(self):
        """
            Test that the user can supply an alpha param on instantiation
            """
        # Instantiate a prediction error plot, provide custom alpha
        visualizer = ResidualsPlot(
            Ridge(random_state=8893), alpha=0.3, hist=False
        )
    
        # Test param gets set correctly
        assert visualizer.alpha == 0.3
    
        # TODO: mock ax and test alpha is passed to pyplot scatterplot
        # not sure how to do this without triggering the Matplotlib error:
        # "ValueError: Axes instance argument was not found in a figure"
        # Mock ax and fit the visualizer
        visualizer.ax = mock.MagicMock(autospec=True)
>       visualizer.fit(self.data.X.train, self.data.y.train)

test_residuals.py:372: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../yellowbrick/regressor/residuals.py:463: in fit
    self.score(X, y, train=True)
../../yellowbrick/regressor/residuals.py:498: in score
    self.draw(y_pred, scores, train=train)
../../yellowbrick/regressor/residuals.py:553: in draw
    plt.sca(self.ax)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

ax = <MagicMock id='4632511600'>

    def sca(ax):
        """
        Set the current Axes instance to *ax*.
    
        The current Figure is updated to the parent of *ax*.
        """
        managers = _pylab_helpers.Gcf.get_all_fig_managers()
        for m in managers:
            if ax in m.canvas.figure.axes:
                _pylab_helpers.Gcf.set_active(m)
                m.canvas.figure.sca(ax)
                return
>       raise ValueError("Axes instance argument was not found in a figure.")
E       ValueError: Axes instance argument was not found in a figure.

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/matplotlib/pyplot.py:963: ValueError

```

tests/test_text/test_tsne.py
```diff
@@ -81,7 +85,7 @@ def test_integrated_tsne(self):
         tsne = TSNEVisualizer(random_state=8392, colormap='Set1')
         tsne.fit_transform(docs, labels)
 
-        tol = 40 if six.PY3 else 55
+        tol = 50 if six.PY3 else 55
```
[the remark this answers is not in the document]
rebeccabilbro: That seems to work at least on my local machine when I run the tests; we'll see what happens in the VM!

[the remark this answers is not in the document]
rebeccabilbro: Looks like Travis doesn't like `tol=40`; with the alpha manually set to 1.0 for this test case, I'm getting `ImageComparisonFailure: images not close (RMS 44.931)` for the integrated tsne test :(

tests/test_regressor/test_residuals.py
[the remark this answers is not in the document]
bbengfort: Ah, that traceback helps a lot. So in `draw()` and possibly in `finalize()`I call `plt.sca` (set current axes) to make sure that `viz.ax` is the global axes instead of `viz.hax` -- I guess that function doesn't like getting a mock! Can you patch the sca function? 

```python
@mock.patch('yellowbrick.regressor.residuals.plt.sca', autospec=True) 
def test_alpha_param(self):
    ... 
```

Patching is a bit of a trial and error thing, especially since the order of imports matters. If that doesn't work, you could also try `@mock.patch('matplotlib.pyplot.sca')` ... and if that doesn't work, then we can try other things.

tests/test_text/test_tsne.py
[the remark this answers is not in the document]
bbengfort: Ah well, 50 it is I guess.

tests/test_regressor/test_residuals.py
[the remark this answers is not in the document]
bbengfort: Oh sorry, this is actually a fairly simple fix*:

```python
@mock.patch('yellowbrick.regressor.residuals.plt.sca', autospec=True) 
def test_alpha_param(self, mock_sca):
    ... 
```

I forgot the second argument to the test function that is passed in by the decorator. 

* though the unittest dependency might screw us up ... 

However, if you want to simply mark this as `xfail` -- that's fine, we can fix it later.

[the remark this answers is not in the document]
rebeccabilbro: Ok, making one last attempt; the test passes on my machine with python 3.6.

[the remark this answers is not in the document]
bbengfort: 🤞

PR #575, opened 2018-08-21
Fix link to Baleen RSS Corpus

Kautumn06:
The link to the Baleen RSS Corpus was broken, so I replaced it with a link its [GitHub repository](https://github.com/DistrictDataLabs/baleen).


rebeccabilbro:
I've just updated your branch; once the tests pass, I'll go ahead and merge in. Thanks again @Kautumn06!

Issue #549, opened 2018-08-07
Color document boundaries (or classes) in dispersion plot


rebeccabilbro:
**Describe the solution you'd like**
I work with text data a lot and am very excited about the new [`DispersionPlots`](http://www.scikit-yb.org/en/latest/api/text/dispersion.html)! As a future feature, it would be very useful to indicate document boundaries on the plot using a high perceptual contrast colormap, or perhaps a qualitative colormap to color the classes (when classes are available).

**Use case**
Adding document boundaries (or per-class document coloring) to the plot would enable me to visualize things like:

 - n-grams that are very common in only a particular subset of the documents
 - n-grams that are very common across all documents
 - n-gram appearance vs. variance in document length
 - n-gram appearance on a per-class basis (if classes are available)

bbengfort:
Closed by #576

lwgray:
Issue #576 Doesn't completely resolve this issue thus it will remain open

Issue #488, opened 2018-06-22
Investigate Usage of sklearn VectorizerMixin with DispersionPlot Visualizer


lwgray:
**Describe the solution you'd like**
It was proposed that we investigate the sklearn.feature_extraction.text. VectorizerMixin. This will make this DispersionPlot visualizer more like a Scikit-Learn text transformer, and will allow us to submit input as:

- list of words
- raw string that is tokenized using sklearn
- list of strings representing documents on disk

More information on the Mixin can be found here:
https://github.com/scikit-learn/scikit-learn/blob/a24c8b46/sklearn/feature_extraction/text.py#L98

lwgray:
@rebeccabilbro Can you help me remember what this issue was about, I created it at your behest.

PR #576, opened 2018-08-21
If applied this commit denotes document boundaries on Dispersion Plot viz

lwgray:
This is a partial fix to issue #549 for which I am attempting to define document boundaries within a dispersion plot.

I had to make some minor changes to make this work.  The visualizer was changed so that the expected input is now a list of documents.  Before the input was a single list representing all the words of the corpora.  This change makes it easier to define the beginning/end of a document and keep track of the word offset.

PLOTS
---------
This plot is of the occurrence of these words in documents from the 'sporting' subset of the hobbies dataset.  The light grey lines denote the document boundaries.

image

Coding (greatly appreciated)
-----
* my variable naming can be improved, any suggestions??? 

Visuals
-------
* Suggestions on alpha, linecolor, linestyle, and linewidth


yellowbrick/text/dispersion.py
```diff
@@ -47,20 +47,24 @@ class DispersionPlot(TextVisualizer):
...
-    def __init__(self, words, ax=None, color=None, ignore_case=False, **kwargs):
+    def __init__(self, words, ax=None, color=None, ignore_case=False,
+                 boundary=False, **kwargs):
         super(DispersionPlot, self).__init__(ax=ax, **kwargs)
         
         self.color = color
         self.words = words
         self.ignore_case = ignore_case
-        
+        self.set_boundaries = boundary
```
bbengfort: I think it's important that the properties on the visualizer match the variables that are passed in the arg list; Scikit-Learn using metaprogramming to detect what's in the `__init__` args for `get_params` and `set_params`, which we also make use of.

yellowbrick/text/dispersion.py
```diff
@@ -47,20 +47,24 @@ class DispersionPlot(TextVisualizer):
     ignore_case : boolean, default: False
 	Specify whether input  will be case-sensitive.
         
+    boundary : boolean, default: False
```
bbengfort: I guess I don't mind `boundaries` or even just `annotate_docs`? It might be important to specify whether the end or the beginning of the document is being annotated in the comment.

yellowbrick/text/dispersion.py
```diff
@@ -84,13 +88,25 @@ def fit(self, text):
             A list of words in the order they appear in the corpus.
         """
             
+        # Concatenating documents and define boundaries by word offset
+        self.boundaries_ = []
+        self.doc_words_ = []
+        self.offset = 0
+
+        for doc in text:
+            for word in doc.split():
```
bbengfort: I think it's reasonable to expect as input a list of lists; e.g. a list of documents where each document is a list of words as opposed to a list of strings. This is closer to what scikit-learn does in `sklearn.feature_extraction.text`, which we'll want to use eventually in this module, e.g. #488 and will also allow the user to use `nltk.word_tokenize` or `nltk.bigrams` if required. 

Therefore I recommend changing this to `for word in doc:` rather than simply splitting on space.

yellowbrick/text/dispersion.py
```diff
@@ -84,13 +88,25 @@ def fit(self, text):
             A list of words in the order they appear in the corpus.
         """
             
+        # Concatenating documents and define boundaries by word offset
+        self.boundaries_ = []
+        self.doc_words_ = []
+        self.offset = 0
```
bbengfort: We don't really want to store the entire corpus in memory on the visualizer, it's possible that it's an extremely big corpus that's being stored on disk and a generator is reading through all the documents; which is why we had the generator test to ensure that we only looped through the input once (this loops through twice). 

Is it possible to compute the boundaries and the points in one single loop?

yellowbrick/text/dispersion.py
```diff
@@ -103,6 +119,11 @@ def draw(self, points, **kwargs):
         kwargs: generic keyword arguments.
         """
         
+        # Define boundaries with a solid vertical line
+        if self.set_boundaries:
+            for xcoords in self.boundaries_:
+                self.ax.axvline(x=xcoords, color='lightgray', alpha=0.75,
```
bbengfort: I like that you're trying to mute the document boundaries so there is a contrast between the words and the annotations - very nice! Perhaps setting `linestyle='--'` or `':'` would also help?

yellowbrick/text/dispersion.py
lwgray: I have decided to go with annotate_docs

yellowbrick/text/dispersion.py
```diff
@@ -31,69 +31,80 @@ class DispersionPlot(TextVisualizer):
...
-	    # NOTE: this will find all indices if duplicate words are supplied 
-	    # In the case that word is not in target words, any empty list is
-	    # returned and no data will be yielded 
-            for y in (self.target_words_ == word).nonzero()[0]:
-                yield (x, y)
+    def _compute_dispersion(self, text):
+        self.boundaries_ = []
+        self.offset = 0
+        for doc in text:
+            for word in doc:
```
lwgray: here, I define boundaries and points in a single loop

lwgray:
I will add a test for boundaries in a follow-up commit

tests/test_text/test_dispersion.py
```diff
@@ -67,12 +67,27 @@ def test_dispersionplot_generator_input(self):
         """
         corpus = self.load_data('hobbies')
 
-        text = chain(*map(lambda s: s.split(), corpus.data))
+        text = (doc.split() for doc in corpus.data)
```
bbengfort: We do still need the `map` here to ensure a generator is passed in; this line of code is creating a tuple in memory. Would you please change it to:

```python
text = map(lambda doc: doc.split() for doc in corpus.data)
```

tests/test_text/test_dispersion.py
```diff
@@ -67,12 +67,27 @@ def test_dispersionplot_generator_input(self):
...
+        corpus = self.load_data('hobbies')
+
+        text = (doc.split() for doc in corpus.data)
+        target_words = ['girl', 'she', 'boy', 'he', 'man']
+
+        visualizer = DispersionPlot(target_words, annotate_docs=True)
+        visualizer.fit(text)
+        visualizer.ax.grid(False)
+
+        self.assert_images_similar(visualizer, tol=25)
```
bbengfort: Good test, thank you for adding it!

tests/test_text/test_dispersion.py
```diff
@@ -21,7 +21,7 @@
 from yellowbrick.text.dispersion import *
 from tests.dataset import DatasetMixin
 from tests.base import VisualTestCase
-from itertools import chain
+# from itertools import chain
```
bbengfort: You can go ahead and delete this -- we won't need it with the new input.

yellowbrick/text/dispersion.py
```diff
@@ -31,69 +31,82 @@ class DispersionPlot(TextVisualizer):
...
         """
         The fit method is the primary drawing input for the dispersion
         visualization. It requires the corpus as a list of words.
-        
+
         Parameters
         ----------
         text : list
-            A list of words in the order they appear in the corpus.
+            Should be provided as a list of documents that contain
```
bbengfort: Thanks for updating the docstring!

tests/test_text/test_dispersion.py
bbengfort: Yes, you're right - sorry; I mistook the parens for a tuple comprehension, but that does indeed wrap a generator expression. Might be a good comment, but you're right, go ahead and merge this in!

https://stackoverflow.com/questions/16940293/why-is-there-no-tuple-comprehension-in-python

PR #578, opened 2018-08-21
Adding interactive classifier example

jlinGG:
This pull request adds a jupyter notebook example illustrating the use of ipywidgets to create an interactive model selection demo.

### DIFF
--- a/examples/jlinGG/Interactive_YB_classifiers.ipynb
+++ b/examples/jlinGG/Interactive_YB_classifiers.ipynb
@@ -1,0 +1,179 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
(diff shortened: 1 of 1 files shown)


jlinGG:
I'd be totally down! Please let me know how I can be helpful in the process.

rebeccabilbro:
@jlinGG awesome! Would you mind opening up a new issue that asks for interactive visualizer demos (using YB and ipywidgets) viewable with Binder? Then if you're game, you and @wagner2010  can do a bit of research to see how other folks are using Binder to show demos. Currently our README has a Binder button, which when clicked launches a Docker with our `examples.ipynb` notebook viewable (which as I understand it is nice because users can interact with it without having to clone YB and run the notebooks locally). Are there other ways of using Binder, maybe to have a sort of gallery of notebooks?

jlinGG:
New issue can be seen here: https://github.com/DistrictDataLabs/yellowbrick/issues/585

PR #580, opened 2018-08-22
[MRG] Clear Older Plot

### LINKED ISSUE #579: Visualizer does not forget previous plot.
**Describe the bug**
When making consecutive plots in python script, the older plots are still visible.

**To Reproduce**
In `doc/api/target/`,
```python
python class_balance.py
```

**Expected behavior**
class_balance_compare
class_balance

**Traceback**
class_balance_compare
class_balance

The second plot contains the first plot.

This can be solved by adding `plt.gcf().clear()` after `viz.poof()`.

 - OS: macOS & Ubuntu
 - Python Version 3.6
 - Yellowbrick Version dev

zjpoh:
Fix #579

### DIFF
--- a/yellowbrick/base.py
+++ b/yellowbrick/base.py
@@ -209,6 +209,7 @@ class Visualizer(BaseEstimator):
 
         if outpath is not None:
             plt.savefig(outpath, **kwargs)
+            plt.gcf().clear()
         else:
             plt.show()
 


zjpoh:
Ready for review.

zjpoh:
@bbengfort What if we add an optional parameter to poof?
```python
def poof(self, outpath=None, clear_plot=False, **kwargs):
    ...
    if clear_plot:
        plt.gcf().clear()
```

bbengfort:
@zjpoh excellent suggestion! Could we call the parameter `clear_figure` though?

yellowbrick/base.py
```diff
@@ -181,7 +181,7 @@ def finalize(self, **kwargs):
         """
         return self.ax
 
-    def poof(self, outpath=None, **kwargs):
+    def poof(self, outpath=None, clear_figure=False, **kwargs):
```
bbengfort: Awesome thank you so much! Would you mind adding this to the `poof` docstring under parameters? It might be nice to explain why you might want to set this to True (e.g. explain the issue you had.

zjpoh: Thanks for the reminder. Always forget docstring..... >.<

Issue #583, opened 2018-08-22
SilhouetteVisualizer Docstring


bbengfort:
The [SilhouetteVisualizer](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/cluster/silhouette.py#L40) does not currently have a docstring; I'm hoping someone would be willing to add one before the next release! 

See #516 -- but perhaps all the checklists deterred folks from taking it on, so here I'll create a single high priority issue for the docstring!

bbengfort:
Thanks to @rlshuhart we now don't have an ugly TODO and we have a description! Final wrap up for this issue includes adding the Parameters and Attributes list.

PR #586, opened 2018-08-23
Update silhouette.py

rlshuhart:
In regards to #583. Taken from http://www.scikit-yb.org/en/latest/api/cluster/silhouette.html.

### DIFF
--- a/yellowbrick/cluster/silhouette.py
+++ b/yellowbrick/cluster/silhouette.py
@@ -37,7 +37,11 @@ __all__ = [
 
 class SilhouetteVisualizer(ClusteringScoreVisualizer):
     """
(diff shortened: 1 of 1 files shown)


rebeccabilbro:
Thank you @rlshuhart for jumping in to fix this empty docstring! It definitely makes a difference when users look at our docs at scikit-yb.org, since Sphinx reads in those docstrings to populate the API docs! 

It looks like Appveyor (aka our automated tests) failed the first time (it happened when it was trying to pull the datasets, which doesn't have anything to do with the changes you made); I've relaunched the build, and once the tests pass, I'll go ahead and merge this in!

rlshuhart:
Glad to help!

Issue #516, opened 2018-07-20
Improvements to Silhouette Visualizer


bbengfort:
The following improvements to the Silhouette Visualizer are left over from #91: 

*Note to contributors: items in the below checklist don't need to be completed in a single PR; if you see one that catches your eye, feel to pick it off the list!*

- [x] Improve the documentation describing what Silhouette scores are and how to use the visualizer to qualitatively evaluate a clustering solution. 
- [ ] Find a real world example rather than just using make_blobs (note: we also have an example using the Iris dataset; ideally we'd having something a bit more unique to YB that we can add to `yellowbrick.datasets` module - perhaps this should be a separate issue?). 
- [x] Instead of hard fixing the limits of the X-axis from -1.0 to 1.0; be more flexible so that the visualizer has a better display (or give the user the option of setting the limits). 
- [x] Move the cluster identity labels away from the middle and to the y-axis.
- [x] Add ability to define cluster colors and improve color selection methodology. 
- [x] Add a legend/annotation that describes the average clustering coefficient (e.g. label the red axvline)

gokriznastic:
@bbengfort  for the third item in the checklist i.e. 

>  Instead of hard fixing the limits of the X-axis from -1.0 to 1.0; be more flexible so that the visualizer has a better display (or give the user the option of setting the limits).

it would be counter-intuitive to provide that flexibility because Silhouette Scores by mathematical definition be ranges from −1 to +1, where a high value indicates that the object is well matched to its own cluster and poorly matched to neighboring clusters.

Please correct me if I did not understand this improvement as you intended.

rebeccabilbro:
Hi @gokriznastic and thanks for your interest in Yellowbrick! 

What @bbengfort is suggesting here is not to change the scale along which computed silhouette scores are computed (which is indeed from -1 to 1), but to change the way that our `SilhouetteScore` visualizer scales the axes of the plot. In other words, if the visualizer computes the silhouette scores for our clusters, and they fall between -0.7 and 0.6, it should adjust the axes of the plot accordingly. One strategy would be to do something like...
```
ax.xlim([min(self.silhouette_samples_), max(self.silhouette_samples_)])
```
...in [L198](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/cluster/silhouette.py#L198) (though some additional padding on either side might be necessary, so this might take a bit of plotting experimentation). Would you be interested in working on this?

gokriznastic:
[...]

PR #591, opened 2018-08-27
Finishes SilhouetteVisualizer docstring, completing #583

rebeccabilbro:
This PR adds parameters, attributes, and an example to the SilhouetteVisualizer docstring, fixing #583.

### DIFF
--- a/yellowbrick/cluster/silhouette.py
+++ b/yellowbrick/cluster/silhouette.py
@@ -37,11 +37,65 @@ __all__ = [
 
 class SilhouetteVisualizer(ClusteringScoreVisualizer):
     """
-    The Silhouette Visualizer displays the silhouette coefficient for each 
-    sample on a per-cluster basis, visualizing which clusters are dense and 
-    which are not. This is particularly useful for determining cluster 
-    imbalance, or for selecting a value for K by comparing multiple 
+    The Silhouette Visualizer displays the silhouette coefficient for each
+    sample on a per-cluster basis, visually evaluating the density and
+    separation between clusters. The score is calculated by averaging the
+    silhouette coefficient for each sample, computed as the difference
+    between the average intra-cluster distance and the mean nearest-cluster
+    distance for each sample, normalized by the maximum value. This produces a
+    score between -1 and +1, where scores near +1 indicate high separation
+    and scores near -1 indicate that the samples may have been assigned to
+    the wrong cluster.
+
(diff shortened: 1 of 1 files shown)


yellowbrick/cluster/silhouette.py
```diff
@@ -60,12 +114,12 @@ def __init__(self, model, ax=None, **kwargs):
...
-        Fits the model and generates the the silhouette visualization.
-
-        TODO: decide to use this method or the score method to draw.
-        NOTE: Probably this would be better in score, but the standard score
-        is a little different and I'm not sure how it's used.
+        Fits the model and generates the silhouette visualization.
         """
+        # TODO: decide to use this method or the score method to draw.
+        # NOTE: Probably this would be better in score, but the standard score
+        # is a little different and I'm not sure how it's used.
```
bbengfort: Thanks for moving these out of the docstring!

yellowbrick/cluster/silhouette.py
```diff
@@ -37,11 +37,65 @@
...
+        Silhouette Coefficient for each samples. Computed via scikit-learn
+        `sklearn.metrics.silhouette_samples`.
+
+    n_samples : integer
+        Number of total samples in the dataset (X.shape[0])
+
+    n_clusters : integer
+        Number of clusters (e.g. n_clusters or k value) passed to internal
+        scikit-learn model.
+
```
bbengfort: `n_samples` and `n_clusters` aren't suffixed with an `_` -- and that's my bad; they should be since they're both computed in `fit()` and also the `# Required internal properties` part of the `__init__` function should be removed so that we can correctly determine if the estimator was fitted or not using the [`sklearn.validation.check_is_fitted`](http://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html) methodology (e.g. these attributes should not be members of the class until after `fit()` is called. 

Thank you so much for adding the docstring because it allows me to review the code that I haven't looked at in a while! I'm now remembering that `SilhouetteVisualizer` was a prototype and still had a lot of work remaining (e.g. in #516). I'd appreciate it if you wouldn't mind deleting lines 109-113 and adding the `_` suffix or adding this to the tasks to complete in #516.

rebeccabilbro: You got it!

bbengfort: Thanks for taking care of my technical debt!

PR #592, opened 2018-08-28
Color DispersionPlot markers based on target values

lwgray:
This PR creates a Dispersion Plot with color coded classes as requested by PR #549 

If the target values are specified, each word in a document is annotated with the target value associated with the document.  This translates into a colored marker representing the word and the document/class it belongs to

Additionally, when target values are specified, labels appear in a legend with the appropriate markers

image

Problem I am facing (Help Needed! 😭 )
-----------------------

I am still faced with the issue of a test passing when text is input as a generator and
no target specified.  Under these conditions, this results in a "ValueError: need at least one array to stack".  This isn't the case when a generator is used with a target specified.  Nor is it the case
when text is input as a list, with or without the target specified. 

Works under these conditions
1.  Input of text as a list and no target is specified
2.  Input of text as a list and target is specified
3.  Input of text as a generator and target is specified

Doesn't work under this condition
1.  Input of text as a generator and no target is specified


rebeccabilbro:
Awesome; thanks @lwgray - I'll take a look tomorrow! Can't wait!!

rebeccabilbro:
Hey there @lwgray! Looks like your tests are passing now; glad you worked through the bug! Does that mean this is ready for review?

lwgray:
It’s ready.

yellowbrick/text/dispersion.py
```diff
@@ -58,19 +71,30 @@ class DispersionPlot(TextVisualizer):
     process, but can and should be set as early as possible.
     """
 
-    def __init__(self, words, ax=None, color=None, ignore_case=False,
-                 annotate_docs=False, **kwargs):
+    # NOTE: cannot be np.nan
+    NULL_CLASS = None
+
+    def __init__(self, words, ax=None, colors=None, ignore_case=False,
```
rebeccabilbro: Is it necessary for `words` to be passed in on instantiation? It looks like the user has to pass `text` when they call `fit()`, and that `words` is either the same as `text` or could be derived from it.

yellowbrick/text/dispersion.py
```diff
@@ -41,9 +47,16 @@ class DispersionPlot(TextVisualizer):
...
-        Specify color for bars
+    labels : list of strings
+        The names of the classes in the target, used to create a legend.
+        Labels must match names of classes in sorted order.
+
+    colors : list or tuple of colors
+        Specify the colors for each individual class
 
+    colormap : string or matplotlib cmap
+        Sequential colormap for continuous target
```
rebeccabilbro: Continuous targets are associated with regression problems, but I think the goal for this visualizer is to be used on classification problems, right? If the user is going to pass in a colormap here, we may want to suggest a qualitative colormap rather than a sequential one, since that will probably be easier to discern classes.

yellowbrick/text/dispersion.py
```diff
@@ -95,34 +120,91 @@ def fit(self, text):
...
-        draws the distribution plot on it.
+        draws the plot on it.
         Parameters
         ----------
         kwargs: generic keyword arguments.
         """
 
+        # Resolve the labels with the classes
+        labels = self.labels if self.labels is not None else self.classes_
+        if len(labels) != len(self.classes_):
```
rebeccabilbro: Might be a good idea to add a test that make sure this error gets raised in the number of labels doesn't match the number of colors.

yellowbrick/text/dispersion.py
```diff
@@ -95,34 +120,91 @@ def fit(self, text):
...
 
+        if y is not None:
+            self.classes_ = np.unique(y)
+        elif y is None and self.labels is not None:
+            self.classes_ = np.array([self.labels[0]])
+        else:
+            self.classes_ = np.array([self.NULL_CLASS])
+
         # Create an index (e.g. the y position) for the target words
         self.target_words_ = np.flip(self.words, axis=0)
```
rebeccabilbro: What happens if one or more of the `self.target_words_` isn't in the corpus?

yellowbrick/text/dispersion.py
```diff
@@ -158,14 +245,27 @@ def dispersion(words, corpus, ax=None, color=None,
...
-        Specify color for bars
+    labels : list of strings
+        The names of the classes in the target, used to create a legend.
+        Labels must match names of classes in sorted order.
+
+    colors : list or tuple of colors
+        Specify the colors for each individual class
+
+    colormap : string or matplotlib cmap
+        Sequential colormap for continuous target
```
rebeccabilbro: Again, I think this should be qualitative colormap for discrete target

yellowbrick/text/dispersion.py
lwgray: @rebeccabilbro I am not sure that I understand.   `words` is the list of targeted words that the user wants to be marked throughout the document.  These are the words that appear on the y-axis.

yellowbrick/text/dispersion.py
lwgray: @rebeccabilbro Sorry, The previous wording got carried over from the Tsne visualizer, I will change the wording to reflect your suggestion of using a qualitative colormap...

yellowbrick/text/dispersion.py
rebeccabilbro: Gotcha! I think I'm getting mixed up because we have both `self.words` and `self.target_words_` here. To make it a bit easier to understand, could we perhaps change `self.words` to `self.target_words` and `self.target_words_` to `self.indexed_words_`, or something like that?

yellowbrick/text/dispersion.py
```diff
@@ -58,19 +71,30 @@ class DispersionPlot(TextVisualizer):
...
+        self.colormap = colormap
+
         self.words = words
         self.ignore_case = ignore_case
         self.annotate_docs = annotate_docs
 
-    def _compute_dispersion(self, text):
+    def _compute_dispersion(self, text, y):
         self.boundaries_ = []
         self.offset = 0
```
rebeccabilbro: Does `self.offset` need to be a class attribute? I looked but it doesn't appear to be used outside of the `_compute_dispersion` function.

yellowbrick/text/dispersion.py
```diff
@@ -95,34 +120,91 @@ def fit(self, text):
...
+
         # Create an index (e.g. the y position) for the target words
         self.target_words_ = np.flip(self.words, axis=0)
         if self.ignore_case:
             self.target_words_ = np.array([w.lower() for w in self.target_words_])
 
         # Stack is used to create a 2D array from the generator
-        points = np.stack(self._compute_dispersion(text))
-        self.draw(points)
+        self.points_target = np.stack(self._compute_dispersion(text, y))
```
rebeccabilbro: Does `self.points_target` need to be a class attribute? It doesn't appear to be used outside of `fit()`.

yellowbrick/text/dispersion.py
```diff
@@ -95,34 +120,91 @@ def fit(self, text):
...
+
+        self.points = np.stack(zip(self.points_target[:,0].astype(int),
```
rebeccabilbro: It looks like `self.points` could just be `points`, since it isn't used outside of `fit()`.

yellowbrick/text/dispersion.py
lwgray: @rebeccabilbro Great Catch!  I came across this the other day and forgot about it.  If no target words are in the corpus then an error is raised.  

I now have to decide if I present a blank graph or raise a yellowbrickValueError in this case

rebeccabilbro: As a user, it would be great to get a custom `YellowbrickValueError` in this case! Maybe some thing like "Could not find {} in the corpus."

yellowbrick/text/dispersion.py
```diff
@@ -79,13 +103,14 @@ def _compute_dispersion(self, text):
...
+                    yield (self.offset, y_coord, target)
             if self.annotate_docs:
                 self.boundaries_.append(self.offset)
         self.boundaries_ = np.array(self.boundaries_, dtype=int)
 
-    def fit(self, text):
+    def fit(self, text, y=None, **kwargs):
         """
         The fit method is the primary drawing input for the dispersion
         visualization. It requires the corpus as a list of words.
```
rebeccabilbro: It looks like there may have been some changes since the original implementation and might need to update the docstring here; does `DispersionPlot` still expect the input data to be a list of words? It looks like it could either be a list of documents or a generator that would yield a list of documents.

yellowbrick/text/dispersion.py
```diff
@@ -79,13 +103,14 @@ def _compute_dispersion(self, text):
...
-                    yield (self.offset, y)
+                for y_coord in (self.target_words_ == word).nonzero()[0]:
+                    y_coord = int(y_coord)
+                    yield (self.offset, y_coord, target)
             if self.annotate_docs:
                 self.boundaries_.append(self.offset)
         self.boundaries_ = np.array(self.boundaries_, dtype=int)
 
-    def fit(self, text):
+    def fit(self, text, y=None, **kwargs):
```
rebeccabilbro: Would you mind if we changed `text` to `X` to make it a bit more similar to how input features are referenced in scikit-learn and in our other text visualizers (e.g. [FreqDist](http://www.scikit-yb.org/en/develop/_modules/yellowbrick/text/freqdist.html#FrequencyVisualizer.fit) and [TSNE](http://www.scikit-yb.org/en/develop/_modules/yellowbrick/text/tsne.html#TSNEVisualizer.fit)?

yellowbrick/text/dispersion.py
```diff
@@ -95,34 +120,91 @@ def fit(self, text):
         text : list
```
rebeccabilbro: For our other text visualizers, we allow users to pass data to `fit()` as either an ndarray or a dataframe, so we may want to anticipate the case that a user passes in the corpus as a dataframe here,

yellowbrick/text/dispersion.py
rebeccabilbro: Awesome, thank you!

rebeccabilbro:
This is looking great @lwgray - I'm so excited to be able to use the new color-by-class feature! I've added in some inline comments above; please let me know if you have any questions. Are you thinking about updating the `DispersionPlot` docs as part of this PR or of doing that in a subsequent one?

yellowbrick/text/dispersion.py
lwgray: I agree!

yellowbrick/text/dispersion.py
lwgray: Are you referring to the `y` variable?

rebeccabilbro: I was referring to the `text` (i.e. the corpus data), which is passed to `fit()`.

yellowbrick/text/dispersion.py
lwgray: Maybe I am missing something very obvious.  Won't the number of labels always match the number of color since the number of colors is dictated by the number of labels.  L179-181

yellowbrick/text/dispersion.py
lwgray: duh! 👎 FAIL

yellowbrick/text/dispersion.py
rebeccabilbro: Sorry - I meant to type "if the number of labels doesn't match the number of classes". I am referring to you `if` statement on line 172. I was suggesting that you add a test that ensures the error is raised if the number of labels doesn't match the number of classes.

lwgray:
@rebeccabilbro I plan on doing a separate PR for the docs as I am quite busy this week.  I don't want to hold up your usage of this Visualizer for your PyData Talks

rebeccabilbro:
Ok, no prob! Mind opening up a new issue for the docs (if there isn't already one)?

yellowbrick/text/dispersion.py
lwgray: I believe I am already catching this mismatch in my tests...  [L115-L129](https://github.com/DistrictDataLabs/yellowbrick/blob/5d6da64aefaa47208fe36db5691ae1200540d3d0/tests/test_text/test_dispersion.py#L115-L129)

rebeccabilbro: How about using a `match` inside your test to ensure that the custom error message you have written for the `YellowbrickValueError` is raised. Something like [this](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/test_classification_report.py#L166-L172) or [this](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_draw.py#L33-L34).

lwgray: Got it!

yellowbrick/text/dispersion.py
lwgray: How would the dataframe be structured?
rows = documents
columns = words

lwgray:
In the last commit,  I address all except 2 of your suggestions.  They require a bit more effort and will be completed ASAP.

To be completed:
-------------------
Test pandas dataframe as input into `fit`
Custom yellowValueError when a word is not found in the corpus

Below are the corrections that were made thus far.
--------------------------------------------------------
1.  Fix colormap docstring entry by changing wording to "Qualitative colormap for discrete targets"
2.  To make it a bit easier to understand, we changed self.words to self.target_words and self.target_words_ to self.indexed_words
3.  Changed name of text variable from 'text' to 'X' to make it a bit more similar to how input features are referenced in scikit-learn and in our other text visualizers (e.g. FreqDist and TSNE)
4. Because `self.offset` is only used in `_compute_dispersion` function it is prudent to change it from class attribute to regular variable `offset`
5.  Modify docstring for the `fit` method to reflect the change in how the method expects input.  It has changed from a list of words to a list/generator of documents that in itself contains a list of words.
6. Because `self.points` and `self.points_target` are only used in `fit` function it is prudent to change it from class attribute to regular variable `points` and `points_target`
7. Create Test that raises if number of labels doesn't match the number of classes

yellowbrick/text/dispersion.py
lwgray: I finally got a fix for this..it is in the latest commit

examples/rebeccabilbro/dispersion_plots.ipynb
```diff
@@ -0,0 +1,436 @@
...
...Gm1krPHu6eWnI+ctCiqskSZJkiRKZpOuBz/C34n4i6eml26ukFpjZ6vgcvK/j1/YqSTfW36r5keIqSZIkSZKkhuScqyRJkiRJkhqS4ipJkiRJkqSGpLhKkiRJkiSpISmukiRJkiRJakhWaE+SJFlOMbOdgYui6GSpEvcGsbo1MFFSr8L2vQlvucKy4cBgSWMWsy9tgXUkPVH0nGtijLl4XakWeKHKyyXdXM/2M3AvwsHAcEkPLVLnK8feFHhY0rfi86G4j+NGkt4ys9a4hc/29YSpL/4N1LjP9eyrN2XXvZHtvvAf/E/0qzmTmaskSZIlwF377DujVOuqFpjZGbi5ccmeBEm9QmgdhNuYnFar/TWCH9B4G5lqzJHUNTwHuwGXNaFAZk0J8+UWYTINXvn+TrzqPLj3339cGCXLJ5m5SpIkWT6ZBvTAsynlnAdcKWlWY4OFFcxQYEu8PtFA3I7lAkn7mVkvvNjndma2K/AjSX2j7cZAb2CumZWqnQ8ysy3i54Nwb8HBwFb4g/25DWTLSqbPn4fB8SBcSG4Ybe9pzDFIus3MvgNcCczDLVv6RB9uxb3/voEbF59YFvIxoKOZ3Yf7/vXBq8ffCHQFhprZisA1uEHxhsB9ks41sx64cfanuGVNr6hkX5V6Yt1AZLnM7HsRq7eZTcV9Bg14Exe4LeMcbB4/l0y4dzGzR4D1gUGSrjWzLrj/5Dz89+l43Lj5FmAdCobTSdPIzFWSJMlyiKQ7qWArYmZfA3YHbqjS9DAzG1P6h2eIwG+sb0vqAHQHfoMbBm9uZl/BMzbzzezrwIG4QXOpL6/H/gZKmhSLr4ss2gxgD9xsebakztH+DxX6tm706wncRua2WL4NcJmkPYC+wElVjm2hY4jM0xDg5MiIXY0LR3Dfu2OBdsA+4ZtX5FGgE268PBl4GtjezFbAvfmexIXQxPA4bAecEG0PBS6R1BH3ZFyrSp+LVItVjS2B/5bUHhdNO0WbGbGsF7BzbPspsBcudPtFRnAI0CPOy+u4QD4BN/HujAu9ZBFIcZUkSdK86AkMkzSvyvphMfTWNcTP6FjeBngCIPzsXsYzOg/jQ2Cb4hmN7rjgGNVAPybH/2/gnn9tcQEzBh9eW6kw5FaiNCzYGc+8/MDMOgKzgOPN7E/4zX/lKvusdgwbSXo2tnkC2DZ+flXSB3GuZlEYYg3G4Mb
```
lwgray: the variable `chapter_text` within the `fit` method should be `X` instead

PR #595, opened 2018-08-30
Removed Override of Poof Method in JointPlot

archaeocharlie:
The poof method seems to override the base poof without including any `plt.show()` and no plot appears when running code. 

This removes the `JointPlot` classes overriden poof method so that the base method is called instead.


yellowbrick/features/jointplot.py
```diff
@@ -292,15 +292,6 @@ def draw_xy(self, X, y, **kwargs):
             self.y_ax.hist(y, bins=hist_bins, color=histcolor_y,
                               orientation='horizontal', **self.xy_args)
 
-    def poof(self, **kwargs):
-        """
-        Creates the labels for the feature and target variables
-        """
-
-        self.joint_ax.set_xlabel(self.feature)
-        self.joint_ax.set_ylabel(self.target)
```
bbengfort: @archaeocharlie thanks for submitting the PR - I think the tests will pass if you move lines 300 and 301 to the `finalize` method, at the top.

pdamodaran:
Hi @archaeocharlie - thanks for catching this issue and submitting the PR!  As @bbengfort  mentioned, all you need to do is take the code that is inside the poof method and move it to the finalize method.  That will ensure the x label and the y lablel get displayed and the tests pass.  Let us know if you have any questions.

bbengfort:
@archaeocharlie @pdamodaran I jumped in here briefly since it actually is a bit more work to fix the tests then I first thought. I updated the tests to not call `poof()` and to call `finalize()` directly, and I added some warning filters and resync'd the images. Hopefully that gets the CI tests to pass.

archaeocharlie:
@bbengfort @pdamodaran Sorry for the delay, but I've been away. Anything else I need to do here? I see the checks still failed??

pdamodaran:
@bbengfort - I moved the two lines to the finalize method as you had suggested earlier and ran the tests on my machine and ran into the same two failures.  Do you have any thoughts on what could be causing the tests to fail?

rebeccabilbro:
@archaeocharlie @pdamodaran - apologies for the slow response! Have you experimented with removing the call to `finalize` in the two visual test cases that are failing (`test_jointplot_has_no_errors` and `test_jointplot_integrated_has_no_errors`)?  This will simplify the tests by removing the extra text that is added to the plot in the call to `finalize`. 

Note that after removing the call to `finalize`, you may have to update the baseline images; our visual tests compare what is in the baseline images folder with those in the actual images folder (as described [here](http://www.scikit-yb.org/en/latest/contributing.html#image-comparison-tests)). We have a utility for doing this:

```
$ python -m tests.images tests/test_features/test_jointplot.py
```

If that doesn't work, let me know and we can brainstorm together!

pdamodaran:
@rebeccabilbro @archaeocharlie I discovered that the code change resulted in the x and y labels no longer showing up in the images and the x tick labels and the y tick labels show up on both sides of the axis as opposed to the left and bottom as they had been previously.  I experimented with moving the code that sets the x label and the y label to the draw_joint() method, which works for the case when the feature and the target labels are provided, but does not if they are not.  I also haven't figured out how to get the x and y tick labels to only show up on the left and bottom.  I have notebooks on my repo showing the difference.  The notebooks are called jointplot_old.ipybb and jointplot_new.ipynb.

https://github.com/pdamodaran/yellowbrick/tree/develop/examples/pdamodaran

@archaeocharlie - would you like to take a stab at this?

rebeccabilbro:
@pdamodaran the label issue you're describing should be fixed with [the commit that @bbengfort pushed to this PR a couple days ago](https://github.com/DistrictDataLabs/yellowbrick/pull/595/commits/eb5935c421818da2f763da60898b9b0b67985ef8), where he added:

```
        self.joint_ax.set_xlabel(self.feature)
        self.joint_ax.set_ylabel(self.target)
``` 

to the `finalize` method -- would you please pull his changes and try again? 

The tests will still need to be fixed though; let me know if removing the calls to `finalize` in those two tests as I described above does the trick!

pdamodaran:
@rebeccabilbro I failed to mention that I tried Ben's solution first and the labels did not show up.  I then experimented with moving those two lines of code under the draw_joint method.  Also, the two tests no longer have the call the finalize.

rebeccabilbro:
@pdamodaran @archaeocharlie - ok, I've removed `finalize` from those two tests, and the automated test suite is now running again. If the image comparison tests continue to fail, I'll bump up the tolerances until they pass.

@pdamodaran - the issues you mention with the labels are indeed problematic, but since they pre-date the removal of `poof` in this PR, let's aim to address them in a separate PR that properly refactors `JointPlot` so that we more effectively handle the double axes -- I recommend we try something similar to the new version of [`ResidualsPlot`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/regressor/residuals.py#L428-L444).

pdamodaran:
@rebeccabilbro - thanks for taking care of this.  I must not have had the latest code for tests.  There is an issue already created to improve the JointPlotVisualizer.  Should the issue with the images be included in that issue or should I open up a new issue?

rebeccabilbro:
@pdamodaran - technically it's a bug, so I'm leaning towards making it a separate issue that we can label as such (though we should make sure there are links between those two issues to keep everyone in the loop, as the solution to one may well impact the other). Either way, if you wouldn't mind adding in some plots to document the bug, that would be a great help us in remembering later what it is we're fixing!  Also, if you could note that this visualizer produces markedly different results depending on whether you do:

```
_, ax = plt.subplots()
visualizer = JointPlotVisualizer(ax=ax, ...)
```

or 

```
fig = plt.figure()
ax = fig.add_subplot()
visualizer = JointPlotVisualizer(ax=ax, ...)
```

... I think that may offer some clues about how best to refactor. Thanks for following up on this @pdamodaran!

PR #608, opened 2018-09-13
Bugfix to prevent wrong percent points showing up in ConfusionMatrix

dschoenleber:
Proposed fix to prevent incorrect percent numbers to be displayed in the confusion matrix since class_counts_ is a row vector, not a column vector.


rebeccabilbro:
Hey there @dschoenleber and thank you for the PR to fix our percent count bug in `ConfusionMatrix`! Looks like our automated tests failed the first time on the visual comparisons, so I just updated the baseline images and pushed them to your branch; the tests are running again now, so we'll see how they turn out!

dschoenleber:
Hi @rebeccabilbro you're welcome and thank you for fixing the build issue!

PR #611, opened 2018-09-15
Adds docs for how to use class labels w/ confusion matrix

rebeccabilbro:
In follow up to the [bug](https://github.com/DistrictDataLabs/yellowbrick/issues/244) that prevented adding class labels to a `ConfusionMatrix`plot, and the subsequent PR (#253) that fixed it, this PR adds an explanation about how to use class labels with this visualizer to our documentation. 

*Note: in working on this PR, I realized that some elements of the plot size appear to be hard-coded in, which results in the labels [sometimes being cut off](https://github.com/rebeccabilbro/yellowbrick/blob/conf-matrix-docs/docs/api/classifier/images/confusion_matrix_iris.png) -- though they seem to look better [in a Jupyter notebook](https://github.com/rebeccabilbro/yellowbrick/blob/conf-matrix-docs/examples/rebeccabilbro/conf_matrix_class_names.ipynb).*


bbengfort:
@rebeccabilbro I noticed the tick labels and axis labels getting cut off as well - where did you see a hardcoded size? I remember we tried to fix that at one point with the `fontsize` property, perhaps we missed an instance or two. 

The tick labels are being set on [L238](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/confusion_matrix.py#L238) - the font size is being used there, but maybe the vertical orientation is causing problems? It's probably worth opening a bug report. Maybe we can make the tick label rotation an option as well (though we also have the same issue in Rank2D I think).

rebeccabilbro:
Thanks @bbengfort! I removed the weirdly sized image; sorry - I'm not sure where that came from. As for the hardcoding, I'm wondering if it might be happening in [L274](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/confusion_matrix.py#L274). Tick label rotation is a good solution.

PR #618, opened 2018-09-19
Require matplotlib>=1.5.1,<3.0

rebeccabilbro:
The new major release of [Matplotlib (version 3.0.0)](https://matplotlib.org/), seems to silently break some of our images. For example:

This is `ResidualsPlot` with `matplotlib==2.2.2`
resids

And this is `ResidualsPlot` with `matplotlib==3.0.0`
uhoh

As a temporary fix, this PR updates all of our `requirements.txt` files to specify `matplotlib>=1.5.1,<3.0`  until we can get to the bottom of what's actually going on.


bbengfort:
Well, that commit fixed all of the tests but 1 - manifold regression (which has been variable in the past), rerunning now to see if that will fix it.

rebeccabilbro:
Thanks for the assist with the broken image tests @bbengfort!

PR #639, opened 2018-10-21
added Iris dataset clustering example and improved scaling of SilhouetteVisualizer

gokriznastic:
Fixes two items from checklist in #516

Added example python notebook with real-world dataset, partially addressing the second item from the checklist of improvements.
Improved scaling of the SilhouetteVisualizer axes.


rebeccabilbro:
Thanks @gokriznastic -- it looks like this is coming together! I will aim to take a look in the next few days; in the meantime, there's one more thing you could do if you're interested. The current automated test failures are due to the updates you've made, since they change the images generated by the test suite. This is totally expected and happens every time we update our visualizers, and it's easy to fix. Essentially, the automated tests run through all the visualizers and build all the images, and then check to see if all the images match the ones in `yellowbrick/tests/baseline_images`; if anything doesn't match, it gives an error. You can fix this by doing the following:

1. run the tests locally on your machine by doing (from inside the top-level directory of your cloned yellowbrick repository):
    ```
    cd tests
    pip install -r requirements.txt
    pytest
    ```

2. then update the `yellowbrick/tests/baseline_images`:
    ```
    cd ..
    python -m tests.images tests/test_cluster/test_silhouette 
    ```

3. then `git add`, `git commit`, and `git push` the new images to your feature branch.

That should fix the tests!

gokriznastic:
Hey @rebeccabilbro, no problem! 
Yeah, I saw the commits. So, were the changes made in the silhouette visualizer alright? This is my first PR contribution so any feedback would be great.

Also, it seems like all tests passed this time.
This fixes 2 items from the checklist of improvements in #516 then I suppose.

yellowbrick/cluster/silhouette.py
```diff
@@ -206,4 +214,4 @@ def finalize(self):
 
         # Set the ticks on the axis object.
         self.ax.set_yticks([])  # Clear the yaxis labels / ticks
-        self.ax.set_xticks(np.linspace(-1,1,11))
+        self.ax.xaxis.set_major_locator(plt.MultipleLocator(0.1))  # Set the ticks at multiples of 0.1
```
rebeccabilbro: The `MultipleLocator` is a great strategy to use here! The one thing I would ask is that we modify the ticker on the subplot axes (`self.ax`) instead of the whole `plt`, since we want to leave things flexible for plots composed of multiple subplots containing multiple visualizers. It looks like you can achieve this using the strategy shown [here](https://matplotlib.org/gallery/ticks_and_spines/tick-locators.html), by adding `import matplotlib.ticker as ticker`  to the import statements at the top and changing this line to `self.ax.xaxis.set_major_locator(ticker.MultipleLocator(0.1))`.

yellowbrick/cluster/silhouette.py
```diff
@@ -194,8 +195,15 @@ def finalize(self):
         ))
 
         # Set the X and Y limits
-        # The silhouette coefficient can range from -1, 1
-        self.ax.set_xlim([-1, 1])
+        # The silhouette coefficient can range from -1, 1;
+        # but here we scale the plot according to our visualizations
+
+        # l_xlim and u_xlim are lower and upper limits of the x-axis,
+        # set according to our calculated maximum and minimum silhouette score along with necessary padding
```
rebeccabilbro: Thank you for not only making the change but for adding in the descriptive comments that explain what is happening here!

yellowbrick/cluster/silhouette.py
```diff
@@ -18,6 +18,7 @@
 ##########################################################################
 
 import numpy as np
+import matplotlib.pyplot as plt
```
rebeccabilbro: See my comment below about using `import matplotlib.ticker as ticker` rather than `import matplotlib.pyplot as plt` and let me know what you think.

rebeccabilbro:
Also, forgot to mention @gokriznastic - be sure to do a `git pull` of your branch from your remote fork to your local machine since you'll want to grab the edits I pushed directly to your branch before you make any changes, so we don't end up with a merge conflict!

gokriznastic:
@rebeccabilbro Thank you very much for the feedback. I'll incorporate the changes requested as soon as I do some reading and testing of your suggestions; meanwhile giving others the time to weigh in and request any other changes that might be required.

No worries, I'll update my local copy of the branch at once! Thanks for the reminder.

yellowbrick/cluster/silhouette.py
gokriznastic: I completely agree with this strategy. I kind of missed the scenario where there can be multiple subplots. Doing this right away.

bbengfort:
@rebeccabilbro @gokriznastic - I agree that a manual instantiation of a `ticker.MultipleLocator` is probably the best bet! Looks like the tests are passing with the new baseline images. These tests do not include negative silhouette scores (even though it's possible it's tough to generate an example of this). I would suggest adding the following to the tests (though this should not block merging the PR):

```python 
@pytest.mark.skip(reason="no negative silhouette example available yet") 
def test_negative_silhouette_score(self):
    """
    Ensure negative silhouette scores are correctly displayed by the visualizer. 
    """
    raise NotImplementedYet("no negative silhouette example available") 
```

gokriznastic:
@bbengfort you are absolutely right that it's tough to get negative silhouette scores, though I came across a few cases while testing this new scaled visualization. I think I got upto -0.3, but not more than that. I could try repeating some tests I did if you would like to see cases of negative silhouette score and the corresponding Yellowbrick visualization.

Also, do I need to add the above lines in `test_silhouette.py`?

bbengfort:
@gokriznastic yes, please add that test to `tests/test_cluster/test_silhouette.py` under the `SilhouetteVisualizerTests` class -- thank you! 

Also, that's awesome if you found a case with a negative silhouette score, if you'd like to add that in this PR - that's great; otherwise we can certainly do that in another PR.

gokriznastic:
@bbengfort I'll run the tests that I previously did and try adding negative silhouette score cases in the next PR. I would be opening another PR to try & solve other items on the checklist on #516 eventually.
Would you like any more changes in this PR related to scaling?

Also, while adding the test to `test_silhouette.py` I am getting an unresolved reference error for `NotImplementedYet`. I understand this is due to the this not being defined anywhere before. Am I missing something?

bbengfort:
@gokriznastic sounds good - the only thing I'd say is that a user-defined scale would be good, e.g. the user might want to ensure that all silhouette visualizations are from [0,1] to compare multiple silhouettes, but that can be done in a different PR if you'd like. 

Sorry, in Python it's [`NotImplementedError`](https://docs.python.org/3/library/exceptions.html#NotImplementedError) - my mistake.

bbengfort:
@gokriznastic that sounds like a good approach to me. 

Yes, this PR is ready to be merged, I've approved the PR - @rebeccabilbro, do you want to look over the changes and make sure your requirements are met? I've also merged with develop, so we're just waiting for the final tests to run so we can merge.

rebeccabilbro:
@gokriznastic thank you for your excellent work on this; we hope you'll consider future contributions to Yellowbrick!

gokriznastic:
@rebeccabilbro @bbengfort thanks a lot for the opportunity! I will be opening a new PR in coming days to address other improvements in #516.

Issue #613, opened 2018-09-16
Labels are sometimes cut off in ConfusionMatrix 


rebeccabilbro:
**Describe the bug**
When a dictionary of encoding-to-string mappings is provided to the `ConfusionMatrix` so that the class names will appear as labels, sometimes the class and axis labels get cut off.

![confusion_matrix_iris](https://user-images.githubusercontent.com/8760385/45598773-2cfd1900-b9af-11e8-9d9b-d6e598b24da4.png)

**To Reproduce**

```python
from sklearn.datasets import load_iris
from sklearn.linear_model import LogisticRegression
from sklearn.model_selection import train_test_split as tts

from yellowbrick.classifier import ConfusionMatrix


iris = load_iris()
iris_X = iris.data
iris_y = iris.target
iris_classes = iris.target_names
i_X_train, i_X_test, i_y_train, i_y_test = tts(
    iris_X, iris_y, test_size=0.2
)
model = LogisticRegression()
iris_cm = ConfusionMatrix(
    model, classes=iris_classes,
    label_encoder={0: 'setosa', 1: 'versicolor', 2: 'virginica'}
)
iris_cm.fit(i_X_train, i_y_train)
iris_cm.score(i_X_test, i_y_test)
i = iris_cm.poof()
```

**Desktop (please complete the following information):**
 - OS:  macOS
 - Python Version 3.6
 - Yellowbrick Version 0.8

**Additional context**
Note that the behavior appears to be somewhat different when the code is executed [inside a Jupyter notebook](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/examples/rebeccabilbro/conf_matrix_class_names.ipynb). Discovered in the process of working on #611

alexander-ponomaroff:
Has this issue been fixed? I don't experience any problems with labels when reproducing your result.

rebeccabilbro:
Hi there @alexander-ponomaroff and welcome to Scikit-YB! Yes, as far as I know, this bug is still an issue, though it doesn't manifest in all scenarios -- I was only seeing it outside of Jupyter notebooks, for instance. Would you mind documenting your OS version, Python version, and YB version and also letting us know whether you were using YB inside or outside an ipynb?

lwgray:
I am reopening this issue because during my implementation of plot directive for the ConfusionMatrix Docs #747 I came across it again.  It doesn't appear to be fixed with recent mpl updates.

* OS:  Ubuntu 18.04.1 LTS
[...]

PR #640, opened 2018-10-23
Autolayout to rcParams to address #613

zhampel:
I added a line to rcmod.py to do auto layout formatting. It worked on my Mac laptop in and out of the jupyter notebook using @rebeccabilbro's example script.

### DIFF
--- a/yellowbrick/style/rcmod.py
+++ b/yellowbrick/style/rcmod.py
@@ -198,6 +198,7 @@ def _axes_style(style=None, rc=None):
         # Common parameters
         style_dict = {
             "figure.facecolor": "white",
+            "figure.autolayout" : True,
             "text.color": dark_gray,
             "axes.labelcolor": dark_gray,
             "legend.frameon": False,


ndanielsen:
Just took a close look and tested this out. Looks like this change the size of the generated image which I don't think that we want to do as this might negatively affect some current users.  We all need to set the global figure size if we are going to make this change.

Current image size:


ranking_develop

With this change:

ranking_autolayout

bbengfort:
@zhampel this is certainly a creative solution to #613 -- but as @ndanielsen mentioned, this simple change affects every single one of our figures. My primary concern stems from the fact that some of our visualizers use `tight_layout` specifically (this is what autolayout does), while others don't because some figures cannot be laid out this way:

https://github.com/matplotlib/matplotlib/issues/7392

I'm not sure if this is fixed in matplotlib 3.0 or not yet, but it may be too difficult for us to go through all visualizers to see what a global change would do. Can I suggest instead that you add a `plt.tight_layout` call to [`ConfusionMatrix.finalize()`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/confusion_matrix.py#L285) -- this should have the same effect as the change to rcParams, but only influence the confusion matrix visualizer. 

Thank you again for figuring out the solution to this issue!

PR #642, opened 2018-10-26
Update the text of documentation code examples

Kautumn06:
I removed the deprecated `as_matrix` method from the code examples in Yellowbrick's documentation.

I also added comments or made small text changes to the comments in the code examples to help give our documentation a more uniform appearance. For example:

```
# Specify the features of interest and target
features = ["feature1", "feature2"]
target = "target variable"

# Extract the instances and target
X = data[features]
y = data[target]
```
And in the cases where I made small changes, I'll comments to files to help explain. Just let me know if you have any questions!

PR #647, opened 2018-10-29
Remove deprecated method from documentation

Kautumn06:
This PR is similar to #642, where I removed the deprecated `as_matrix` method from `.rst` files in our documentation. Here, I removed it from the `.py` files in the `docs/` folder, which are used to generate the images in the documentation.


docs/api/features/pcoords.py
```diff
@@ -15,9 +15,9 @@ def load_occupancy_data():
     features = ["temperature", "relative humidity", "light", "C02", "humidity"]
     classes = ['unoccupied', 'occupied']
 
-    # Extract the numpy arrays from the data frame
-    X = data[features].as_matrix()
```
Kautumn06: Removeed the deprecated `as_matrix` method.

Kautumn06:
Sorry @bbengfort, there was some confusion here on my part. Originally I had thought when I ran `make clean` and `make html` Sphinx auto-regenerated the images. I did check the rebuilt docs, but since I was only removing the deprecated `as_matrix()` method, I only checked to make sure that the images looked the same as before. 

However, I just ran the updated `.py` files and pushed the newly generated image `.png` files. Visually, they're identical to the ones before, but in this case, that is actually a good thing! So everything should be all set now, but please just let me know if you have any questions or need me to do anything else.

Kautumn06:
Hi @bbengfort, I added `plt.tight_layout()` and that was able to prevent the labels from being clipped. However, I'll also take a look at #446 and see if it's something I may be able to implement so we can save ourselves the extra step in the future.

PR #658, opened 2018-11-29
Bug fix proposal

jeromemassot:
But : impossible to edit colors for plot.
Cause : resolve_colors() did not use the correct parameter colors
Fix the problem

PR #659, opened 2018-11-30
Fix Bug mixing colors and labels if y is passed in the fit() and not …

jeromemassot:
…the constructor of the Visualizer

If the labels are passed only with in the fit() function, there is a mismatch between the colors and the labels due to the fact that the np.unique() method sorts its output.
Do avoid this behavior, one needs to bypass this automatic sorting by working on the unique indexes.


rebeccabilbro:
Hello again @jeromemassot and apologies for my slow response (it's exam season)! Thank you so much for the new PR; quick question -- did you plan to add to this your [proposed fix](https://github.com/DistrictDataLabs/yellowbrick/pull/658/files) to [`resolve_colors` function call](https://github.com/DistrictDataLabs/yellowbrick/blob/fd31bd0d8846955e639a12d0c118a1954e2d344f/yellowbrick/text/tsne.py#L324)? That fix would be very welcome!

jeromemassot:
Hi Rebecca,
no worries. Sorry but I do not understand your question?
If the fix is verified, I think that you can push in any branch you
consider valuable.
Good luck for the exams.
Best regards
Jerome

Le dim. 9 déc. 2018 à 15:06, Rebecca Bilbro <notifications@github.com> a
écrit :

rebeccabilbro:
Hello again @jeromemassot! I'm sorry it has taken me so long to respond to your pull request. I have been trying to reproduce the bug you mentioned, and I think I understand ([see this notebook](https://github.com/jeromemassot/yellowbrick/blob/patch-2/examples/rebeccabilbro/jerome_tsne_fixes.ipynb)) -- however, unless I am missing something, it does not appear that your solution changes the outcome (see the same notebook). Can you please clarify what your solution is aiming to do and provide a demonstration to illustrate how it works? Thank you!

jeromemassot:
Hi Rebecca,

Merry XMass and Happy New Year.

Your message is double:

1- about my bug, the problem is very difficult to detect as each time you
visualize the TSNE visualizer, you obtain a different figure due to the
random character of the mapping.
The bug is not observed in the attached notebook. It affects the resorting
of the colors when you use the np.unique() method. My fix avoids this
resorting and gives a consistent match between labels, colors and real data

2- the notebook you have done shows a bug when the both colors and labels
are passed to the constructor and the fit() method is used only with the
docs.
I think that it is due to a confusion in the code between the classes and
the labels : if the labels are passed in the constructor and if they are
passed in the fit() method, they were not used the same way and several
issues affected the original code.
I have done a fix that I attach to this email as a notebook.
The TSNEVisualizer_fixed class includes the both bug fixes : the first one
for the labels/color resorting problem discussed before and the second for
the crash we discuss in this second bullet item.

If you agree with these both fixes, you can use the code in the notebook
after you have removed the ## DEBUG code bricks :)

Best regards

Jerome


Le sam. 22 déc. 2018 à 09:38, Rebecca Bilbro <notifications@github.com> a
écrit :

rebeccabilbro:
Happy Holidays to you as well, @jeromemassot, and thank you for the added explanation about the bug!! I can't seem to access the new notebook you mentioned in your last message (it looks like the branch was deleted?) -- would you please re-send the link? I would very much like to ensure this is addressed in our next version release!

In the meantime, I'm going to go ahead and fix the other TSNE bug you noticed (self.color -> self.colors) -- I'll be sure to credit you for noticing the bug and identifying the solution!

Issue #660, opened 2018-11-30
Update default ranking algorithm in Rank2D documentation


Kautumn06:
In the documentation for the `Rank2D` visualizer, it says `covariance` is the default ranking algorithm when it is actually `pearson`. 
![image](https://user-images.githubusercontent.com/19958803/49296997-a91fdc80-f487-11e8-84c2-7bfdb50ca077.png)

@DistrictDataLabs/team-oz-maintainers

PR #661, opened 2018-11-30
Update default ranking algorithm in Rank2D docs

Kautumn06:
The `Rank2D` visualizer uses the Pearson score by default but the documentation says it uses `covariance` as the default ranking algorithm (and shows an example with the code and an image). Next, it is followed by an example using the Pearson score by setting `algorithm='pearson'`. 

So I not only updated the text but I also moved the example using the Pearson score above the covariance example since I think it makes more sense to first show the visualizer using the default settings and then show ways that it can be customized.

Resolves: #660

### DIFF
--- a/docs/api/features/rankd.rst
+++ b/docs/api/features/rankd.rst
@@ -27,7 +27,7 @@ In this example, we'll use the credit default data set from the UCI Machine Lear
 Rank 1D
 -------
 
-A one dimensional ranking of features utilizes a ranking algorithm that takes into account only a single feature at a time (e.g. histogram analysis). By default we utilize the Shapiro-Wilk algorithm to assess the normality of the distribution of instances with respect to the feature. A barplot is then drawn showing the relative ranks of each feature.
+A one-dimensional ranking of features utilizes a ranking algorithm that takes into account only a single feature at a time (e.g. histogram analysis). By default we utilize the Shapiro-Wilk algorithm to assess the normality of the distribution of instances with respect to the feature. A barplot is then drawn showing the relative ranks of each feature.
 
 .. code:: python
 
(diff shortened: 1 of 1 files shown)


docs/api/features/rankd.rst
```diff
@@ -27,7 +27,7 @@ In this example, we'll use the credit default data set from the UCI Machine Lear
 Rank 1D
 -------
 
-A one dimensional ranking of features utilizes a ranking algorithm that takes into account only a single feature at a time (e.g. histogram analysis). By default we utilize the Shapiro-Wilk algorithm to assess the normality of the distribution of instances with respect to the feature. A barplot is then drawn showing the relative ranks of each feature.
+A one-dimensional ranking of features utilizes a ranking algorithm that takes into account only a single feature at a time (e.g. histogram analysis). By default we utilize the Shapiro-Wilk algorithm to assess the normality of the distribution of instances with respect to the feature. A barplot is then drawn showing the relative ranks of each feature.
```
Kautumn06: Here I just added hyphenated "one-dimensional" since it's a compound adjective.

docs/api/features/rankd.rst
```diff
@@ -43,42 +43,40 @@ A one dimensional ranking of features utilizes a ranking algorithm that takes in
 
 .. image:: images/rank1d_shapiro.png
 
-
 Rank 2D
 -------
 
-A two dimensional ranking of features utilizes a ranking algorithm that takes into account pairs of features at a time (e.g. joint plot analysis). The pairs of features are then ranked by score and visualized using the lower left triangle of a feature co-occurence matrix.
+A two-dimensional ranking of features utilizes a ranking algorithm that takes into account pairs of features at a time (e.g. joint plot analysis). The pairs of features are then ranked by score and visualized using the lower left triangle of a feature co-occurence matrix.
```
Kautumn06: Added a hyphen to "two-dimensional".

docs/api/features/rankd.rst
```diff
@@ -43,42 +43,40 @@ A one dimensional ranking of features utilizes a ranking algorithm that takes in
...
 
-The default ranking algorithm is covariance, which attempts to compute the mean value of the product of deviations of variates from their respective means. Covariance loosely attempts to detect a colinear relationship between features.
+By default, the ``Rank2D`` visualizer utilizes the Pearson correlation score to detect colinear relationships.
```
Kautumn06: Since I moved the text, code, and image for the Pearson example above the covariance example, here it looks like I deleted the text describing `Rank2D` when used with `covariance` as the ranking algorithm. However, I simply moved it below, which you'll see further down.

PR #668, opened 2018-12-10
Add Polish to Missing Value Vizualizers in Contrib

ndanielsen:
This PR is addressing comments in the original PR for the missing values bar and dispersion visualizers. 

The original PR is located here: https://github.com/DistrictDataLabs/yellowbrick/pull/519

PR #673, opened 2018-12-11
Docs headings

Kautumn06:
In the documentation table of contents, I noticed that the subsection "API Reference" was no longer rendering for several of our visualizers, such as for `ConfusionMatrix`:

confusion_matrix

When I looked into it further, I saw that new subsections had been added to those pages uses `###` instead of `---`, which was what caused Sphinx not to render it properly. 

In addition, I added additional backticks to the inline code and variables names in the new text, since with only single backticks Sphinx italicizes the text.


docs/api/classifier/confusion_matrix.rst
```diff
@@ -3,22 +3,22 @@
 Confusion Matrix
 ================
 
-The ``ConfusionMatrix`` visualizer is a ScoreVisualizer that takes a
-fitted scikit-learn classifier and a set of test X and y values and
+The ``ConfusionMatrix`` visualizer is a ``ScoreVisualizer`` that takes a
```
Kautumn06: Added double backticks to inline visualizers and variable names.

docs/api/classifier/confusion_matrix.rst
```diff
@@ -3,22 +3,22 @@
...
 
-Below are a few examples of using the ConfusionMatrix visualizer; more
+Below are a few examples of using the ``ConfusionMatrix`` visualizer; more
 information can be found by looking at the
 scikit-learn documentation on `confusion matrices <http://scikit-learn.org/stable/modules/generated/sklearn.metrics.confusion_matrix.html>`_.
 
 .. code:: python
 
-    from sklearn.datasets import load_digits
+    from sklearn.datasets import load_digits, load_iris
```
Kautumn06: Added `load_iris` to the import statement since it is called in the second code example.

docs/api/classifier/confusion_matrix.rst
```diff
@@ -66,7 +66,7 @@ Class names can be added to a `ConfusionMatrix` plot using the `label_encoder` a
     y = iris.target
     classes = iris.target_names
 
-    X_train, X_test, y_train, y_test = tts(X, y, test_size=0.2)
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2)
```
Kautumn06: Changed `tts` to `train_test_split` since the function is imported above using `from sklearn.model_selection import train_test_split`

docs/api/classifier/rocauc.rst
```diff
@@ -53,7 +53,7 @@ This leads to another metric, area under the curve (AUC), which is a computation
 
 
 Multi-class ROCAUC Curves
-#########################
+-------------------------
```
Kautumn06: Created the sub-section using `---` instead of `###` so that it renders properly in the table of contents.

docs/api/contrib/missing/bar.rst
```diff
@@ -1,7 +1,7 @@
 .. -*- mode: rst -*-
 
 MissingValues Bar
-=============================
+=================
```
Kautumn06: Updated headings format to match Sphinx specifications.

bbengfort: @Kautumn06 thank you very much for fixing this, although @ndanielsen just fixed this as well in #668 -- no need to do anything, but I wanted you both to be aware in case there was a merge conflict when we got it sorted.

PR #678, opened 2018-12-29
Extend FeatureImportances docs/tests for stacked importances

### LINKED ISSUE #531: FeatureImportances stack documentation
Thank you so much @zjpoh for enhancing the `FeatureImportance` visualizer with a stack bar chart! There are a couple of minor tasks to complete that I will take care of in the next couple of days:

1. Add documentation about the stack argument and the implications 
2. Add `skipif` pandas not available, or non-pandas testing 
3. When stack gets set to False, issue warning about what happened 
4. Non-classifier multi-output regressors test 

See #492 and #510

bbengfort:
Wraps up the enhancement to the FeatureImportances visualizer which
added a stacked bar chart optional parameter in the case of
multi-dimensional importances. Updated the documentation to reflect the
stack=True situation as well as issued a warning if stack is False but
should be True. Updated tests for better coverage.

Closes #531


docs/api/text/umap_vis.rst
```diff
@@ -25,33 +26,34 @@ documents. The yellowbrick visualizer then plots the scatter plot,
 coloring by cluster or by class, or neither if a structural analysis is
 required.
 
-.. code:: ipython3
```
bbengfort: This was causing a document build error - no pygments for ipython3 so updated it while I had the chance.

docs/api/features/importances.py
```diff
@@ -46,6 +47,18 @@ def coef_(outpath):
     viz.poof(outpath=outpath)
 
 
+def stacked_coef_(outpath):
```
bbengfort: Generates the figure for the documentation.

docs/changelog.rst
```diff
@@ -20,6 +20,7 @@ Minor Changes:
     - Added polish to missing value visualizers code, tests, and documentation.
     - Improved RankD tests for better coverage.
     - Added quick method test for ``DispersionPlot`` visualizer.
+    - Extended FeatureImportances documentation and tests for stacked importances and added a warning when stack should be true.
```
bbengfort: Trying to update the changelog with new commits to make releases easier.

tests/test_features/test_importances.py
```diff
@@ -242,10 +242,12 @@ def test_multi_coefs(self):
...
-        visualizer = FeatureImportances(model)
-        visualizer.fit(
-            np.random.rand(100, len(np.mean(coefs, axis=0))), np.random.rand(100)
-        )
+        visualizer = FeatureImportances(model, stack=False)
+
+        with pytest.warns(YellowbrickWarning):
+            visualizer.fit(
+                np.random.rand(100, len(np.mean(coefs, axis=0))), np.random.rand(100)
+            )
```
bbengfort: Test the new warning functionality

tests/test_features/test_importances.py
```diff
@@ -256,12 +258,11 @@ def test_multi_coefs_stacked(self):
...
-        X_iris, y_iris = load_iris(True)
-        X_iris_pd = pd.DataFrame(X_iris, columns=['f1', 'f2', 'f3', 'f4'])
+        X, y = load_iris(True)
 
-        viz = FeatureImportances(LogisticRegression(), stack=True)
-        viz.fit(X_iris_pd, y_iris)
-        viz.poof()
+        viz = FeatureImportances(LogisticRegression(random_state=222), stack=True)
+        viz.fit(X, y)
+        viz.finalize()
```
bbengfort: Don't use pandas in this test to ensure complete test compatibility.

yellowbrick/features/importances.py
```diff
@@ -150,8 +151,11 @@ def fit(self, X, y=None, **kwargs):
         # therefore we flatten by taking the average by
         # column to get shape (n_features,)  (see LogisticRegression)
         if not self.stack and self.feature_importances_.ndim > 1:
-            self.feature_importances_ = np.mean(self.feature_importances_,
-                                                axis=0)
+            self.feature_importances_ = np.mean(self.feature_importances_, axis=0)
+            warnings.warn((
+                "detected multi-dimensional feature importances but stack=False, "
+                "using mean to aggregate them."
+            ), YellowbrickWarning)
```
bbengfort: This warning is issued because it is likely the user desires stack=True and should set it to avoid misinterpreting the feature importances.

docs/api/features/importances.py
lwgray: thanks for adding the "stacked image to documentation"

docs/api/features/importances.rst
```diff
@@ -112,6 +108,33 @@ title case our features for better readability:
...
+    X, y = data.data, data.target
+
+    viz = FeatureImportances(LogisticRegression(), ax=ax, stack=True, relative=False)
+    viz.fit(X, y)
+    viz.poof()
+
+
+.. image:: images/feature_importances_stacked.png
+
+
```
lwgray: thanks for adding stacked description/explanation and examples to importance page

docs/changelog.rst
lwgray: really appreciate you for doing this..I have to make sure I do it also

tests/test_features/test_importances.py
lwgray: I thought we should test for pandas per Audit

bbengfort: You're right of course, in this case we do have a test for Pandas for this visualizer on [L271](https://github.com/DistrictDataLabs/yellowbrick/blob/08ee4e41bdb25bdc24f32904dc31f771676f0362/tests/test_features/test_importances.py#L271).  The pandas test is a regression test to make sure that our tests work with a data frame and a non-random example. We don't have to have a matrix of pandas/non-pandas for every test, and in fact, there should really be only 1 pandas test that doesn't test anything else per visualizer since it injects more complexity into the test. 

In this case I took off pandas because we already had a pandas test and because the test didn't have a `@skipif(pd is None, reason='pandas required for this test)`; Pandas didn't add anything to the test so it was removed!

bbengfort:
@lwgray thanks for the review!

tests/test_features/test_importances.py
lwgray: understood...I totally missed L271

bbengfort: No worries - I'm glad you asked!

PR #679, opened 2018-12-29
Bugfix: RFECV step display bug

### LINKED ISSUE #664: step: removing 5 features in each iteration
Using RFECV, the graph shows best features with one increment in each iteration.
I want to show each iteration should work with an increment of 5 features.
Example: score  should be plotted against x-axis of [5,10,15,20,25].
I used step=5, but its not working.

bbengfort:
The RFECV visualizer had a bug when the hyperparameter step > 1. The
step was correctly passed to the internal RFE estimator, which removed
that number of features per iteration, however the feature subsets that
were tried for cross-validation did not match the step resulting in a
figure that looked like no step was actually applied. This patch fixes
the bug and creates a test to ensure this works correctly.

To manage the feature selection subspace, a new learned attribute,
`n_feature_subsets_` was added.

Fixes #664


bbengfort:
This is from #664 -- added here to make the review easier. 

Here is the code to reproduce the bug (note the step=5 in the RFECV constructor):

```python
from sklearn.datasets import make_classification
from sklearn.linear_model import LogisticRegression
from yellowbrick.features import RFECV 

X, y = make_classification(
    n_samples=200, n_features=30, n_informative=18, n_redundant=6,
    n_repeated=0, n_classes=8, n_clusters_per_class=1, random_state=0
)

oz = RFECV(LogisticRegression(), step=5).fit(X, y)
oz.poof()
```

This resulted in the following image:

example_step_bug_rfecv

Although step is passed to the `RFE` estimator to eliminate that many features each time, the CV methods are still using a step of 1. I fixed this and now the output looks as follows:

example_stepped_rfecv

tests/test_features/test_rfecv.py
```diff
@@ -124,7 +126,7 @@ def test_quick_method(self):
         model = LogisticRegression()
         X, y = self.dataset
 
-        ax = rfecv(model, X, y, step=3, cv=cv, scoring='f1_weighted')
+        ax = rfecv(model, X, y, step=2, cv=cv, scoring='f1_weighted')
```
bbengfort: made the quickmethod step a little less agressive.

tests/test_features/test_rfecv.py
```diff
@@ -77,7 +79,7 @@ def test_fit(self, mock_draw):
         X, y = self.dataset
         params = (
             "n_features_", "support_", "ranking_",
-            "cv_scores_", "rfe_estimator_",
+            "cv_scores_", "rfe_estimator_", "n_feature_subsets_"
```
bbengfort: Test that the new learned parameter is correctly handled.

tests/test_features/test_rfecv.py
```diff
@@ -156,11 +158,27 @@ def test_pandas_integration(self):
...
-    def test_valid_step(self):
+    def test_invalid_step(self):
         """
         Test step hyperparam validation
         """
         # TODO: parametrize when unittest is removed
-        with pytest.raises(YellowbrickValueError):
-            oz = RFECV(SVC(kernel="lnear"), step=-1)
+        with pytest.raises(YellowbrickValueError, match="step must be >0"):
+            oz = RFECV(SVC(kernel="linear"), step=-1)
```
bbengfort: Typo was probably causing this test to pass incorrectly!

tests/test_features/test_rfecv.py
```diff
@@ -156,11 +158,27 @@ def test_pandas_integration(self):
...
             oz.fit(self.dataset.X, self.dataset.y)
+
+    def test_rfecv_step(self):
```
bbengfort: Added a step-specific test to specifically catch this bug.

yellowbrick/features/rfecv.py
```diff
@@ -172,7 +176,7 @@ def fit(self, X, y=None):
 
         # Create the RFE model
         rfe = RFE(self.estimator, step=step)
-        n_feature_subsets = np.arange(1, n_features+1)
+        self.n_feature_subsets_ = np.arange(1, n_features+step, step)
```
bbengfort: This is essentially the bugfix, while the step param was correctly passed to rfe in L178, the cross-validation portion of the code still stepped by 1. Note that we had to use RFE instead of RFECV so that we could draw the variance of cross-validation on the figure. 

I am a little squirrely about `np.arange(1, n_features+step, step)`, particularly the `n_features+step` part ... this is required so that the maximum number of features can be selected by `RFE`, but it will actually pass `n_features_to_select > n_features` to `RFE`. However, this does not seem to cause a problem with `RFE` ... so 🤷‍♂️ ? I'm open to better suggestions. 

Also note that a new attribute `n_feature_subsets_` also had to be created; this is because the `step` param is parsed and validated during `fit()` and nowhere else, `draw` requires the number of features per CV iteration for the x-axis, so I stored it here. It also seems like information the user might want to get out as well, so I thought it was valuable.

yellowbrick/features/rfecv.py
```diff
@@ -214,7 +218,7 @@ def draw(self, **kwargs):
         Renders the rfecv curve.
         """
         # Compute the curves
-        x = np.arange(1, len(self.cv_scores_)+1)
+        x = self.n_feature_subsets_
```
bbengfort: This fixes the drawing bug.

bbengfort:
Thanks @wagner2010 !

PR #680, opened 2018-12-30
fixes `resolve_colors` bug in TSNEVisualizer identified by jerome massot

rebeccabilbro:
Jerome Massot reported in #658 that there is a bug in the `TSNEVisualizer` that means that colors that are passed in on instantiation do not affect the colors of the plot. Documentation of the bug can be found in [this notebook](https://github.com/rebeccabilbro/yellowbrick/blob/jerome-tsne-colors-fix/examples/rebeccabilbro/tsne_resolve_colors.ipynb). This pull request corrects `self.color` to `self.colors` in the `resolve_colors` function call inside the `draw` method to fix the bug. Thanks to @jeromemassot for identifying the bug and its solution!


rebeccabilbro:
Sure thing @bbengfort - I've just added some tests for both; let me know if those are what you had in mind.

bbengfort:
@rebeccabilbro those look like excellent tests - above and beyond what I was suggesting! Thanks for adding them! 

Just one quick question - the new tSNE test calls `fit()` twice, which is one of our most intensive operations in the library; does this add noticeable time to the tests? If so, I'd recommend that in the dataset generation, create 25 samples with 10 features, 10 informative and 0 redundant to reduce the compute time. If not, please feel free to merge!

rebeccabilbro:
@bbengfort, some quick stats on the test runtimes on my macbook pro w/ Amtrak wifi, FWIW:

TSNE + UMAP no color check tests - 134.86 seconds
TSNE only w/ color test, first call to fit - 134.76 seconds
TSNE only w/ color test, both calls to fit - 135.63 seconds
TSNE + UMAP w/ color test, first call to fit - 138.70 seconds
TSNE + UMAP w/ color test, both calls to fit - 136.86 seconds

Sorry, only tests I had time to run before getting off train. Looks like UMAP is the main slowdown

bbengfort:
Looks like no significant difference to me! I'm going to merge since you're on the train, if that's ok. Thanks for testing the test performance!

rebeccabilbro:
Thanks @bbengfort and @jeromemassot!

PR #688, opened 2019-01-18
[MRG] Bug fix for PR curve multiclasses labels

### LINKED ISSUE #601: Precision-Recall Curve does not show class labels
**Describe the bug**

When labels are passed into the `PrecisionRecallCurve` the visualization is not drawn correctly.

**To Reproduce**

```python
from yellowbrick.classifier import PrecisionRecallCurve 
from yellowbrick.dataset import load_game 

from sklearn.preprocessing import LabelEncoder 
from sklearn.naive_bayes import MultinomialNB 

# Load the dataset and label encode the target 
data = load_game() 
X = data.iloc[:, data.columns != 'outcome']
y = LabelEncoder().fit_transform(data['outcome'])

# Create train test splits 
X_train, X_test, y_train, y_test = tts(X, y, test_size=0.2, shuffle=True)

oz = PrecisionRecallCurve(
    MultinomialNB(), per_class=True, iso_f1_curves=True, fill_area=False, 
    micro=False, classes=["loss", "draw", "win"]
)
oz.fit(X_train, y_train)
oz.score(X_test, y_test)
oz.poof()
```

**Dataset**

I used the game multi-label dataset from the UCI Machine Learning repository, as wrangled by the yellowbrick datasets module.

**Expected behavior**

When the target, `y` is label encoded (e.g. via the `LabelEncoder`) to classes 0, 1, and 2 and class names are passed in via the `classes` param of the visualizer, the legend should display the class names. However, in this case the visualization does not appear at all:

figure_1

**Traceback**

No exception is raised, however, the following warning is issued:

```
/Users/benjamin/.pyenv/versions/3.6.2/envs/yellowbrick3/lib/python3.6/site-packages/numpy/lib/arraysetops.py:522: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  mask |= (ar1 == a)
/Users/benjamin/.pyenv/versions/3.6.2/envs/yellowbrick3/lib/python3.6/site-packages/sklearn/metrics/ranking.py:444: RuntimeWarning: invalid value encountered in true_divide
  recall = tps / tps[-1]
objc[46460]: Class FIFinderSyncExtensionHost is implemented in both /System/Library/PrivateFrameworks/FinderKit.framework/Versions/A/FinderKit (0x7fffb473cc90) and /System/Library/PrivateFrameworks/FileProvider.framework/OverrideBundles/FinderSyncCollaborationFileProviderOverride.bundle/Contents/MacOS/FinderSyncCollaborationFileProviderOverride (0x12814ecd8). One of the two will be used. Which one is undefined.
```

 - OS: macOS High Sierra 10.13.6
 - Python Version 3.6.2
 - Yellowbrick Version 0.9 develop 

**Additional context**

This was a known bug during development but was left in as a rare case.

zjpoh:
Fixes #601. This bug is arises when the `classes` provided when initializing `PrecisionRecallCurve` does not exist in `y`. This problem can arise because `y` is supposed to be encoded for model training while `classes` contains the names of the classes.

In both `fit` and `score` functions in `PrecisionRecallCurve`, `label_binarize` is called to binarize `y`. `label_binarize` uses `np.unique(y)` in `fit` while in `score`, `label_binarize` uses `classes` provided when initializing `PrecisionRecallCurve`. A more consistent way to binarize `y` is to binarize `y` using the same classes. 

Hence, in this PR, I added a variable that stores the classes used to binarize `y` in fit and reuse the same class in the score function.


zjpoh:
Travis complains about image not close. I'm wondering if there is a way to take a look at the actual images generated by Travis? The baseline image and the actual image generated on my computer looks very similar. Thanks.

zjpoh:
Also AppVeyor is complaining about 
```
self = <tests.test_classifier.test_confusion_matrix.ConfusionMatrixTests testMethod=test_score_returns_score>
    def test_score_returns_score(self):
        """
        Test that ConfusionMatrix score() returns a score between 0 and 1
        """
        data = self.load_data("occupancy")
        X = data[[
            "temperature", "relative_humidity", "light", "C02", "humidity"
        ]]
    
        y = data['occupancy']
    
        # Convert X to an ndarray
>       X = X.copy().view((float, len(X.dtype.names)))
E       ValueError: Changing the dtype to a subarray type is only supported if the total itemsize is unchanged
tests\test_classifier\test_confusion_matrix.py:349: ValueError
```
I'm unable to reproduce this on my local machine.

bbengfort:
Hi @zjpoh - sorry it's taken a while to respond; yes these tests are a problem with CI and I think they're related to an update in numpy and pyflakes. I've created issue #699 to deal with this, and once it's fixed I can review your PR. Thank you so much for your patience!

tests/test_classifier/test_prcurve.py
```diff
@@ -238,6 +238,44 @@ def test_multiclass_probability(self):
...
+    def test_multiclass_probability_with_class_labels(self):
+        """
+        Visual similarity of multiclass classifier with predict_proba function
+        with class labels
+        """
+        # Create and fit the visualizer
+        oz = PrecisionRecallCurve(
+            GaussianNB(), per_class=True, micro=False, fill_area=False,
+            iso_f1_curves=True, ap_score=False,
+            classes=["a", "b", "c", "d", "e", "f"]
```
bbengfort: Ok, so it's been a while since I've looked at this bug report and I just want to check to make sure I'm on the same page. By adding in this `classes=` argument, we're expecting the lines to be drawn, which they were not previously. The bug did not affect the legend correct? 

The reason I ask is that in the comment I make about the tolerance below, I suggest removing `finalize()`, which would remove the legend which would not test case of the legend being correctly set. Alternatives would be to directly test the legend object without assert image similarity, but that could be tricky; or to simply increase the tolerance so that the legend was tested. However, if the legend was part of the bug and removing finalize solves things, then perhaps we should just go that route.

yellowbrick/classifier/prcurve.py
```diff
@@ -173,7 +173,8 @@ def fit(self, X, y=None):
             self.estimator = OneVsRestClassifier(self.estimator)
 
             # Use label_binarize to create multi-label ouptut for OneVsRestClassifier
-            Y = label_binarize(y, classes=np.unique(y))
+            self.classes_from_y = np.unique(y)
```
bbengfort: This is a good solution, thank you! In order to remove confusion with `classes_`, might I propose that we make this `_classes_from_y` or `_unique_y` or `_target_labels`?

rebeccabilbro:
Hi there @zjpoh! We’ve been having some difficult with our CI and wanted to ensure that your PR was up-to-date with our fixes, so we’ve merged in our main develop branch to the PR. Please make sure that you do a pull before making any changes!

tests/test_classifier/test_prcurve.py
```diff
@@ -238,6 +238,44 @@ def test_multiclass_probability(self):
         tol = 6.6 if sys.platform == 'win32' else 1.0 # fails with RMSE 6.583 on AppVeyor
         self.assert_images_similar(oz, tol=tol)
 
+    def test_multiclass_probability_with_class_labels(self):
+        """
+        Visual similarity of multiclass classifier with predict_proba function
+        with class labels
```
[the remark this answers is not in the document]
zjpoh: Got it. I will make that shorter.

tests/test_classifier/test_prcurve.py
zjpoh: The only changes to the legend are the values of the area. 

I think at the minimum we should test the text in the legend because we want to make sure that `classes` and the calculated areas in the legend are correct.

Let me try to test for the text in the legend before increasing the tolerance.

yellowbrick/classifier/prcurve.py
zjpoh: Sure.

tests/test_classifier/test_prcurve.py
```diff
@@ -263,55 +329,3 @@ def test_no_scoring_function(self):
         oz = PrecisionRecallCurve(FakeClassifier())
         with pytest.raises(ModelError, match="requires .* predict_proba or decision_function"):
             oz._get_y_scores(self.binary.X.train)
-
-    def test_custom_iso_f1_scores(self):
```
bbengfort: @zjpoh Is there a reason these tests were removed? Was this just a merge conflict thing?

zjpoh: Interesting.... I do not recall remove this.

zjpoh:
As you suggested, the diff is due to the text. So, I remove all the text from the figure and perform separate checks on the texts.

tests/test_classifier/test_prcurve.py
bbengfort: Probably it was the automatic merge - no worries we caught it!

tests/test_classifier/test_prcurve.py
```diff
@@ -29,7 +30,6 @@
 from sklearn.datasets import make_regression
 from sklearn.tree import DecisionTreeClassifier
 from sklearn.linear_model import RidgeClassifier
-from sklearn.model_selection import train_test_split as tts
```
bbengfort: This might have to also be added back to get the tests to pass ... sorry!

tests/test_classifier/test_prcurve.py
```diff
@@ -239,6 +239,72 @@ def test_multiclass_probability(self):
...
+            "PR for class b (area=0.36)",
+            "PR for class c (area=0.44)",
+            "PR for class d (area=0.52)",
+            "PR for class e (area=0.37)",
+            "PR for class f (area=0.49)",
+        ]
+        assert oz_legend_txt == expected_legend_txt
+        handles, _ = oz.ax.get_legend_handles_labels()
+        empty_labels = [""] * len(handles)
+        oz.ax.legend(handles=handles, labels=empty_labels, loc='lower left',
```
bbengfort: I've got to hand it to you, this is pretty extensive testing of the text in the legend and labels - very nice work!

tests/test_classifier/test_prcurve.py
bbengfort: Or maybe not ... looks like the tests did pass; I think that's probably because of an `import *` somewhere.

Issue #683, opened 2019-01-02
Make yellowbrick.datasets prime time


bbengfort:
In #535 we implemented a new `yellowbrick.datasets` module that dramatically reduces the amount of work users have to do to access our examples and reproduce our visualizations. These datasets also will enhance our testing with more robust loading and dataset management. 

In this epic, we make the `yellowbrick.datasets` module primetime, using it in all of our documentation and testing for maximum effect; replacing all other data related tools that were used in the past.

PR #695, opened 2019-01-22
Update code to import data in regression notebook

Kautumn06:
Since the recent overhaul of Yellowbrick's datasets module in PR #535, the code to import the sample datasets and extract the X and y data has changed. Here, I updated the code to import the concrete dataset in our notebook of regression examples.

Previously, here is how the `concrete` dataset had been loaded in `regression.ipynb`:
datasets

And here is how it is now:
```
# Use Yellowbrick to load the concrete dataset
X, y = load_concrete()

# Create the train and test data 
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=42)
```

See also: #683


bbengfort:
@Kautumn06 thank you for updating the notebook! Once we get #699 and the CI tests passing, I'll take a look and get this merged in. Thank you for your patience!

bbengfort:
Right now we're waiting to merge in #712 so that we can update the tests; if we merge this into develop now the tests will fail. Should be able to merge these and other PRs tonight; thanks again for your patience!

Issue #682, opened 2019-01-02
Use yellowbrick.datasets in tests


bbengfort:
In #535 we implemented new `Dataset` and `Corpus` datasets. These datasets should replace the current `tests.datasets` to ensure standardization and that our examples work.

Subtasks:

- [x] Rename tests.fixtures.TestDataset to tests.fixtures.Dataset (prefix warnings in runner) 
- [x] Remove `tests/datasets.py`
- [x] Remove all `DatasetMixin` extensions
- [x] ~Create fixtures for synthetic datasets~ (moved to #910)
- [x] Replace `self.load_concrete` with `yellowbrick.datasets.load_concrete`, etc. 
- [x] Remove `dataset.py` and `DatasetMixin` classes
- [ ] In numpy structured array tests use `Dataset.to_numpy` 
- [ ] in pandas tests use `Dataset.to_pandas` 
- [x] ~Update tests to use `windows_tol` instead of conditional tolerances~ (moved to #911)
- [x] ~update contributor guide with use of datasets/fixtures in tests (both random and real)~ (moved to #441)

This task will likely affect/touch _most_ of our tests. I expect that many baseline images will also have to be replaced because the datasets may not be exactly the same.

There has been some question about the use of PyTest fixtures (which are currently primarily used for the randomly generated test data sets) to protect our tests from modifications to the data. This adds a little bit more complication to the test suite but protects us from future changes. Throughout this PR we should inspect our use of data and fixtures and determine a consistent method for using both randomly generated data and yellowbrick datasets. Then add this methodology to the contributor documentation.

This also might be future-proofing against future PyTest releases (#714). 

See #746 and #735 

**UPDATE**: Based on some of our discussions we feel that datasets in tests should be _read only_. The scenario we imagined where datasets might be modified is to "corrupt" them to add missing values or lower their F1 score for the purposes of better tests. In this case, however, we should create a new dataset in the dataset module or use a randomly generated dataset. Note also that Yellowbrick visualizers that are transformers should also not modify the data but should create a new arrray to return.

PR #705, opened 2019-01-30
Fix Ci Tests

### LINKED ISSUE #699: CI Tests are Failing
It looks like recent updates to a couple of our dependencies have broken our CI tests on both Travis and AppVeyor. This needs to be resolved so that we can review the PRs currently in the backlog. 

Travis: 

The Travis failure appears to be a pyflakes related issues:

```
___________ pyflakes-check(ignoring ImportStarUsage ImportStarUsed) ____________
/home/travis/build/DistrictDataLabs/yellowbrick/tests/test_classifier/test_threshold.py:298: IsLiteral
use ==/!= to compare str, bytes, and int literals
```

AppVeyor:

The AppVeyor tests seem to be failing with a change to Numpy with `X.copy().view((float, len(X.dtype.names)))` now failing:

```
        # Convert X to an ndarray
>       X = X.copy().view((float, len(X.dtype.names)))
E       ValueError: Changing the dtype to a subarray type is only supported if the total itemsize is unchanged
```

This occurs in the following test files:

- tests\test_classifier\test_confusion_matrix.py:349
- tests\test_contrib\test_scatter.py:150
- tests\test_contrib\test_scatter.py:237
- tests\test_contrib\test_scatter.py:188

bbengfort:
Addresses the CI test failures that were occurring on both Travis and AppVeyor. 

- Fix pyflakes error on Travis-CI tests
- Fix ValueError on numpy tests on AppVeyor

Tests affected:

- tests/test_classifier/test_confusion_matrix.py
- tests/test_contrib/test_scatter.py
- tests/test_contrib/test_classifier/test_boundaries.py
- tests/test_features/test_importances.py

Notes on the `ValueError: Changing the dtype to a subarray type is only supported if the total itemsize is unchanged`: This was brought up by me and others on Stack Overflow concerning numpy v1.14 and later. To fix the tests, I switched over to simply using the `yellowbrick.datasets` module so that structured arrays are completely out of the equation. This conforms with #682 but might reduce our coverage in some places where we were specifically testing structured arrays. I ensured that we were still using structured arrays with the `Dataset.to_numpy` method, but no slicing or view functions occurred. 

see also: https://stackoverflow.com/questions/48267058/using-np-view-with-changes-to-structured-arrays-in-numpy-1-14

This PR also attempts to handle some of the massive number of warnings we get with the newer libraries, e.g. ignoring `sklearn.exceptions.ConvergenceWarning` in setup.cfg. We can handle a number of warnings by removing `poof()` from our tests and instead calling `finalize()` directly.

Fixes #699


bbengfort:
@rebeccabilbro hoping the CI tests pass, if they do would you mind doing a review so we can repair the current PRs? If the CI tests don't pass then I'll keep at it, but should be done tonight.

tests/test_contrib/test_classifier/test_boundaries.py
```diff
@@ -320,9 +326,7 @@ def test_fit_draw_poof(self):
         viz.draw.assert_called_once_with(X_two_cols, y)
         viz.poof.assert_called_once_with()
 
-    @pytest.mark.xfail(
-        sys.platform == 'win32', reason="images not close on windows"
-    )
+    @pytest.mark.xfail(reason="numpy structured arrays have changed since v1.14")
```
bbengfort: Couldn't find a solution for this test, so just marked as xfail since this is a contrib package. If we move this back into our classifier module, we'll either have to find a solution for structured arrays or simply not support them.

tests/test_contrib/test_scatter.py
```diff
@@ -201,22 +184,15 @@ def test_integrated_scatter_with_pandas(self):
...
+        # Load the data from the fixture
+        X, y = load_occupancy(return_dataset=True).to_pandas()
 
         # Test the visualizer
-        features = ["temperature", "relative_humidity"]
+        features = ["temperature", "relative humidity"]
         visualizer = ScatterViz(features=features)
         visualizer.fit_transform_poof(X, y)
 
+    @pytest.mark.xfail(reason="numpy structured arrays have changed since v1.14")
```
bbengfort: Couldn't find a solution for this test, so just marked as xfail since this is a contrib package. If we move this back into main as a mixin, we'll either have to find a solution for structured arrays or simply not support them.

tests/test_features/test_importances.py
```diff
@@ -93,22 +86,19 @@ def test_integration_coef(self):
...
-        feats = ['cement','slag','ash','water','splast','coarse','fine','age']
-
-        # Create X and y datasets as numpy arrays
-        X = concrete[feats].copy()
-        X = X.view((float, len(X.dtype.names)))
-        y = concrete['strength']
+        # Load the test dataset
+        dataset = load_concrete(return_dataset=True)
+        X, y = dataset.to_numpy()
+        features = dataset.meta["features"]
```
bbengfort: Whoo! New datasets module ftw!

tests/test_features/test_importances.py
```diff
@@ -65,23 +65,16 @@ def test_integration_feature_importances(self):
...
+        # Load the test dataset
+        X, y = load_occupancy(return_dataset=True).to_numpy()
 
         fig = plt.figure()
         ax = fig.add_subplot()
 
         clf = GradientBoostingClassifier(random_state=42)
         viz = FeatureImportances(clf, ax=ax)
         viz.fit(X, y)
-        viz.poof()
```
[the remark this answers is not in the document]
rebeccabilbro: Worth opening up a follow-up issue? Or just rolling this into an existing one?

tests/test_features/test_importances.py
rebeccabilbro: Agreed! I love how easy and readable the new oneliner style is, e.g. `X, y = load_dataset(return_dataset=True).to_numpy()`!

tests/test_features/test_importances.py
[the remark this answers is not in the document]
bbengfort: Perhaps we can add it to it to #706?

bbengfort:
@rebeccabilbro actually that's a good point about the documentation, I'll create a new story to replace `poof()` with `finalize()` in the tests and update the documentation to note that warning. 

@lwgray are we far enough along in the audit that we shouldn't add this to the tests checklist?

lwgray:
We aren't far enough along in the audit...  So we are checking to make sure all tests utilize `finalize`, correct ???

bbengfort:
@lwgray awesome, thanks! Yes, the audit will include two new items:

1. has tests that use `poof()` before `assert_image_similarity`
2. has tests that use `finalize()` before `assert_image_similarity` (yes, no, mixed)

The first audit item has an associated action item, change to `finalize`. 

The second audit item is only informational: if a visualizer doesn't have tests that use `finalize` then it may not have complete coverage, e.g. matplotlib actions that get called in finalize aren't being tested; on the other hand, calls to `finalize()` may make the image similarity tests impossible or not available on Windows. It would be interested to note which tests definitely do, which don't, and which do both.

PR #718, opened 2019-02-02
[WIP] Adds Effect Plot visualizer for regressors

### LINKED ISSUE #604: Effect Plot for Linear Models
Would love to have an Effect Plot for aiding with interpreting linear models. I realize that a feature importance plot does some of this. An effect plot shows the weights as a bar plot so you can see whether the impact is positive or negative and also how large the variance is.

**Examples**

There is a great example here 
https://christophm.github.io/interpretable-ml-book/limo.html#visual-parameter-interpretation

My scouring has not turned up any Python code to generate this plot in the wild.

souravsingh:
Fixes #604 by adding a visualizer for regressors. Would need help in parameter setup.

### DIFF
--- a/yellowbrick/regressor/effect.py
+++ b/yellowbrick/regressor/effect.py
@@ -1,0 +1,51 @@
+"""
+Regressor visualizers that score residuals: prediction vs. actual data.
+"""
+
+##########################################################################
+## Imports
+##########################################################################
+
+
+import matplotlib.pyplot as plt
+
+try:
+    # Only available in Matplotlib >= 2.0.2
+    from mpl_toolkits.axes_grid1 import make_axes_locatable
+except ImportError:
+    make_axes_locatable = None
+
+from sklearn.model_selection import train_test_split
+
+from .base import RegressionScoreVisualizer
+from ..draw import manual_legend
+from ..style.palettes import LINE_COLOR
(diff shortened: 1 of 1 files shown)


bbengfort:
@souravsingh thank you for your interest in developing this visualizer for Yellowbrick! It looks like you've got a good start going, and I would certainly be happy to discuss the parameter setup, but the easiest way to do that would be if we could more specifically think through how this would be used. 

Therefore, before we get started on the visualizer, perhaps you could create a simple example as a [gist](https://gist.github.com/) and we could explore it together? This wouldn't be a full visulalizer, but rather just straight sklearn and matplotlib code on a sample dataset (e.g. `load_concrete` or `load_bikeshare` to see how it turns out. Once we see the example it will be much easier to suss out the requirements for the full visualizer. 

Note also that there is a non-python example link in #604 that you could look at for ideas.

bbengfort:
@souravsingh again, before getting started on this visualizer, I would strongly encourage you to come up with a simple example either in a notebook or in a simple Python script!

rebeccabilbro:
Hi there @souravsingh &mdash; since the conversation here has gone a bit stale, I'm going to go ahead and close this PR out. Thanks again for your interest in Yellowbrick and maybe we'll see you again in a future PR!

Issue #716, opened 2019-02-01
JointPlotVisualizer not poofing


Naba7:
OS:Linux 18.04LTS
Python:3.6.8 Anaconda inc.
Yellowbrick:0.9

So,as said by @rebeccabilbro  I tried to use fit_transform instead of fit in the JointPlotVisualizer,but no change .I am tracing the code along with my terminal output;
CODE:
```python
import pandas as pd
data=pd.read_csv('bikeshare.csv')
X=data[["season","month","hour","holiday","weekday","workingday","weather","temp","feelslike","humidity","windspeed"]]
y=data["riders"]


from yellowbrick.features import JointPlotVisualizer

visualizer=JointPlotVisualizer(feature='temp',target='feelslike')
visualizer.fit_transform(X["temp"],X["feelslike"])
visualizer.poof()
```

OUTPUT:
```
chad7@superuser:~$ cd Desktop/YellowBrick/
chad7@superuser:~/Desktop/YellowBrick$ python3 yb1.py
chad7@superuser:~/Desktop/YellowBrick$ python3 yb1.py
chad7@superuser:~/Desktop/YellowBrick$ 
```
No,changes no figures showing up.

I know I am increasing your tasks with my questions,Sorry!

rebeccabilbro:
Hi again @Naba7 - please try [providing an outpath for your visualizer](http://www.scikit-yb.org/en/latest/faq.html#how-can-i-save-a-yellowbrick-plot):

```python
import pandas as pd
from yellowbrick.features import JointPlotVisualizer

data=pd.read_csv('bikeshare.csv')

X=data[["season","month","hour","holiday","weekday","workingday","weather","temp","feelslike","humidity","windspeed"]]
y=data["riders"]

visualizer=JointPlotVisualizer(feature='temp',target='feelslike')
visualizer.fit_transform(X["temp"],X["feelslike"])
visualizer.poof(outpath="nabas_jointplot.png") # <= provide a path to store the image
```

If what you're looking for is more immediate/interactive viewing of plots, you might try using yellowbrick inside a Jupyter notebook.

Out of curiosity, could you please run the following in a Python REPL and let us know what you see?
```bash
>>> import matplotlib
>>> if not matplotlib.get_backend() == 'Agg':
...     matplotlib.use('Agg', force=True)
>>> print('backend:', matplotlib.get_backend())
```

Naba7:
I had saved it already in outpath and tried,but no success.
[...]

Issue #707, opened 2019-01-30
Manage test warnings: 'Agg backend'


bbengfort:
Updates in our dependencies have created a lot of `PendingDeprecation` and `UserWarnings` in our tests, which we should address by ignoring them in the test spec or replacing the code that generates the warning with an appropriate expression that does not. Currently this involves the following actions:

- [x] ignore sklearn.exceptions.ConvergenceWarning in setup.cfg
- [x] replace `poof()` with `finalize()` in tests to remove matplotlib UserWarning about 'Agg' backend
- [x] document image comparison tests to reflect use of `poof()` and `finalize()` in tests
- [x] create issues to fix other warnings issued in tests

Note - warnings are generated in both the doc build and the tests; we should determine what all the unique warnings are and try to suppress them. 

See #705 for related discussion.

bbengfort:
This will be fixed in #1187

Issue #605, opened 2018-09-11
JointPlotVisualizer produces different visualizations depending on how the figure and axes objects are defined


pdamodaran:
**Describe the bug**
JointPlotVisualizer produces different visualizations depending on how the figure and axes objects are defined.  

If the figure and axes are defined in the following manner:

_, ax = plt.subplots()
visualizer = JointPlotVisualizer(ax=ax, feature=feature, target=target, joint_plot="hex")

The results are as follows:

1) feature and target labels are displayed in the image
2) the tick labels are displayed only on the left and bottom axes
3) there are issues with the tick label values

On the other hand, if the figure and axes are defined as such:

fig = plt.figure()
ax = fig.add_subplot()

The results are as follows:

1) the feature and target labels are not displayed
2) the tick labels show up in all the axes
3) the tick label values are displayed with more meaningful values

Refer to the following jupyter notebook to see a demonstration of the differences:

https://github.com/pdamodaran/yellowbrick/blob/develop/examples/pdamodaran/JointPlot_Examples.ipynb

NOTE: any fixes to this issue should be considered in conjunction with the following issues:

https://github.com/DistrictDataLabs/yellowbrick/issues/434
https://github.com/DistrictDataLabs/yellowbrick/issues/214

Issue #214, opened 2017-05-05
JointPlot updates


bbengfort:
Update the joint plot as follows:

- [x] Rename `JointPlot` and alias `JointPlotVisualizer` 
- [x] Add pearson correlation as label 
- [x] Allow 2 feature joint plot 
- [ ] Use target as "hue" for classes 
- [ ] Multiple regressions per each class

bbengfort:
Consider the first two things as quick and easy things for new contributors.

PR #719, opened 2019-02-02
Adds plot directives to quickstart

bbengfort:
This PR was primarily intended to add plot directives to the quickstart guide, ensuring that the images in the tutorial were always up to date with the library. The quickstart guide separates the code blocks from the plot directives so that the code is a linear narrative rather than the verbosity required for each independent code block. This duplicates the code a little bit but makes it more readable.

To ensure that the code is correct I also created a notebook examples/walkthrough.ipynb with the code from the quickstart.

Along the way there were some bugs with jointplot and rankd that I also fixed. The JointPlotVisualizer needs some more work but it is now stable.

Additionally, there is a commit to issue warning in poof without ax. Previously, in poof() if the visualizer didn't have an axes object it would exit silently. This made it hard to find bugs. Now instead of exiting we simply issue a warning and carry on.


docs/quickstart.rst
```diff
@@ -105,22 +111,35 @@ These quick functions give you slightly less control over the machine learning w
...
-After downloading the dataset and unzipping it in your current working directory, we can load our data as follows:
+We can load our data using the ``yellowbrick.datasets`` module as follows:
 
 .. code-block:: python
 
     import pandas as pd
+    from yellowbrick.datasets import load_bikeshare
+
+    X, y = load_bikeshare()
+    print(X.head())
```
bbengfort: Using the new datasets module ensures that any changes to the datasets are reflected in the docs. Unfortunately, we do lose the nice CSV loading in Pandas which makes the workflow more realistic. However, I think it is more important that the quickstart code works correctly.

docs/quickstart.rst
```diff
@@ -105,22 +111,35 @@ These quick functions give you slightly less control over the machine learning w
 Walkthrough
 -----------
```
bbengfort: Created examples/walkthrough.ipynb to ensure that this code works correctly. It might be nice to put a binder link here in the future.

docs/quickstart.rst
```diff
@@ -132,7 +151,19 @@ The machine learning workflow is the art of creating *model selection triples*,
    visualizer.fit_transform(X)
    visualizer.poof()
 
-.. image:: images/quickstart/bikeshare_rank2d.png
+.. plot::
+    :include-source: False
+    :alt: Rank2D of Bikeshare Features
```
bbengfort: As you can see, only the image was replaced and by using `:include-source: False` the image is generated exactly as below while allowing us to have `code-block` examples above.

docs/quickstart.rst
```diff
@@ -143,10 +174,20 @@ In this chart, we see that the features ``temp`` and ``feelslike`` have a strong
     from yellowbrick.features import JointPlotVisualizer
 
     visualizer = JointPlotVisualizer(feature='temp', target='feelslike')
-    visualizer.fit(X['temp'], X['feelslike'])
+    visualizer.fit_transform(X['temp'], X['feelslike'])
```
bbengfort: Addresses #716

examples/walkthrough.ipynb
```diff
@@ -0,0 +1,135 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Walkthrough\n",
+    "\n",
+    "This notebook contains the code for the [walkthrough in the quickstart guide](http://www.scikit-yb.org/en/latest/quickstart.html#walkthrough). We've purposefully omitted the text of the guide so that you can follow along in code using this notebook as a template! The scikit-yb developers also use this notebook to verify that the quickstart code is correct, so if this code doesn't match what's on the guide, please leave us a note on our [GitHub Issues](https://github.com/DistrictDataLabs/yellowbrick/issues)."
```
bbengfort: Contains code only, no text.

docs/quickstart.rst
lwgray: good solution!

tests/test_base.py
```diff
@@ -51,6 +59,26 @@ def test_visualizer_ax_property(self):
         assert viz._ax == "foo"
         assert viz.ax == "foo"
 
+    def test_size_property(self):
```
bbengfort: This was just moved from below to organize it with the other property tests.

tests/test_base.py
```diff
@@ -73,31 +101,58 @@ def test_finalize_interface(self):
         viz = Visualizer()
         assert viz.finalize() is viz.ax
 
-    def test_size_property(self):
+    @patch("yellowbrick.base.plt")
+    def test_poof_show_interface(self, mock_plt):
         """
-        Test the size property on the base Visualizer
+        Test poof calls plt.show and other figure finalization correctly
```
bbengfort: A new test to ensure poof is called correctly

tests/test_base.py
```diff
@@ -73,31 +101,58 @@ def test_finalize_interface(self):
...
-        assert all(viz.size) == all(fsize)
+        viz.finalize.assert_called_once_with()
+        mock_plt.show.assert_called_once_with()
+        mock_plt.savefig.assert_not_called()
 
-        viz.size = (1080, 720)
-        assert viz._size == (1080, 720)
-        assert viz.size == (1080, 720)
+    @patch("yellowbrick.base.plt")
+    def test_poof_savefig_interface(self, mock_plt):
```
bbengfort: a new test to ensure `poof(outpath="myfig.png")` is working correctly

tests/test_base.py
```diff
@@ -73,31 +101,58 @@ def test_finalize_interface(self):
...
+    def test_poof_warns(self, mock_plt):
+        """
+        Test poof issues a warning when no axes has been modified
+        """
+        class CustomVisualizer(Visualizer):
+            pass
+
+        with pytest.warns(YellowbrickWarning):
+            viz = CustomVisualizer()
+            viz.poof()
```
bbengfort: Test the new warning issue code.

tests/test_base.py
```diff
@@ -159,7 +214,7 @@ def test_draw_with_cols(self):
         grid = VisualizerGrid(visualizers, ncols=2)
 
         grid.fit(X, y)
-        grid.poof()
+        grid.poof() # poof is required here (do not replace with finalize)!
```
bbengfort: These notes are in reference to #707

yellowbrick/base.py
```diff
@@ -206,7 +209,14 @@ def poof(self, outpath=None, clear_figure=False, **kwargs):
...
         # Ensure that draw has been called
-        if self._ax is None: return
+        if self._ax is None:
+            warn_message = (
+                "{} does not have a reference to a matplotlib.Axes "
+                "the figure may not render as expected!"
+            )
+            warnings.warn(
+                warn_message.format(self.__class__.__name__), YellowbrickWarning
+            )
```
bbengfort: This is part of the JointPlot fix for #716 -- the `self._ax is None` check is to ensure that we don't poof and therefore lock a figure before draw is called. However, having it return silently meant that it was hard to figure out the joint plot bug. 

Right now I'm issuing a warning and continuing but I don't know if this is the best idea. 

Other options:

1. Raise an exception
2. Issue a warning and return 
3. Drop the check 

If we think this discussion is too big for this PR we can make it another issue.

yellowbrick/features/jointplot.py
```diff
@@ -145,7 +147,6 @@ def __init__(self, ax=None, feature=None, target=None,
         self.joint_args = joint_args
         self.xy_plot = xy_plot
         self.xy_args = xy_args
-        self.size = (size, size)
```
bbengfort: Setting size like this, although it kept the image square also interfered with the plot directive. So now the joint plot is 9x6 ... which is weird looking. I'll need to figure something out to fix it.

yellowbrick/features/jointplot.py
```diff
@@ -275,18 +277,18 @@ def draw_xy(self, X, y, **kwargs):
...
 
 
         if self.xy_plot == "hist":
             hist_bins = self.xy_args.pop("bins", 50)
             self.xy_args.setdefault("alpha", 0.4)
-            histcolor_x = self.xy_args.pop("histcolor_x", "#6897bb")
+            histcolor_x = self.xy_args.pop("histcolor_x", HISTCOLOR)
             self.x_ax.set_facecolor(facecolor_x)
-            histcolor_y = self.xy_args.pop("histcolor_y", "#6897bb")
+            histcolor_y = self.xy_args.pop("histcolor_y", HISTCOLOR)
```
bbengfort: Trying to make the color a constant so it could be changed in one place.

yellowbrick/features/rankd.py
```diff
@@ -174,7 +174,7 @@ def rank(self, X, algorithm=None):
 
         # Extract matrix from dataframe if necessary
         if is_dataframe(X):
-            X = X.as_matrix()
+            X = X.values
```
bbengfort: Fixes the deprecation warning issued in Rank2D call.

yellowbrick/base.py
lwgray: I am also unsure what the best course of action would be

yellowbrick/features/jointplot.py
lwgray: I think that was similar to what I about different dimensions over in [446](https://github.com/DistrictDataLabs/yellowbrick/pull/446#discussion_r252865766)

docs/quickstart.rst
```diff
@@ -105,22 +111,35 @@ These quick functions give you slightly less control over the machine learning w
 Walkthrough
 -----------
 
-Consider a regression analysis as a simple example of the use of visualizers in the machine learning workflow. Using a `bike sharing dataset <https://s3.amazonaws.com/ddl-data-lake/yellowbrick/bikeshare.zip>`_ based upon the one uploaded to the `UCI Machine Learning Repository <https://archive.ics.uci.edu/ml/datasets/bike+sharing+dataset>`_, we would like to predict the number of bikes rented in a given hour based on features like the season, weather, or if it's a holiday.
```
rebeccabilbro: Good catch!

yellowbrick/features/jointplot.py
rebeccabilbro: Maybe something like what we did w/ the residuals plot when we introduced the histogram? See also #214 and #605

yellowbrick/features/jointplot.py
rebeccabilbro: 👍

yellowbrick/features/rankd.py
rebeccabilbro: Ah, thanks!

yellowbrick/base.py
bbengfort: Ok, sounds good - let's all just be aware that if we see this warning reported in an issue or as we are working with YB, we should take action and create a new issue for it. There are two possibilities if this happens:

- user error, calling `poof()` without calling `fit()`
- developer/api error, as in joint plot 

I'm worried about the latter because I can't remember why we put this check in originally.

yellowbrick/features/jointplot.py
bbengfort: Yes, I think that's exactly the solution! I can get started on that but it's going to require a fairly significant overhaul of the JointPlot code, so I didn't want to do it in this PR (and that's what I was being blocked on by forgetting to create a feature branch ...).

bbengfort:
Thanks for the review @lwgray and @rebeccabilbro!

Issue #434, opened 2018-05-16
Add index selection to JointPlotVisualizer


bbengfort:
The joint plot visualizer shows the relationship between a single feature in `X` with the target `y`. Right now the choices are for a user to pass in a name for a column to select from a DataFrame or to pass in a single column array. We should expand this to allow the user to pass an index to ensure that different data types can be passed to the visualizer. 

### Proposal/Issue

Currently, the plot accepts as hyperparameters `feature` (soon to be `column`) and `target`. Both `feature`/`column` and `target` act as x and y labels in the visualizer respectively (and in fact, `target` has no purpose other than to act as a label). The `feature`/`column` attribute, however, acts as a selection mechanism to describe what to plot against the target from the 2D array, `X` passed into fit. Right now the options are:

1. Pass in a string to identify a column in a DataFrame 
2. Pass in None and a single column `X` 

We should expand this to include:

1. Pass in an int - select the column by index and label with the id 
2. If `target` is specified but not `y`, then use `target` similarly to `feature`/`column` to select from `X` 
3. Test use cases with matrix, python list, structured array, and data frame 

**Caveats**: I think 2. requires some discussion, which we should undertake during the PR. I'm not sure this makes a lot of sense, but would be useful. Potentially changing the names of the visualizer parameters would help make this behavior more clear.

bbengfort:
Note, this task depends on #414

Issue #687, opened 2019-01-14
Add plot directives


rwhitt2049:
Checklist to go through for (potentially) replacing python code + png import with the plot directive. 
This is just a project management issue to give contributors some direction.

 ## Quickstart

http://www.scikit-yb.org/en/latest/quickstart.html
quickstart.rst

- [x] bikeshare_rank2d.png
- [x] temp_feelslike_jointplot.png
- [x] bikeshare_ols_residuals.png
- [x] bikeshare_ridge_alphas.png
- [x] bikeshare_ridge_prediction_error.png

## Gallery 

&hellip; lots to do and figure out here &hellip;

**update**: we've made the gallery independent and created a script to generate the gallery docs (the plot directive didn't work well in the gallery). See #732 

## Tutorial

**update:** this has been updated to use the new datasets module (see #818) but we've opted not to do the plot directive because it disrupts the flow of the tutorial and makes the docs take a long time to build

~Lot's of split code here, so I'll need to know how the team wants it formatted~

~http://www.scikit-yb.org/en/latest/tutorial.html#
tutorial.rst~

~- [ ] modelselect_linear_svc.png~
~- [ ] modelselect_nu_svc.png~
~- [ ] modelselect_svc.png~
~- [ ] modelselect_sgd_classifier.png~
~- [ ] modelselect_kneighbors_classifier.png~
~- [ ] modelselect_logistic_regression_cv.png~
~- [ ] modelselect_logistic_regression.png~
~- [ ] modelselect_bagging_classifier.png~
~- [ ] modelselect_extra_trees_classifier.png~
~- [ ] modelselect_random_forest_classifier.png~

## API 
- [ ] [Anscombe's quartet](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/docs/api/anscombe.rst): anscomb.png

### Radviz
- [x] [Features\radviz](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/docs/api/features/radviz.rst) radviz.png

### Rankd
- [x] [Features\rankd](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/docs/api/features/rankd.rst) rank1d_shapiro.png
- [x] [Features\rankd](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/docs/api/features/rankd.rst) rank2d_pearson.png
[...]

Issue #68, opened 2016-10-07
Improved Rank2D


bbengfort:
The Rank2D visualizer is a feature analysis visualizer that ranks pairwise joint plots of feature columns together (similar to a SPLOM) using a metric in the space [-1, 1] or [0, 1]. The rankings are visualized by a heatmap with only the lower left triangle visible and a diverging or sequential color map scheme that shows the relative ranks of pairs of features. 

By using different ranking metrics (Pearson, Covariance, etc), data scientists can detect issues in dependent variables that might impact machine learning - for example covariance, entropy, non-uniformity etc. 

See #6 for more 

The following ranking metrics should be added:

- [x] Pearson correlation 
- [x] Covariance 
- [ ] Spearman correlation 
- [ ] Kendall Tau correlation 
- [ ] Least Squares Error 
- [ ] Quadracity 
- [ ] Density based outlier detection 
- [ ] Uniformity (entropy of grids) 
- [ ] Number of items in most dense region 

See: Seo, Jinwook, and Ben Shneiderman. "A rank-by-feature framework for interactive exploration of multidimensional data." Information visualization 4.2 (2005): 96-113.

The following visual improvements need to be made: 

- [ ] Make the colobar smaller and nicer 
- [ ] Add xlabels, ylabels and ticks that are nicely spaced 
- [ ] Allow for annotations in the cells 
- [ ] Make the "null" variables greyed out 
- [ ] Show joint plot on click 

See: https://github.com/mwaskom/seaborn/blob/master/seaborn/matrix.py#L94

and 

https://stanford.edu/~mwaskom/software/seaborn/examples/many_pairwise_correlations.html

tabishsada:
I created a pull request (https://github.com/DistrictDataLabs/yellowbrick/pull/429) to add spearman correlation to the list of ranking metrics.

bbengfort:
#645 adds Kendall-Tau

PR #721, opened 2019-02-03
Refresh the JointPlot Visualizer for ML

### LINKED ISSUE #605: JointPlotVisualizer produces different visualizations depending on how the figure and axes objects are defined

bbengfort:
Our original (and wonderful, thank you @pdamodaran!) JointPlotVisualizer was based on seaborn's implementation and used `mpl.GridSpec` for its layout. Since that implementation, we've grown a bit more sophisticated in our axes management, but JointPlot hasn't been updated to reflect this.

This PR refreshes JointPlot with the `make_axes_locatable` method of layout, similar to what we use in the ResidualsPlot. Additionally, this PR rethinks the JointPlot from the ML perspective of passing data as X and y, rather than seaborn's more data-standard approach. For example, the JointPlot can plot two vectors x, y or it can index into a data table to plot against the target, or correlate to features against each other with respect to the target (colors). Said simply, this JointPlot is an ML diagnostic that is intended to be directly used with Rank2D and with the datasets that sklearn users expect.

Fixes #605, #434, #214

The primary change revolves around the selection of the data to visualize when you pass in `X` and maybe `y` to `fit()`. The selection is based on what you pass to a new argument, `columns` that replaces `feature` and `target` as follows:

1. direct data visualization: pass in `X` and `y` as 1D vectors or `X` as 2 columns and `y` as None.
2. feature to target joint plot, columns is an index into `X` to select a column plotted against `y`.
3. feature to feature joint plot, columns contains to indexes into `X`, whose columns are plotted against each other. In this case if the target is specified, the plot can be colored with a heatmap or colors for each class. 

A lot has happened here, but there is still a lot of work to do:

- [x]  rename JointPlot and alias JointPlotVisualizer
- [x]  add pearson correlation as label
- [x] update layout to modify axes rather than using GridSpec
- [x] allow direct data visualization
- [x] feature to target joint plot 
- [x] feature to feature joint plot
- [ ] <strike>implement colors based on the target in feature to feature joint plot</strike>
- [x] finish up the tests
- [ ] <strike>re-add the best fit line</strike> 
- [ ] <strike>add multiple lines of best fit</strike>
- [ ] <strike>update the documentation</strike> 
- [x] implement spearman, covariance, and kendalltau correlation 
- [ ] <strike>quick method</strike>
- [ ] <strike>make aspect ratio equal (square plot)</strike>

This is the current state of the visualizer output for the viz from the quickstart:

jointplot

Note to reviewers: it may be easier to look at the new files rather than at the diffs; a lot has changed!


yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,369 @@
...
+
+
 ##########################################################################
 ## Joint Plot Visualizer
 ##########################################################################
 
 
-class JointPlotVisualizer(FeatureVisualizer):
+class JointPlot(FeatureVisualizer):
     """
```
bbengfort: Still need to finish the docstring.

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,369 @@
...
     kwargs : dict
         Keyword arguments that are passed to the base class and may influence
         the visualization as defined in other Visualizers.
 
+    Attributes
+    ----------
+    corr_ : float
+        The correlation or relationship of the data in the joint plot, specified by the
+        correlation algorithm.
+
```
bbengfort: The correlation coefficient is now stored on the visualizer for easy access.

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,369 @@
...
-        correct data type
-
-        Fit calls draw
+    # TODO: should we couple more closely with Rank2D?
+    correlation_methods = {
+        'pearson': lambda x, y: pearsonr(x,y)[0],
+        'spearman': lambda x, y: 1.0,
+        'covariance': lambda x, y: 1.0,
+        'kendalltau': lambda x, y: 1.0,
+    }
```
bbengfort: Maybe we should more directly couple this with Rank2D?

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,369 @@
...
+        self.joint_kws = joint_kws
+        self.hist_kws = hist_kws
+
+    @property
+    def xhax(self):
+        if self._xhax is None:
+            raise AttributeError(
+                "this visualizer does not have a histogram for the X axis"
+            )
+        return self._xhax
```
bbengfort: xhax is the axes of the histogram for the X-axis that is on the top of the JointPlot. yhax is the histogram for the Y-axis that is to the right of the JointPlot

bbengfort:
Going through the tests, I encountered some issues with making the joint plot square using the aspect ratio. I've raised this question on Stack Overflow:

https://stackoverflow.com/questions/54545758/create-equal-aspect-square-plot-when-data-limits-are-different

In the meantime, I think this PR is getting large, so my plan is to remove the aspect ratio, finish tests, and then add the remaining tasks to a new PR.

yellowbrick/features/jointplot.py
```diff
@@ -1,11 +1,11 @@
 
 # yellowbrick.features.jointplot
-# Implementations of joint plots for univariate and bivariate analysis.
+# Implementation of joint plots for univariate and bivariate analysis.
 #
 # Author:   Prema Damodaran Roman
 # Created:  Mon Apr 10 21:00:54 2017 -0400
 #
-# Copyright (C) 2017 District Data Labs
+# Copyright (C) 2017 The scikit-yb developers.
```
rebeccabilbro: Nice

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,386 @@
...
+
+    columns : int, str, [int, int], [str, str], default: None
+        Determines what data is plotted in the joint plot and acts as a selection index
+        into the data passed to ``fit(X, y)``. This data therefore must be indexable by
+        the column type (e.g. an int for a numpy array or a string for a DataFrame).
+
+        If None is specified then either both X and y must be 1D vectors and they will
+        be plotted against each other or X must be a 2D array with only 2 columns. If a
+        single index is specified then the data is indexed as ``X[columns]`` and plotted
+        jointly with the target variable, y. If two indices are specfied then they are
```
rebeccabilbro: Small typo: specfied => specified

yellowbrick/features/jointplot.py
rebeccabilbro: 👍

yellowbrick/features/jointplot.py
rebeccabilbro: Some text you can use as a start if you'd like:

Joint plots are useful for machine learning on multi-dimensional data, allowing for the visualization of complex interactions between different data dimensions, their varying distributions, and even their relationships to the target variable for prediction.  The Yellowbrick `JointPlot` can be used for both pairwise feature analysis and for feature-to-target plots. For pairwise feature analysis, the `columns` argument can be used to specify the indexes of the two desired columns in `X`. If `y` is also specified, the plot can be colored with a heatmap or by class. For feature-to-target plots, the user can provide either `X` and `y` as 1D vectors, or a `columns` argument with an index to a single feature in `X` to be plotted against `y`. Histograms can be included by setting the `hist` argument to `True` for a frequency distribution, or to `density` for a probability density function.

yellowbrick/features/jointplot.py
rebeccabilbro: I don't think this is strictly necessary, though it would be useful to get someone to do a blog post showing the information conveyed via different correlation methods as depicted using YB visualizers. Potentially it might also be useful to add a note to #68, so that in case someone contributes new correlation measures we remember that we can incorporate them here, too.

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,386 @@
...
+        used to plot the color of points.
+
+        Note that these names are also used as the x and y axes labels if they aren't
+        specified in the joint_kws argument.
+
+    correlation : str, default: 'pearson'
+        The algorithm used to compute the relationship between the variables in the
+        joint plot, one of: 'pearson', 'covariance', 'spearman', 'kendalltau'.
+
+    kind : str in {'scatter', 'hex'}, default: 'scatter'
```
rebeccabilbro: Might be nice to add as a todo to the docs (maybe #687) to add images that show a few different versions of JointPlot (e.g. with hex, with a few different correlation measures, and with the feature-feature-target heatmap option)

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,386 @@
...
+        super(JointPlot, self).__init__(ax=ax, **kwargs)
+        self._xhax, self._yhax = None, None
+
+        # Set and validate the columns
+        self.columns = columns
+        if self.columns is not None and not isinstance(self.columns, (int, str)):
+            self.columns = tuple(self.columns)
+            if len(self.columns) > 2:
+                raise YellowbrickValueError((
+                    "'{}' contains too many indices or is invalid for joint plot"
```
rebeccabilbro: Given my observations with students using different implementations of joint plots, I'm guessing this is going to be one of the more common errors that comes up for users of this visualizer (e.g. providing the wrong combos of column, X, and y args). I'm wondering if we should make this error slightly more verbose (or even break it into two different errors, one for too many indices and another for otherwise invalid input?), so that people have a little bit more to go on.

yellowbrick/features/jointplot.py
rebeccabilbro: Maybe add the above to the docstrings of `def xhax(self)` and `def yhax(self)`? E.g. 

```
    @property
    def xhax(self):
    """
    Return the axes of the histogram for the top of the JointPlot (X-axis)
    """
    ...
```

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,386 @@
...
+        Fits the JointPlot, creating a correlative visualization between the columns
+        specified during initialization and the data and target passed into fit:
 
-            else:
-                self.feature = ['x']
+            - If self.columns is None then X and y must both be specified as 1D arrays
+              or X must be a 2D array with only 2 columns.
+            - If self.columns is a single int or str, that column is selected to be
+              visualized against the target y.
+            - If self.columns is two ints or strs, those coumns are visualized against
```
rebeccabilbro: Small typo:  coumns=> columns

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,386 @@
...
+                    "or X as a matrix with 2 columns"
+                ))
 
-        self.fig = fig
-        self.joint_ax = joint_ax
-        self.x_ax = x_ax
-        self.y_ax = y_ax
-        self._ax = joint_ax
+            if y is None:
+                # Draw the fist column as x and the second column as y
```
rebeccabilbro: Small typo: fist=>first

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,386 @@
...
+            The data to plot for the x axis and the y axis
 
-        self.joint_args.setdefault("alpha", 0.4)
-        facecolor = self.joint_args.pop("facecolor", FACECOLOR)
-        self.joint_ax.set_facecolor(facecolor)
+        xlabel, ylabel : str
+            The labels for the x and y axes.
+        """
+        # This is a little weird to be here, but it is the best place to perform
+        # this computation given how fit calls draw and returns.
```
rebeccabilbro: Agreed

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,386 @@
...
-        if self.joint_plot == "scatter":
-            aspect = self.joint_args.pop("aspect", "auto")
-            self.joint_ax.set_aspect(aspect)
-            self.joint_ax.scatter(X, y, **self.joint_args)
+    def finalize(self, **kwargs):
+        """
+        Finalize executes any remaining image modifications making it ready to show.
+        """
+        # Set the aspect ratio to make the visualization square
+        # TODO: stil unable to make plot square using make_axes_locatable
```
rebeccabilbro: Small typo: stil=>still

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,386 @@
...
-        if self.xy_args is None:
-            self.xy_args = {}
-
-        facecolor_x = self.xy_args.pop("facecolor_x", FACECOLOR)
-        self.x_ax.set_facecolor(facecolor_x)
-        facecolor_y = self.xy_args.pop("facecolor_y", FACECOLOR)
-        self.y_ax.set_facecolor(facecolor_y)
 
+# Alias for JointPlot
+JointPlotVisualizer = JointPlot
```
rebeccabilbro: 👍

yellowbrick/features/jointplot.py
bbengfort: I've updated #728 with this since I think that the documentation is going to be replaced with a bit more detail. I'll also update #687.

yellowbrick/features/jointplot.py
bbengfort: I like the idea of a blog post! I've added the note to #68. It could probably also use a line number for where `JointPlot` does correlations, but I can't get the line number until this is merged.

yellowbrick/features/jointplot.py
bbengfort: I've expanded the error message - but that does raise an issue; I have validation occurring at both `__init__` and in `fit` e.g. in the case where someone does this:

```python
oz = JointPlot(columns="foo")
oz.columns = ["foo", "baz", "bear"] 
oz.fit(X)
```

The validation in `fit()` is actually more specific because it only occurs as it goes through the possible columns. I wonder if I should just remove the validation in `__init__` and have it only happen in `fit()` and `draw()`? I wonder how sklearn handles this.

yellowbrick/features/jointplot.py
bbengfort: I've added the docstrings, but note that this is for developers only because of the `@property` decorator. Using `help(oz.xhax)` will return the docstring for the `matplotlib.Axes` object.

yellowbrick/features/jointplot.py
```diff
@@ -14,308 +14,386 @@
...
+                "this visualizer does not have a histogram for the Y axis"
+            )
+        return self._yhax
 
-        kwargs: dict
-            keyword arguments passed to Scikit-Learn API.
+    def _layout(self):
         """
+        Creates the grid layout for the joint plot, adding new axes for the histograms
+        if necessary and modifying the aspect ratio. Does nothing if hist is falsey.
```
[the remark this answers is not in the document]
bbengfort: I actually did mean falsey here, but perhaps I could have communicated it better. It does nothing not only if `hist is False` but also if `hist is None` or `hist == ""`, etc. I guess I am using some [JavaScript terminology](http://adripofjavascript.com/blog/drips/truthy-and-falsy-values-in-javascript.html), though a quick Google reveals the term is used in Python too. I'll update to `if hist is False or None` since those are really the only two values I expect to be passed in as the arguments.

yellowbrick/features/jointplot.py
bbengfort: But a ✊  column would be cool, sort of a "power to the people" kind of thing.

bbengfort:
Thanks for the review @rebeccabilbro updates applied and the wrap up is in #728 -- will merge when tests pass.

PR #729, opened 2019-02-06
Alter ParallelCoordinates docs to include plot:: directives

lwgray:
_Note: this is towards #687_

Adding in the plot directives isn't as straightforward as it seems.

Mainly, the produced plots aren't identical to those being replaced
 ~~a. They sometimes had extra gridlines~~
 ~~b. They sometimes had added margins~~

~~Secondly, the plot directive option :include-source: was set to false on a couple of plots~~

The plots generated are below
pcoord-1
pcoord-2
pcoord-3


docs/api/features/pcoords.rst
```diff
@@ -5,52 +5,54 @@ Parallel Coordinates
 
 Parallel coordinates is multi-dimensional feature visualization technique  where the vertical axis is duplicated horizontally for each feature. Instances are displayed as a single line segment drawn from each vertical axes to the location representing their value for that feature. This allows many dimensions to be visualized at once; in fact given infinite horizontal space (e.g. a scrolling window), technically an infinite number of dimensions can be displayed!
 
-Data scientists use this method to detect clusters of instances that have similar classes, and to note features that have high variance or different distributions. We can see this in action after first loading our occupancy classification dataset:
+Data scientists use this method to detect clusters of instances that have similar classes, and to note features that have high variance or different distributions. We can see this in action after first loading our occupancy classification dataset. Note that the visualization can be drawn with either the ``ParallelCoordinates`` visualizer or using the ``parallel_coordinates`` quick method:
```
lwgray: Tried my best to make this as fluid as possible.  the "Note that the visualization can be drawn..." was previously used between split inline code.

docs/api/features/pcoords.rst
```diff
@@ -5,52 +5,54 @@ Parallel Coordinates
...
 
-.. code:: python
+.. plot::
+
+    from yellowbrick.features import ParallelCoordinates
+    from yellowbrick.datasets import load_occupancy
 
     # Load the classification data set
-    data = load_data("occupancy")
+    X, y = load_occupancy()
```
lwgray: utilizing new datasets module...

docs/api/features/pcoords.rst
```diff
@@ -5,52 +5,54 @@ Parallel Coordinates
...
-The visualization can be drawn with either the ``ParallelCoordinates`` visualizer or using the ``parallel_coordinates`` quick method:
-
-.. code:: python
-
-    from yellowbrick.features import ParallelCoordinates
-
     # Instantiate the visualizer
     visualizer = ParallelCoordinates(
-        classes=classes, features=features, sample=0.5, shuffle=True
+        classes=classes, features=features, sample=0.05, shuffle=True
```
lwgray: correcting sampling from as pointed out by @Naba7

docs/api/features/pcoords.rst
```diff
@@ -5,52 +5,54 @@ Parallel Coordinates
...
     )
 
     # Fit and transform the data to the visualizer
     visualizer.fit_transform(X, y)
 
     # Finalize the title and axes then display the visualization
-    visualizer.poof()
-
+    visualizer.finalize()
+    visualizer.ax.grid(False)
```
lwgray: Grid lines are getting drawn for some reason so I just run ``finalize`` and turned them off

docs/api/features/pcoords.rst
```diff
@@ -78,11 +80,96 @@ Parallel coordinates can take a long time to draw since each instance is represe
 
 The "fast" drawing mode vastly improves the performance of the parallel coordinates drawing algorithm by drawing each line segment by class rather than each instance individually. However, this improved performance comes at a cost, as the visualization produced is subtly different; compare the visualizations in fast and standard drawing modes below:
 
-.. image:: images/fast_vs_slow_parallel_coordinates.png
+.. plot::
+    :include-source: False
```
lwgray: In the docs, only an image is shown so I utilized this directive option to turn off the source code

docs/api/features/pcoords.rst
bbengfort: Yeh, that's tricky - but thanks for working it in! Another option with the documentation is to use a callout. E.g.

```
.. note:: These visualizations can be produced with either the ``ParallelCoordinates`` visualizer or by using the `parallel_coordinates`` quick method.
```

Which renders similarly to:

screenshot 2019-02-06 14 02 58

I don't know if that's better here, but it is an option.

docs/api/features/pcoords.rst
bbengfort: Part of the reason we're extending the documentation with this code is so that we can get a visual confirmation of exactly what the latest version of the visualizer is doing. It could be that the grid being drawn is a bug and we need to fix it. Either way, it would probably be best to leave the code as we expect our users to run it (e.g. calling `poof()`) and then we can debug if the problem is in the visualizer or if it is something to do with the `.. plot::` directive. 

Thanks for keeping an eye out on the changes to the rendered images though, I think that's a pretty important part of going through these documentation updates.

docs/api/features/pcoords.rst
```diff
@@ -60,9 +62,9 @@ Normalization techniques can be directly applied to the visualizer without pre-t
 
     # Fit the visualizer and display it
     visualizer.fit_transform(X, y)
-    visualizer.poof()
+    visualizer.finalize()
+    visualizer.ax.grid(False)
```
bbengfort: See comment above.

docs/api/features/pcoords.rst
```diff
@@ -5,52 +5,54 @@ Parallel Coordinates
...
+    from yellowbrick.datasets import load_occupancy
+
+    # Load the classification data set
+    X, y = load_occupancy()
+
+    # Specify the features of interest and the classes of the target
+    features = [
+        "temperature", "relative humidity", "light", "C02", "humidity"
+    ]
+    classes = ["unoccupied", "occupied"]
```
bbengfort: This is really just something for your attention, @lwgray and you don't have to update the docs unless you think it's valid, but with the new datasets you can actually get the features and classes without specifying them manually:

```python
data = load_occupancy(return_dataset=True)

features = data.meta['features']
classes = data.meta['labels'].keys()

X, y = data.to_data() # or to_numpy or to_pandas
```

I don't know if this is helpful for the documentation since it seems a little extra to get the raw dataset object rather than directly getting the dataframe, but I thought you might be interested for your own use or tests or other documentation.

docs/api/features/pcoords.rst
bbengfort: So I ran this code on my machine and I ended up with the following (I ran the code directly, I didn't build the documentation):

pcoords

So this may have something to do with how the docs are configured and we can look into it.

docs/api/features/pcoords.rst
bbengfort: Perfect, this nicely demonstrates fast vs standard parallel coordinates and the source doesn't need to be shown to the user since it's just a demonstration!

docs/api/features/pcoords.rst
```diff
@@ -78,11 +80,96 @@ Parallel coordinates can take a long time to draw since each instance is represe
...
+    ax.fill_between(
+        factors, speedups - variance, speedups + variance, alpha=0.25,
+        color='b',
+    )
+
+    ax.set_ylabel("speedup factor")
+    ax.set_xlabel("dataset size (number of repeats in Iris dataset)")
+    ax.set_title("Speed Improvement of Fast Parallel Coordinates")
+    ax.grid(True)
+    ax.margins(0)
```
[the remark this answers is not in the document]
bbengfort: So I struggled with this one - this is actually benchmark code and was part of what @thekylesaurus contributed in #420. Although it could be something we want to rerun when we build the documentation, I would rather continue to report his original results. Thank you so much for your hard work on it, and I'm sorry that I forgot this was here and didn't get a chance to warn you. But perhaps it would be better if we put this code back into a `pcoord_benchmark.py` file and kept `.. image:: images/fast_parallel_coordinates_speedup.png` here?

docs/api/features/pcoords.rst
lwgray: I ran the code also but It is different once you build the docs...

bbengfort: Ok, that's good -- that means we can look into the documentation settings. If you can change the docs back to `poof()` then I can clone your PR and build the docs to see if I can spot what's going on.

docs/api/features/pcoords.rst
lwgray: I will try to incorporate the note directive..let me know what you think

docs/api/features/pcoords.rst
lwgray: For a new user, II think that have the features & classes spelled out would be useful

lwgray: but then again...this is an opportunity to show how the datasets module works

bbengfort: Don't worry about it too much - it was more to let you know; I'd say keep the documentation the way it is.

docs/api/features/pcoords.rst
lwgray: great suggestions...now I have to decide if I should place before the first plot directive or after

lwgray:
I added images to PR description 😄

docs/api/features/pcoords.rst
lwgray: I changed them back to poof and included the images in the PR description

docs/api/features/pcoords.rst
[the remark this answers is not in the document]
lwgray: As requrested, I moved the code to pcoord_benchmark.py then dug through my git commits to find/restore fast_parallel_coordinates_speedup.png

bbengfort:
@lwgray I pushed a change to the PR that I think solved the problem, basically, in every directive, I did this:

```
.. plot::
    :context: close-figs
```

From the [plot directive docs](https://matplotlib.org/devel/plot_directive.html):

> context : bool or str
> If provided, the code will be run in the context of all previous plot directives for which the :context: option was specified. This only applies to inline code plot directives, not those run from files. If the :context: reset option is specified, the context is reset for this and future plots, and previous figures are closed prior to running the code. :context:close-figs keeps the context but closes previous figures before running the code.

I tried `:context: reset` also but that didn't seem to do anything. Adding the close-figs context changed the look and feel of all the visualizations currently implemented using the directive, so we need to pay attention to this as we move forward. 

Could you do a pull, build the docs and see if the fix worked for you?

docs/api/features/pcoords.rst
```diff
@@ -20,7 +20,7 @@ Data scientists use this method to detect clusters of instances that have simila
 
     # Specify the features of interest and the classes of the target
     features = [
-        "temperature", "relative humidity", "light", "C02", "humidity"
+        "temperature", "relative humidity", "light", "CO2", "humidity"
```
lwgray: great catch!

bbengfort:
@lwgray well I tried to reproduce the behavior, and I couldn't -- I'm not sure what to do except to merge the PR and see what happens in the docs. However, it is probably wise to add the context to importances and elbow - would you mind putting those back in?

bbengfort:
Also now that we've merged #732 - perhaps we could also clean up the images that remained from the gallery plot?

lwgray:
I am not sure if I follow regarding cleaning up images.  Because some static images are needed, like in the case of the Pcoords benchmark image, and should not be deleted...I think everyone should clean up the images for their assigned visualizers.  I will go back through the ones I created plot directives for and also importances and elbow

lwgray:
@bbengfort This PR is ready to merge once it finished it's test.

bbengfort:
@lwgray I'll merge - I think I was thinking of the dispersion plot that still remained because it was used in the gallery.

PR #732, opened 2019-02-08
Make gallery independent of plot directives

bbengfort:
This PR is an attempt to solve the problem that the gallery is dependent on static images generated by the previous inline scripts. As we replace these with plot directives in #687 we may run into gallery conflicts because of this dependency, which means we can't clean up the code that the plot directives replaced. This PR, therefore, does the following:

- Creates a `docs/gallery.py` script to generate  gallery-specific images
- Relinks the gallery images to the `images/gallery` directory 
- Cleans up the main `images/` directory to remove the `name_` prefixes to `name/` dirs 
- Adds a favicon (!) 

This solution has pros and cons:

- **Pro**: the thumbnails are generated better (with a specific figsize and params)
- **Pro**: only the `gallery.py` script generates gallery images 
- **Pro**: we can generate specific thumbnails on demand 
- **Pro**: we can include complex plots that take time to build since we won't build them every time the doc is built. 
- **Con**: we have to maintain the gallery script 
- **Con**: the gallery will not be up to date with code changes
- **Con**: running "all" gallery images takes a _long_ time

Seaborn has a custom [sphinxext](https://github.com/mwaskom/seaborn/blob/master/doc/sphinxext/gallery_generator.py) that generates their entire RST for their gallery; this is something to consider for the future, but I thought it would be quicker just to make the gallery images independent for now.

I did consider adding plot directives into the gallery, but that ended up like this:

screenshot 2019-02-07 22 08 15

Which is not ideal because the layout shifts and the source code and image links are at the top. I'd like these links to be present in the rest of the code and there is no way to turn them off for specific plots. 

This will only take me another day to complete, but is obviously a lot of effort so if you have thoughts @DistrictDataLabs/team-oz-maintainers please let me know soon before I go farther down this road!


README.md
```diff
@@ -13,15 +13,15 @@
 
 **Visual analysis and diagnostic tools to facilitate machine learning model selection.**
 
-[![Classifiers](docs/images/readme_classifiers.png)](https://www.scikit-yb.org/en/latest/gallery.html)
-[![Clusterers](docs/images/readme_clusterers.png)](https://www.scikit-yb.org/en/latest/gallery.html)
-[![Regressors](docs/images/readme_regressors.png)](https://www.scikit-yb.org/en/latest/gallery.html)
+[![Classifiers](docs/images/readme/classifiers.png)](https://www.scikit-yb.org/en/latest/gallery.html)
+[![Clusterers](docs/images/readme/clusterers.png)](https://www.scikit-yb.org/en/latest/gallery.html)
+[![Regressors](docs/images/readme/regressors.png)](https://www.scikit-yb.org/en/latest/gallery.html)
```
bbengfort: part of the cleanup of the images directory, I replaced the `_` with a `/` for a subdirectory.

docs/conf.py
```diff
@@ -225,7 +225,7 @@
 # the docs.  This file should be a Windows icon file (.ico) being 16x16 or 32x32
 # pixels large.
 #
-# html_favicon = None
+html_favicon = "images/favicon.ico"
```
bbengfort: We have a favicon!

docs/gallery.py
```diff
@@ -0,0 +1,166 @@
...
+##########################################################################
+## Helper Methods
+##########################################################################
+
+def newfig():
+    """
+    Helper function to create an axes object of the gallery dimensions.
+    """
+    # NOTE: this figsize generates a better thumbnail
+    _, ax = plt.subplots(figsize=(8,4))
```
bbengfort: Setting this figsize makes the thumbnails a lot better looking.

docs/matplotlib.rst
```diff
@@ -173,7 +173,7 @@ For this example, we’ll draw a line showing an average and include labels show
     # Hide the legend
     ax.legend().set_visible(False)
 
-.. image:: images/matplotlib_single.png
+.. image:: images/matplotlib/single.png
```
bbengfort: more images directory cleanup replacing `_` with `/`

docs/conf.py
bbengfort: Note the new favicon on the left most tab and the RTD icon on the right most tab:

screenshot 2019-02-07 22 22 02

docs/gallery.py
bbengfort: This is the new features gallery:

screenshot 2019-02-07 22 21 51


This is the old one:

screenshot 2019-02-07 22 21 43

docs/conf.py
bbengfort: I'm not 100% sure this will show up on RTD but I'm hopeful.

lwgray: Loving the favicon

docs/gallery.py
lwgray: You are right, the new ones do look better

docs/gallery.py
```diff
@@ -0,0 +1,166 @@
...
+
+def jointplot():
+    X, y = load_concrete()
+    oz = JointPlot(columns=["cement", "splast"], ax=newfig())
+    oz.fit_transform(X, y)
+    savefig(oz, "jointplot")
+
+
+if __name__ == "__main__":
+    plots = {
```
lwgray: ``plots`` is going to get bloated when all the visualizers are added.

docs/gallery.rst
```diff
@@ -182,95 +182,69 @@ Clustering Visualizers
...
     :width: 200px
     :height: 100px
-    :alt: Dispersion Plot
+    :alt: Dispersion of Words in a Corpus
     :target: api/text/dispersion.html#dispersion-plot
 
-Decision Boundaries Visualizer
-------------------------------
-
-.. image:: api/contrib/images/knn_decisionviz.png
```
bbengfort: Moved to "Model Selection" section

yellowbrick/target/binning.py
```diff
@@ -118,15 +118,6 @@ def fit(self, y, **kwargs):
         self.draw(y)
         return self
 
-
-    def poof(self, **kwargs):
```
bbengfort: Removed this method because it shouldn't be in visualizer subclasses and because it was making the image not be rendered.

bbengfort:
@lwgray ok, I've finished up all of the gallery figures - it was a slog but it should mean that we can clean up all stray figures while doing the plot directives. Would you mind giving this a review so we can merge it?

yellowbrick/target/binning.py
lwgray: @bbengfort I found that in my audit...thanks for fixing it.

docs/gallery.rst
```diff
@@ -182,95 +182,69 @@ Clustering Visualizers
...
 
-.. image:: api/target/images/class_balance_compare.png
+.. image:: images/gallery/class_balance.png
     :width: 200px
     :height: 100px
     :alt: Class Balance
     :target: api/target/class_balance.html#class-balance
 
-.. image:: api/target/images/feature_correlation_pearson.png
+.. image:: images/gallery/feature_correlation.png
```
lwgray: any particular reason to remove other feature_correlation images

rebeccabilbro:
The [new gallery](https://www.scikit-yb.org/en/develop/gallery.html) looks terrific &mdash; nice work @bbengfort and @lwgray!

docs/gallery.py
bbengfort: That's true, not sure of what the alternative is though.

docs/gallery.rst
bbengfort: They just seemed kind of redundant and didn't add anything new to the gallery; it was just three bar charts next to each other. I wanted the gallery to feel like a high-level overview of all the things YB could do.

bbengfort:
Thanks for the review @lwgray !

docs/gallery.rst
lwgray: noted!

Issue #509, opened 2018-07-19
Fix code headers and run ID line script


bbengfort:
This is not a big deal but it seems like folks enjoy having their names on the individual files they've worked on via the code headers. The code header in the form of:

```python
# yellowbrick.target
# Implements visualizers related to the dependent (target) variable, y.
#
# Author:  Benjamin Bengfort <benjamin@bengfort.com>
# Created: Thu Jul 19 08:57:05 2018 -0400
# License: Apache-2.0
#
# ID: __init__.py [abcd123] benjamin@bengfort.com $
```

Is actually kind of a throwback from some of my earliest coding days when we had git commit hooks that did automatic checking for safety and linting (before the days of GitHub services that did this). The git commit hook would actually fill in all of the information except for the description line. 

I actually did dig up the script that did this and did some repairs to it. I'd like to make all the files in the package consistent with this header by running the script. Unfortunately that will make a lot of changes to a lot of files; and I don't want to do that in a PR - rather I think that we should do this on our next release as part of the version bump. 

CC: @rebeccabilbro @ndanielsen @lwgray

bbengfort:
*Update*: In a recent call we decided to have the copyright line attributed to "The scikit-yb developers", the header should be updated to reflect this.

bbengfort:
Fixed in #957

rebeccabilbro:
Here's the [gist](https://gist.github.com/rebeccabilbro/4bde1df0bce8b0e241a90576e874ec21) for the ID line script

PR #735, opened 2019-02-10
Update text visualizer tests with new datasets module

rebeccabilbro:
This PR is towards #682 and updates our text visualizer tests to use the new `yellowbrick.datasets`. I also went ahead and updated the copyright line in those files per the new convention.


tests/test_text/test_base.py
```diff
@@ -4,7 +4,7 @@
 # Author:   Benjamin Bengfort <bbengfort@districtdatalabs.com>
 # Created:  Mon Feb 20 06:34:50 2017 -0500
 #
-# Copyright (C) 2016 District Data Labs
+# Copyright (C) 2016 The scikit-yb developers
```
bbengfort: I think this looks really good, I'm glad we're going this route! I think it also reinforces our Apache 2 license.

tests/test_text/test_dispersion.py
```diff
@@ -21,23 +21,28 @@
...
 from yellowbrick.text.dispersion import *
-from tests.dataset import DatasetMixin
 from tests.base import VisualTestCase
 import matplotlib.pyplot as plt
 
+##########################################################################
+## Data
+##########################################################################
+
+corpus = load_hobbies()
```
bbengfort: Just out of curiosity, why make this variable a module level var? The reason I ask is that typically when I see this, there is some advantage to be gained by caching expensive work across all tests in the module, but I want to make sure we're on the same page in that the data is read from disk (e.g. not cached) every time `corpus.data` is called. This is not true for `corpus.labels`, which is cached on the first read. 

The `Corpus` object cannot be easily modified because of it's use `@property` decorators, so I'm not worried about testing errors due to `corpus` being modified between different tests, and so placing this here will save some lines of code in the tests below.

tests/test_text/test_dispersion.py
```diff
@@ -47,12 +52,10 @@ def test_quick_method(self):
 
         self.assert_images_similar(ax=ax, tol=25)
 
-    def test_integrated_dispersionplot(self):
+    def test_integrated_dispersion_plot(self):
```
bbengfort: Thanks for resyncing the baseline images with this name change (and the below name changes) and cleaning up the older images!

tests/test_text/test_dispersion.py
```diff
@@ -110,35 +107,29 @@ def test_dispersionplot_annotate_docs(self):
...
-    def test_dispersionplot_color_words_by_class(self):
+    def test_dispersion_plot_color_by_class(self):
         """
         Assert no errors occur during DispersionPlot integration
         when target values are specified
         """
-        corpus = self.load_data('hobbies')
-
-        text = (doc.split() for doc in corpus.data)
+        target = corpus.target
```
bbengfort: exactly where I would have put this :-)

tests/test_text/test_tsne.py
```diff
@@ -4,7 +4,7 @@
 # Author:   Benjamin Bengfort <benjamin@bengfort.com>
 # Created:  Mon Feb 20 07:23:53 2017 -0500
 #
-# Copyright (C) 2016 Bengfort.com
```
bbengfort: Whoops ...

tests/test_text/test_postag.py
```diff
@@ -2,13 +2,13 @@
...
-# Author:   Rebecca Bilbro <rbilbro@districtdatalabs.com>
+# Author:   Rebecca Bilbro
 # Created:  2017-03-22 15:46
 #
-# Copyright (C) 2017 District Data Labs
+# Copyright (C) 2018 The scikit-yb developers
 # For license information, see LICENSE.txt
 #
-# ID: test_postag.py [bd9cbb9] rebecca.bilbro@bytecubed.com $
+# ID: test_postag.py [bd9cbb9] $
```
bbengfort: We should probably discuss this in #509 - the script I was referring to automatically creates this `ID` line and includes the email of the git user that made the first commit of the file. If you don't want those emails here, we'll have to adapt the script.

tests/test_text/test_dispersion.py
rebeccabilbro: I don't have a strong opinion about how this should be done, and I couldn't find an example of how we were planning to use the datasets as specified in #682. The closest I could find were things like [this](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/test_class_prediction_error.py#L39), [this](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_cluster/test_elbow.py#L49), and [this](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_features/test_jointplot.py#L58). Would you like to propose a default method for data fixtures (e.g. just load the data for each test) that can be used across all the updated tests? I can update the issue with that spec and then as I go through an update more of the tests, I'll aim to make everything consistent.

tests/test_text/test_postag.py
rebeccabilbro: Yes, I don't have a clear sense of what specifically #509 is going to do (e.g. what is manual and what is automatic). Perhaps some people would prefer not to share their emails?

tests/test_text/test_dispersion.py
bbengfort: Ok, I just wanted to make sure it wasn't there for some reason I didn't understand - it's perfectly fine being there! 

As for a default method, I think that's going to have to be a little nuanced. The first two examples you gave both have the same caveats like the ones I mentioned above, and I'm not a fan of the middle example, I'd prefer to use `make_classification` as in the first example. The third example is very different, it is defining a function and every time `rand1d()` is called a new dataset is generated (`partial` assigns default arguments to the original function). This is sort of a strange example because in that case the tests are relying on the randomness rather than setting a fixed random state. 

The safest thing to do is to use a [fixture](https://docs.pytest.org/en/latest/fixture.html) as in the [binary classification fixture](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/conftest.py#L30) and [here](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_classifier/conftest.py#L30), but that may simply be adding an extra layer of obscurity that prevents new contributors from understanding our tests. 

For now, this looks good and I think we should merge the PR without changes!

tests/test_text/test_postag.py
bbengfort: The emails are already on the git commit and can be seen by `git log` but you're right that we're making it a lot more prominent by putting it in the header of the files. When we get to #509 I'll make sure I ping you to carry on this discussion.

PR #733, opened 2019-02-09
Implement plot directive for ClassificationReport docs

lwgray:
Addresses #687 
Adds plot directive for classification report visualizer

PR #737, opened 2019-02-10
Implement plot directive for Class_prediction_error (towards #687)

lwgray:
The purpose of this PR is to implement plot directives for the `Class_prediction_Error` in relation to #687 

Implementing Class Prediction Error plot directives led to a couple hurdles/observations.

1.  the script that generates the image for ``class_prediction_error`` docs (/docs/api/classifier/class/prediction_error.py) also produces a second image that isn’t used in the documentation or anywhere else as far as i can tell.  The problem is, Do i preserve the script to keep the the code for the second image?  @bbengfort thought that both images looked good and should probably go into the documentation.  He suggested that I include the plot directive for the second plot but hide it with ``nofigs`` and `include_source` options and leave it to be reviewed for inclusion in this PR.  Both images are shown below.
1st image(originally in documentation)
plot_directive_class_prediction_error

2nd image(added into documentation)
cpe_credit

2.  This is a tag along to observations made by @rebeccabilbro in [PR 733](https://github.com/DistrictDataLabs/yellowbrick/pull/733#discussion_r255340978) in which she noticed differences between old script derived images and plot directive created images.  For ``class_prediction_error`` I noticed that the image generated by the new plot directive is smaller than the old image (see below). Although this might not be an issue at all but I wanted to note it here for future reference.

NEW
plot_directive_class_prediction_error

OLD
old_class_prediction_error


docs/api/classifier/class_prediction_error.rst
```diff
@@ -19,13 +23,39 @@ The class prediction error chart provides a way to quickly understand how good y
...
 
-    from sklearn.ensemble import RandomForestClassifier
+    # Evaluate the model on the test data
+    visualizer.score(X_test, y_test)
 
+    # Draw visualization
+    visualizer.poof()
+
+
+.. plot::
```
lwgray: Should this 2nd plot-directive be included?

rebeccabilbro:
Thanks @lwgray -- I'll review this tomorrow after work if that's ok!

lwgray:
more than ok :smile_cat:

docs/api/classifier/class_prediction_error.rst
rebeccabilbro: @lwgray I like the idea of including (and showing) both plots! One thought; would you be open to adding a few lines of interpretation to each?

docs/api/classifier/class_prediction_error.rst
```diff
@@ -19,13 +23,39 @@ The class prediction error chart provides a way to quickly understand how good y
...
+    :context: close-figs
+    :include-source: False
+    :nofigs:
```
rebeccabilbro: I'm inclined to just go ahead and show the additional fig here

docs/api/classifier/class_prediction_error.rst
```diff
@@ -19,13 +23,39 @@ The class prediction error chart provides a way to quickly understand how good y
...
+
+    from sklearn.model_selection import train_test_split
+    from sklearn.ensemble import RandomForestClassifier
     from yellowbrick.classifier import ClassPredictionError
+    from yellowbrick.datasets import load_credit
 
+    X, y = load_credit()
+    
+    classes = ['default', 'current']
```
rebeccabilbro: Just for interpretability, do you mind changing this to something like: 
```classes = ['account in default', 'current with bills']```
(assuming the text will fit :D )

File: docs/api/classifier/class_prediction_error.rst [hunk 4]
```diff
@@ -19,13 +23,39 @@ The class prediction error chart provides a way to quickly understand how good y
...
+    # Fit the training data to the visualizer
+    visualizer.fit(X_train, y_train)
 
-    from sklearn.ensemble import RandomForestClassifier
+    # Evaluate the model on the test data
+    visualizer.score(X_test, y_test)
 
+    # Draw visualization
+    visualizer.poof()
+
```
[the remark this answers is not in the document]
lwgray: The plot generated each time the docs are built is random and this makes it impossible to write an interpretation since it changes with the plot.  I tried to set the `random_state` for sklearn `make_classification` but that didn't work.  Any suggestions?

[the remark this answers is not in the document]
rebeccabilbro: Ah, too bad that setting the `random_state` didn't work! How about if we swap this dataset with our game dataset? I think it's important to add some interpretation so that folks will know how to use this cool visualizer!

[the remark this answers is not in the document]
lwgray: I got it to work...I just needed to set the `random_state` of the `RandomForestClassifier`

rebeccabilbro:
@lwgray - I think @bbengfort and I sorted out the Appveyer problems. I'm going to update your branch and see how the tests go!

rebeccabilbro:
Thanks for making those updates @lwgray! Now that tests are passing, I'll get this merged in.

PR #740, opened 2019-02-11
Implement plot directive for ROCAUC

lwgray:
Towards #687
Implements plot directive for ROCAUC

Things to note:
---------------------

ROCAUC uses the load_game dataset.  There are a couple of things that should be noted when using this dataset.
@bbengfort  says:
1.  You’ll have to encode `X` it using `OrdinalEncoder` or `OneHotEncoder` :
2.  You'll have to encode `y` using `LabelEncoder `  because `y` needs to be numeric

```python
from sklearn.preprocessing import OrdinalEncoder
from yellowbrick.datasets import load_game
from sklearn.preprocessing import OrdinalEncoder, LabelEncoder

X, y = load_game()
X = OrdinalEncoder().fit_transform(X)
y = LabelEncoder().fit_transform(y)
```


docs/api/classifier/rocauc.rst
```diff
@@ -9,38 +9,29 @@ The Receiver Operating Characteristic (ROC) is a measure of a classifier's predi
...
     visualizer.fit(X_train, y_train)  # Fit the training data to the visualizer
     visualizer.score(X_test, y_test)  # Evaluate the model on the test data
-    g = visualizer.poof()             # Draw/show/poof the data
-
-
-.. image:: images/rocauc_binary.png
+    visualizer.poof()             # Draw/show/poof the data
 
 
 .. warning::
```
rebeccabilbro: It's just occurring to me that now that v0.9 is out we might want to update this to say something like:


Versions of Yellowbrick =< v0.8 had a `bug <https://github.com/DistrictDataLabs/yellowbrick/blob/develop/examples/rebeccabilbro/rocauc_bug_research.ipynb>`_ that triggered an ``IndexError`` when attempting binary classification using a Scikit-learn-style estimator with only a ``decision_function``.  This has been fixed as of v0.9, where the ``micro``,
    ``macro``, and ``per-class`` parameters of ``ROCAUC`` are set to ``False`` for such classifiers.

lwgray:
@rebeccabilbro thanks for the merge

Issue #706, opened 2019-01-30
Remove support for Python 2


bbengfort:
Add Python 3.7 to the Travis and AppVeyor test matrix. Deprecate Python 2.7 support in v1.0.

And while we're at it - let's add CI posts to Slack.

PR #746, opened 2019-02-13
[MRG] Use yellowbrick.datasets in tests for yellowbrick.cluster and yellowbrick.features

zjpoh:
Addresses #682.

Also made changes on `DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working`.


tests/test_cluster/test_elbow.py
```diff
@@ -69,6 +70,10 @@
 
 y = np.array([0, 0, 1, 1, 0, 0, 1, 1, 0, 1, 0, 0, 1, 1, 1, 1, 0, 0])
```
bbengfort: While we're here ... would you mind changing this to a pytest.fixture and use `make_blobs` to create the dataset? Similar to what's happening in test_icdm.py?

tests/test_cluster/test_icdm.py
```diff
@@ -43,6 +43,8 @@
 ## Fixtures
 ##########################################################################
 
+Dataset = namedtuple('Dataset', 'X,y')
```
bbengfort: I very much appreciate that you're trying to remove the `tests/dataset.py` file -- but this does lead to some code duplication; perhaps it would be better to put the `Dataset` and `Split` namedtuples into a `tests/fixtures.py` or `test/utils.py` for import?

tests/test_features/test_pcoords.py
```diff
@@ -163,11 +170,12 @@ def test_pandas_integration_sampled(self):
         """
         Test on a real dataset with pandas DataFrame and Series sampled for speed
         """
-        df = self.load_pandas("occupancy")
+        data = load_occupancy(return_dataset=True)
+        df = data.to_dataframe()
```
bbengfort: Awesome! This is exactly what I hoped would happen with the pandas based tests! This ensures that it's a data frame. However, this brings up another issue, do we need to also ensure that we test numpy arrays, e.g. `test_numpy_integration_sampled` with `data.to_numpy()`? Using `load_occupancy()` will return a DataFrame if pandas is installed, otherwise, it will return numpy arrays. This could cause problems with the image similarity tests. What do you think we should do here?

yellowbrick/cluster/elbow.py
```diff
@@ -18,7 +18,7 @@
 ## Imports
 ##########################################################################
 
-import collections
+from collections.abc import Iterable
```
bbengfort: Thank you so much for fixing this deprecation warning!

zjpoh:
I think splitting up to multiple PRs is a great idea. Let me change the name of this PR to reflect that.

tests/test_cluster/test_elbow.py
zjpoh: Sure. I'll do that.

tests/test_cluster/test_icdm.py
zjpoh: Good idea. I will put it in `tests/fixtures.py`. In the future, we can add other duplicated fixtures there~

tests/test_features/test_pcoords.py
zjpoh: If the main problem is with the text in image similarity tests, then we can test the texts separately. But I'm actually not sure how to use the same test function for both pandas DataFrame and numpy array because the features are selected based on column name.

What if we have another test function that runs only when `pd is None`?

yellowbrick/cluster/elbow.py
zjpoh: I might have to revert this because `collections.abc` is not available in python 2.7.... Any thoughts?

bbengfort: We're planning on deprecating Python 2.7 support in version 1.0 (#706) -- so I guess now is as good a time as any to do that! I'll get a PR open for that so that this PR doesn't break CI.

tests/test_features/test_pcoords.py
bbengfort: It's not about text so much as my wanting to ensure that our visualizers work both for numpy and pandas. Previously everything was numpy arrays and we did something special for pandas. But with the new datasets module, if pandas is installed the tests will be all pandas and no numpy. 

It's a good idea to have a test that only runs when `pd is not None` to specifically test that case for numpy!

tests/test_cluster/test_icdm.py
bbengfort: Ok, perfect thank you! Just one caution, anything that is in `conftest.py` is special and is available to all tests in that module without importing, see https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions for more detail.

bbengfort:
Currently blocked by #757

rebeccabilbro:
@zjpoh @bbengfort &mdash; #757 has been merged so you're officially unblocked!

zjpoh:
Changes in the previous 2 commits.
- Removed all `tests.dataset` from `test_cluster` and `test_features`. 
- Updated the baseline image for `test_features/test_radviz/test_integrated_radiz_with_pandas.png` because the text is `C02` (C-zero-two) not `CO2` (C-Oooo-two).
- **Question**: I made an example changes to address the numpy / pandas difference by simply using `.to_data()` and remove `pytest.mark.skipif` since it will return `.to_numpy()` when `pd is None` and `.to_pandas()` otherwise. Let me know what you think about this modification before I made similar changes to other test functions.
https://github.com/DistrictDataLabs/yellowbrick/blob/b4edd4ca7c60d1356e20083d2e4b52b2a0284950/tests/test_features/test_rfecv.py#L138

zjpoh:
@bbengfort Ready for review. Thanks!

lwgray:
@zjpoh Thank you for working so hard on this PR.  @bbengfort is on sabbatical and will not be around for the next couple of weeks.  I know he looks forward to reviewing this when he returns.

bbengfort:
@zjpoh it looks like there were going to be some merge conflicts with the develop branch - so I merged in the latest develop branch manually; please make sure you do a pull before you make updates!

PR #748, opened 2019-02-14
Adds plot directive to validation curves documentation

bbengfort:
This PR has two primary changes:

1. Cleanup the feature importances documentation according to lessons learned about the plot directive. 
2. Implement plot directive for one of the visualizations in the validation curves documentation. 

Unfortunately, two of the code snippets in the validation curve documentation can take hours to run even with `joblib` parallelism. There was already a warning in the docs about this, but the goal was to get a useful diagnostic image to demonstrate the utility of the visualizer than to create an example that users could easily run. I added some comments about this to the text and fixed up the examples to use the datasets module but kept the script to generate the images and the images themselves. 

We should probably discuss what we want to do here - we could simplify the examples or we could look for a way to speed up the visualizer. Will `GridSearch` strategies get us anywhere here? 

Note: current docs build time is 35.8 seconds


rebeccabilbro:
Thank you very much for the PR @bbengfort! I am running a little slower than usual with the PR reviews given the recent spike in interest in YB 🎉😅; please 🐻 with me &mdash; I'll be taking a look at this soon!

ndanielsen:
@bbengfort just took a look at this, and my two cents is that I think the mission is accomplished here with documentation with high fidelity example especially with the notes of warning about the length of time to build.  

My pragmatic :shipit: instinct is saying that this is great for v1.0, and we should create a technical debt item to optimize the image build process. For the optimization, my gut sense would be to utilize a lower dimensional dataset to create an illustrative, yet still useful example.

bbengfort:
@ndanielsen awesome, thanks for the review! I'll make the technical debt issue and merge this in!

PR #712, opened 2019-01-31
Fix errors related to release of pytest 4.2

bbengfort:
The release of pytest 4.2 broke some of our custom pytest-spec related code (e.g. how the test report is printed). This PR attempts to solve some of the issues, but I've also [opened an issue](https://github.com/pchomik/pytest-spec/issues/21) with pytest-spec to see if we can resolve some of the warnings and add the parametrize ids back into the code base.

- [x] Fix the errors and get tests running again 
- [x] Check if CI now passes 
- [ ] Repair the parametrize custom ids 
- [ ] Fix the (**1146**!!!) warnings now issued by pytest

PR #751, opened 2019-02-15
Attempt to reintegrate spec and verbose as pytest addopts

### LINKED ISSUE #714: Add spec and verbose back to test configuration
See #712 for a more detailed discussion and the changes that relate to this issue. 

The update to pytest 4.2, unfortunately, broke our tests in two ways:

1. The test object no longer has a `_genid` attribute, hurting our verbose names with parametrize
2. pytest-spec issues 1142 warnings making the test output unreadable. 

Temporarily, we have removed pytest-spec and have lost our nice test report output, but I would like to return to verbose mode and pytest-spec when we fix the above bugs.

Note that the user warnings have been reported here: https://github.com/pchomik/pytest-spec/issues/21

rebeccabilbro:
This PR is an attempt to reintegrate spec and verbose as `pytest` `addopts`, by bumping the `pytest` version and reincorporating `pytest-spec`. We had to remove this in #712 due to an update to `pytest 4.2` that broke our tests, which means we no longer have quite as nice a test reporting output format. 

I noticed that `pytest 4.2.1` [was just released this week](https://twitter.com/nicoddemus/status/1095634334437048320) and thought it might be worth a shot to see if the new version resolved the bug that removed the `_genid` attribute from the test obj. If this works, it will fix #714, otherwise we can just close this PR, keep #714 open, and wait for the next `pytest` version.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,8 +13,7 @@ tests = True
 test=pytest
 
 [tool:pytest]
-# TODO: add --spec and --verbose back to addopts
-addopts = --cov=yellowbrick --flakes
+addopts = --verbose --cov=yellowbrick --flakes --spec
 python_files = tests/*
 flakes-ignore =
     __init__.py UnusedImport
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -6,10 +6,10 @@ numpy>=1.13.0
 cycler>=0.10.0
 
 # Testing Requirements
-pytest>=4.2.0
+pytest>=4.2.1
 pytest-cov>=2.6.1
 pytest-flakes>=4.0.0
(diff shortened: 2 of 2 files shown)


bbengfort:
Well this is interesting:

```
========================== deprecated python version ==========================
You are using Python 2.7.15, which will no longer be supported in pytest 5.0
For more information, please read:
  https://docs.pytest.org/en/latest/py27-py34-deprecation.html
```

bbengfort:
Unfortunately, it looks like pytest-spec is still issuing 1108 warnings: thanks for trying though -- it was definitely worth a shot!

rebeccabilbro:
Oh well!

PR #755, opened 2019-02-15
 Changed labels to y vector,passed labels as list 

### LINKED ISSUE #684: Update TSNE docs to clarify what is passed in on instantiation or fit
**Describe the bug**
Currently, the TSNE docs [show an example](http://www.scikit-yb.org/en/latest/api/text/tsne.html#t-sne-corpus-visualization) where a variable named "labels" is passed into the `fit` method; however this variable actually stores `y` (the target), not the labels (which the `fit` method can derive from `y`). This has [lead to some confusion](https://github.com/DistrictDataLabs/yellowbrick/pull/659) about what values should and can be passed in when and where.

**Proposed Fix**
Update the TSNE docs to something like:

```python
from yellowbrick.text import TSNEVisualizer
from sklearn.feature_extraction.text import TfidfVectorizer

# Load the data and create document vectors
corpus = load_corpus('hobbies')
tfidf  = TfidfVectorizer()

X = tfidf.fit_transform(corpus.data)
y = corpus.target

tsne = TSNEVisualizer()
tsne.fit(X, y)
tsne.poof()
```

We should also add another example to illustrate how a user can pass in the labels as a list on instantiation, e.g.:

```python
...

labels = corpus.categories

tsne = TSNEVisualizer(labels=labels)
tsne.fit(X, y)
tsne.poof()
```

Sincere thanks to @jeromemassot for alerting us to this point of confusion!

Naba7:
In this pr,I have changed the docs and labels to X and y vector in the fit method which created confusion in tsne documentation.
Also,I have passed labels as a list to the TSNEVisualizer(labels=labels). I am attaching the image poofed after passing labels as a list.
fixes #684.
tsne_projection_with_labels


docs/api/text/tsne.rst
```diff
@@ -20,20 +20,30 @@ After importing the required tools, we can :doc:`load the corpus <corpus>` and v
...
 
+Now,let's pass on a list of labels to TSNEVisualizer.
+
+.. code:: python
+
+    #Pass on a list of labels
+    labels = corpus.categories
+    tsne = TSNEVisualizer(labels=labels)
+    tsne.fit(X, y)
+    tsne.poof()
```
bbengfort: This certainly implements the change requested by the issue by creating another code snippet that uses the labels argument. However, because this code would draw the same exact figure as the code above, might I suggest we move forward with one of two possible revisions? 

The first option is to simply refactor this into a note, something like:

```
Note that you can pass the class labels or document categories directly to the ``TSNEVisualizer`` as follows:

.. code:: python

    # Pass a list of class labels for the legend
    labels = corpus.categories 
    tsne = TSNEVisualizer(labels=labels)
```

This is a little more compact, and also note the use of the double backticks (` ``TSNEVisualizer`` `) which indicate this is a piece of code. 

The second option is to update the clustering code snippet below on LIne 69 as follows:

```
tsne = TSNEVisualizer(labels=["c{}".format(c) for c in range(5)])
tsne.fit(X, clusters.labels_)
```

You can then add some text to the line above or below that reflect the use of the `labels` argument as in the previous text. 

What do you think of these ideas, do you think one of them is more clear than the other?

docs/api/text/tsne.rst
```diff
@@ -20,20 +20,30 @@ After importing the required tools, we can :doc:`load the corpus <corpus>` and v
     corpus = load_corpus('hobbies')
     tfidf  = TfidfVectorizer()
 
-    docs   = tfidf.fit_transform(corpus.data)
-    labels = corpus.target
+    X   = tfidf.fit_transform(corpus.data)
+    y   = corpus.target
```
bbengfort: Thank you for making this change! Please note that on Line 53 below and Line 66 below, `docs` needs to be updated to `X` because of this change (in the clustering section).

docs/api/text/tsne.rst
Naba7: I will change the docs as instructed.Thank you for your patience and guidance.

docs/api/text/tsne.rst
Naba7: Sure,I am going to change

docs/api/text/tsne.rst
Naba7: I think both are okay.We should keep both the formats so as the other contributors can learn to write a piece of code in those formats.Both are basic python statements,and it will be fun to know both when playing around with code.

docs/api/text/tsne.rst
```diff
@@ -20,27 +20,37 @@ After importing the required tools, we can :doc:`load the corpus <corpus>` and v
...
+    tsne.poof()
+
 If we omit the target during fit, we can visualize the whole dataset to see if any meaningful patterns are observed.
 
 .. code:: python
 
     # Don't color points with their classes
-    tsne = TSNEVisualizer(labels=["documents"])
-    tsne.fit(docs)
+    tsne = TSNEVisualizer(y=["documents"])
```
bbengfort: Unfortunately, this is a small bug, it should be `labels=["documents"]`

rebeccabilbro:
Hi @Naba7 &mdash; just checking, have you been able to make any progress on the follow-up feedback that @bbengfort provided you? We'd like to merge this in but there are still a few things left for you to fix first!

Naba7:
I was stuck with my exams and GSoC-19.Sorry for replying late.I am working on the issues.

docs/api/text/tsne.rst
Naba7: Yes,Thank you .I have made a PR.

pdamodaran:
Hi @Naba7 - thank you so much for updating the PR with the suggested changes.  We are very busy right now and hope to get back to you in a few weeks.

Naba7:
sure thing @pdamodaran .Thank you

Issue #741, opened 2019-02-12
[elbow visualizer] Error in distortion score calculation


maximilian1064:
Instead of the sum of the squared distance of each sample to its closest centroid, what you are calculating here is the sum of mean distortion within each cluster. Do you intend to calculate in this way?

maximilian1064:
https://github.com/DistrictDataLabs/yellowbrick/blob/80169ab4ac8b066abd4c5f4f8f3af34fc2c76d20/yellowbrick/cluster/elbow.py#L102

This line should be 

> distortion += distances.sum()

if you intend to follow the way described in https://bl.ocks.org/rpgove/0060ff3b656618e9136b

bbengfort:
@maximilian1064 Thank you for reporting this, unfortunately right now we're a bit tied up with a number of PRs and open issues so we can't get to this right away. Please ping us in a week or two if we haven't followed up!

zjpoh:
@maximilian1064 I agree with you.

According to the documentation, the distortion should be sum of squared distances, not mean. https://github.com/DistrictDataLabs/yellowbrick/blob/80169ab4ac8b066abd4c5f4f8f3af34fc2c76d20/yellowbrick/cluster/elbow.py#L49

As a simple example, consider the following four points in the same cluster.
```
X = np.array([[0, 1], [1, 0], [-1, 0], [0, -1]])
y = np.array([0, 0, 0, 0])
```
The centroid is the origin and the distortion score should be `4`. However, `distortion_score(X, y)` returns `1.0` instead.

Changing `.mean()` to `.sum()` as proposed by @maximilian1064 solves the problem.

### Code for easy reproducibility
```
import numpy as np
from yellowbrick.cluster.elbow import distortion_score
X = np.array([[0, 1], [1, 0], [-1, 0], [0, -1]])
y = np.array([0, 0, 0, 0])
distortion_score(X, y)
```

pdamodaran:
@maximilian1064 - thanks for bringing up this issue and @zjpoh - thanks for reviewing and providing your feedback.  We will be updating the code with the suggested fix.

maximilian1064:
@pdamodaran Thanks for the fix! I think you should also change this comment
https://github.com/DistrictDataLabs/yellowbrick/blob/80169ab4ac8b066abd4c5f4f8f3af34fc2c76d20/yellowbrick/cluster/elbow.py#L101
to `# Add the sum of square distances to the distortion`.

[...]

PR #762, opened 2019-02-19
Addresses Issue #741 - [elbow visualizer] Error in distortion score calculation

pdamodaran:
As reported by @maximilian1064 and confirmed by @zjpoh, the current version of the elbow visualizer was incorrectly calculating the distortion score.  

The line distortion += distances.mean() needed to be replaced with distortion += distances.sum()

This pull request makes the fix and it also updates the expected results of the tests in test_elbow.py based on the fix made to elbow.py.


yellowbrick/cluster/elbow.py
```diff
@@ -99,7 +99,7 @@ def distortion_score(X, labels, metric='euclidean'):
         distances = distances ** 2
 
         # Add the mean square distance to the distortion
```
bbengfort: Thank you for making this change - would you mind also updating the comment above to reflect the new sum? I know it's small - but I just want to make sure we're consistent!

tests/test_cluster/test_elbow.py
```diff
@@ -221,7 +221,7 @@ def test_distortion_metric(self):
         )
         visualizer.fit(X)
 
-        expected = np.array([ 7.677785,  8.364319,  8.893634,  8.013021])
+        expected = np.array([ 69.100065, 54.081571, 43.146921, 34.978487])
```
bbengfort: Thank you for updating all these values!

yellowbrick/features/pca.py
```diff
@@ -24,6 +24,7 @@
 from yellowbrick.features.base import MultiFeatureVisualizer
 from yellowbrick.style import palettes
 from yellowbrick.exceptions import YellowbrickValueError
+from yellowbrick.draw import manual_legend
```
bbengfort: This file looks like work from a different PR/feature? Would you mind reverting it so we don't include these changes in this PR?

pdamodaran: @bbengfort - I copied the latest code but it looks like there is a hidden character in the last line.  I am not sure how to fix it.  Do you have any ideas?

bbengfort: Add a newline to the end of the file? I think that should fix it.

Issue #600, opened 2018-09-05
Verify and document quick methods


bbengfort:
Our codebase is primarily class-based, however we do have quick-methods for quickly drawing a visualizer from data without a lot of setup (for the common case). Unfortunately, most of these quick methods are undocumented and are not verified. Before the 1.0 release, it would be good if we could include them in our documentation. 

Verifications:

- Does the quick method return a visualizer? (Most probably return an axes object) 
- Does the quick method have an associated test? 
- Does the quick method support all its parent visualizer arguments? 
- Does the quick method behave as expected?

For documentation:

- Should we create a new page for the quick methods? 
- Should we add the docstring automodule to the relevant documentation pages? 

Other questions:

- Should quick methods call `poof()` or `finalize()` should this be an argument (e.g. `poof=True`)? 
- Should quick methods do train/test splits? (See ROCAUC and PrecisionRecallCurve) 
- Do the quick methods require other arguments? 
- Is there any reason not to make the quick methods prime time?

bbengfort:
Marked as high priority for the version 1.0 release (not v0.9)

Kautumn06:
@bbengfort I've finished documenting all of our quick methods in the Yellowbrick audit spreadsheet (#669). The information answers the first three questions under "Verifications" and the second question under "Other Questions" listed above.

bbengfort:
@Kautumn06 awesome! Thank you so much! I'll take a look and then add some comments on this issue about what the next steps are to make this prime time. @ndanielsen would you mind also taking a look?

Kautumn06:
@bbengfort Also, I did add two additional columns to the "Quick Methods" tab to record some additional information I noticed while finishing it. One column to track when a quick method has additional parameters not found in the Visualizer and another that contains notes about when a parameter has different defaults (for instance, `balanced_binning_reference` has `target='Frequency'` while `BalancedBinningReference` has `target=None`). 

The third column I added includes information about when I saw that the difference in parameter names was because of capitalization. For example, `freqdist` has the parameter `N=50`, while it is `n=50` in `FrequencyVisualizer` . Most of the additional examples where this occurred involved one having `x` while the other has `X`.

So I hope that this is helpful! Please just let me know if you have any questions.

bbengfort:
@Kautumn06 very helpful! Now that we have the audit I think we need to do the following:

- [ ] repair all return values 
- [ ] ensure all quick methods have tests 
[...]

PR #768, opened 2019-02-20
Reboot of the PosTag visualizer

rebeccabilbro:
The `PosTagVisualizer` was introduced to Yellowbrick very early on, in [v0.4](https://github.com/DistrictDataLabs/yellowbrick/releases/tag/v0.4), as a prototype `TextVisualizer` designed particularly to assist with diagnostics for NLP with grammar-based feature extraction tools. The original `PosTagVisualizer` was never documented, and never fully fleshed out. In this PR, I've prototyped a reboot of this visualizer that I believe will be more practical for NLP-based ML.

Sample image:

image

Still to do:
 - [x] better docstrings for Sphinx autodoc
 - [x] create a page for this visualizer in the docs
 - [x] add to docs index
 - [x] add to docs gallery
 - [x] add to changelog
 - [x] add spacy to `test/requirements.txt`
 - [ ] feel like i'm forgetting something?

Questions:
 1. is it ok for the Penn Treebank-tagged plots to have different labels than the Universal-tagged plots?
 2. should more of the tags be collapsed (e.g. determiners combined with predeterminers) to make the plot more readable?
 3. what additional features should be included on the roadmap for this visualizer (e.g. to be added to the backlog for follow-on work)?


yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
+
+        Returns
+        -------
+        self : instance
+            Returns the instance of the transformer/visualizer
         """
-        super(PosTagVisualizer, self).__init__(ax=ax, **kwargs)
+        # TODO: add support for other tagsets?
+        if self.tagset == "penn_treebank":
+            self.tag_map = self._penn_tag_map()
```
bbengfort: `self.tag_map` is a learned parameter of this estimator (e.g. it only exists after `fit()` and a user might want access to this data). Therefore might I recommend that this property become `self.tag_map_` with the underscore suffix (e.g. following the sklearn convention) or potentially a bit more verbosely, `self.pos_tag_counts_`. This should also be identified in the `Attributes` section of the class docstring such as [this example](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/features/importances.py#L86).

Minor notes from the review process:

I had originally suggested that `self._penn_tag_map` and `self._uni_tag_map` could be candidates for `@property` decorators, but I see now that you're using these functions to avoid updating mutable data structures on successive calls to `fit` -- nice!

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
+                    elif tag == "NUM":
+                        self.tag_map["number"] += 1
+                    elif tag == "INTJ":
+                        self.tag_map["interjection"] += 1
+                    elif tag == "SYM":
+                        self.tag_map["symbol"] += 1
+                    elif tag == "SPACE":
+                        continue
+                    else:
+                        self.tag_map["other"] += 1
```
bbengfort: These handle methods are extensive and well done. I do very much like how you're distilling the much larger tagsets into a manageable number of larger, human-understandable word classes for use in the visualizer (and how you described that distillation in the documentation). 

An alternative to the case statements is to use a jump table as follows:

```python
jump = {
    # combine proper and regular nouns 
    "NOUN": "noun", "PROPN": "noun",
    "ADJ": "adjective",
    "VERB": "verb",
    # include particles with adverbs 
    "ADV": "adverb", "PART": "adverb",
    "ADP": "adposition",
    "PRON": "pronoun",
    "CCONJ": "conjunction",
    "PUNCT": "punctuation",
    "DET": "determiner",
    "NUM": "number",
    "INTJ": "interjection",
    "SYM": "symbol",
}

for tagged_doc in X:
    for tagged_sent in tagged_doc:
        for _, tag in tagged_sent:
            if tag == "SPACE":
                continue
            self.tag_map_[jump.get(tag, "other")] += 1 
```

This change means that in the worst case (everything is "other") will have the same performance as the best case (everything is a "noun"). This change may be noticible for much larger corpora (e.g. gigabytes of data). If you like this alternate formulation (e.g. maybe more readable/maintainable), then that's a good reason to change; but otherwise we can always mark this is technical debt in the case that this visualizer is used on much larger datasets.

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
+                        self.tag_map["non-English"] += 1
+                    elif tag == "IN":
+                        self.tag_map["preposition"] += 1
+                    elif tag == "POS":
+                        self.tag_map["possessive"] += 1
+                    elif tag == "LS":
+                        self.tag_map["list"] += 1
+                    elif tag == "MD":
+                        self.tag_map["modal"] += 1
+                    elif tag in self.punct_list:
```
bbengfort: See note above about punctuation list, which should be easy enough to adapt to the alternate formulation if required. If more extensive punctuation checking is required, e.g. for unicode, let me know and I can supply a `unicodedata.category` method of punctuation checking.

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
+        Returns
+        -------
+        ax : matplotlib axes
+            Axes on which the PosTagVisualizer was drawn.
+        """
+        colors = resolve_colors(
+            n_colors=len(self.tag_map), 
+            colormap=self.colormap, 
+            colors=self.colors
+        )
```
bbengfort: nice! I certainly like the idea of using a colormap for this visualizer!

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
+        kwargs: dict
+            generic keyword arguments.
         """
-        self.tagged = [
-            (self.TAGS.get(tag),tok) for tok, tag in tagged_tuples
-        ]
-        #
-        # print(' '.join((colorize(token, color) for color, token in self.tagged)))
-        # print('\n')
+        # NOTE: not deduping here, so this is total, not unique
```
bbengfort: I think this makes sense with the title.

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
+        self.ax.set_xticks(range(len(self.tag_map)))
+        self.ax.set_xticklabels(list(self.tag_map.keys()), rotation=90)
+        
+        # Set the axis labels
+        self.ax.set_xlabel(
+            '{} part-of-speech tags'.format(
+                self.tagset_names[self.tagset]
+            )
+        )
+        self.ax.set_ylabel('Count')
```
bbengfort: In the gallery, I commented that `plt.tight_layout` might assist with the tick labels being cut off. From #613 it may be reasonable to add that call here to finalize. Happy to discuss that if you'd like.

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
+
+##########################################################################
+## Quick Method
+##########################################################################
+
+def postag(X, y=None, ax=None, tagset="penn_treebank", colormap=None,
```
bbengfort: This is a great quick method, and I can see it being the primary entry point for this use case. I would recommend that `y=None` might be removed from the arguments unless you think in the future that you might do pos tag counts by category?

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
+            colors=None, **kwargs):
+    """
+    Display a barchart with the counts of different parts of speech
+    in X, which consists of a part-of-speech-tagged corpus, which the
+    visualizer expects to be a list of lists of lists of (token, tag) 
+    tuples.
+
+    Parameters
+    ----------
+    ax : matplotlib axes
```
bbengfort: Don't forget to document `X`!

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
+    # Instantiate the visualizer
+    visualizer = PosTagVisualizer(
+        ax, tagset, colors, colormap, **kwargs
+    )
+
+    # Fit and transform the visualizer (calls draw)
+    visualizer.fit(X, y, **kwargs)
+
+    # Return the axes object on the visualizer
+    return visualizer.ax
```
bbengfort: Part of #600 is to refactor quick methods to return the visualizer instead of the ax - not a big deal but if you want to get ahead of the game, I'd appreciate it!

docs/api/text/postag.rst
```diff
@@ -0,0 +1,164 @@
...
+                (',', ','),('dark', 'JJ'),('and', 'CC'),('deep', 'JJ'),(',', ','),
+                ('But', 'CC'),('I', 'PRP'),('have', 'VBP'),('promises', 'NNS'),
+                ('to', 'TO'),('keep', 'VB'),(',', ','),('And', 'CC'),('miles', 'NNS'),
+                ('to', 'TO'),('go', 'VB'),('before', 'IN'),('I', 'PRP'),
+                ('sleep', 'VBP'),(',', ','),('And', 'CC'),('miles', 'NNS'),
+                ('to', 'TO'),('go', 'VB'),('before', 'IN'),('I', 'PRP'),
+                ('sleep', 'VBP'),('.', '.')
+                ]
+        ]
+    ]
```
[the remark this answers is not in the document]
rebeccabilbro: :+1: I've added a SpaCy example.

docs/gallery.py
```diff
@@ -378,6 +378,65 @@ def dispersion():
...
+                ('But', 'CC'),('I', 'PRP'),('have', 'VBP'),('promises', 'NNS'),
+                ('to', 'TO'),('keep', 'VB'),(',', ','),('And', 'CC'),('miles', 'NNS'),
+                ('to', 'TO'),('go', 'VB'),('before', 'IN'),('I', 'PRP'),
+                ('sleep', 'VBP'),(',', ','),('And', 'CC'),('miles', 'NNS'),
+                ('to', 'TO'),('go', 'VB'),('before', 'IN'),('I', 'PRP'),
+                ('sleep', 'VBP'),('.', '.')
+                ]
+        ]
+    ]
+    oz = PosTagVisualizer()
```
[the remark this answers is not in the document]
rebeccabilbro: :+1:

docs/gallery.py
```diff
@@ -378,6 +378,65 @@ def dispersion():
...
+    oz.fit(tagged_stanzas)
```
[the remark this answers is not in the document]
rebeccabilbro: Tight layout definitely does something, but maybe not what we want it to do :/  I wonder if this is related to #613?

without tight layout
postag

with tight layout
postag

tests/test_text/test_postag.py
```diff
@@ -21,68 +21,241 @@
...
+## PosTag Utils
+##########################################################################
 
 def check_nltk_data():
     """
     Returns True if NLTK data has been downloaded, False otherwise
     """
     try:
-        nltk.data.find('corpora/treebank')
+        nltk.data.find('corpora')
```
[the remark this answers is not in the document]
rebeccabilbro: At the time I was thinking I might include other tagsets from nltk, but ended up just with penn treebank. I'll change it back

tests/test_text/test_postag.py
```diff
@@ -21,68 +21,241 @@
...
         pytest.xfail("error occured because nltk postag data is not available")
 
+def check_spacy_data():
+    """
+    Returns True if SpaCy data has been downloaded, False otherwise
+    """
+    try:
+        spacy.load('en_core_web_sm')
+        return True
+    except OSError: # should this be OSError?
```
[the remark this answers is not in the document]
rebeccabilbro: Not sure what the best way to handle auxiliary data dependencies like this and SpaCy isn't really meant to be on the critical path, so I've commented it out on the requirements.

tests/test_text/test_postag.py
```diff
@@ -21,68 +21,241 @@
...
     @pytest.mark.skipif(nltk is None, reason="test requires nltk")
+    @pytest.mark.skipif(spacy is None, reason="test requires spacy")
     def test_integrated_postag(self):
         """
         Assert no errors occur during postag integration
         """
-
         # Fail if data hasn't been downloaded
         check_nltk_data()
+        check_spacy_data()
```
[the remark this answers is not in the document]
rebeccabilbro: :+1: updated

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
+        colormap=None, 
+        colors=None, 
+        **kwargs
+    ):
+        super(PosTagVisualizer, self).__init__(ax=ax, **kwargs)
+       
+        self.tagset_names = TAGSET_NAMES
+        
+        if tagset not in self.tagset_names:
+            raise NotImplementedError(
```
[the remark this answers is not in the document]
rebeccabilbro: :+1:

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
 
 from yellowbrick.text.base import TextVisualizer
+from yellowbrick.style.colors import resolve_colors
+
+
+##########################################################################
+# Part-of-speech tag punctuation and labels
+##########################################################################
+
+PUNCT_LIST = [".", ":", ",", "``", "''", "(", ")", "#", "$"]
```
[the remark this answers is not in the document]
rebeccabilbro: NLTK's implementation of Penn Treebank automatically converts all sentence closers, including `!` and `?`, to periods. For now, these tags only get used for Penn Treebank (SpaCy turns everything to `PUNCT`), though I imagined that this may change as other tagsets got included. To ensure no one gets confused about this, I've added a note and changed the var name. Also changed `list` to `frozenset`

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,371 @@
...
+        Parameters
+        ----------
+        X : list or generator
+            Should be provided as a list of documents or a generator
+            that yields a list of documents that contain a list of 
+            sentences that contain (token, tag) tuples.
+
+        y : ndarray or Series of length n
+            An optional array or series of target or class values for
+            instances.
```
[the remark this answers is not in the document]
rebeccabilbro: 👍

yellowbrick/text/postag.py
rebeccabilbro: Changed to `self.pos_tag_counts_`

yellowbrick/text/postag.py
rebeccabilbro: Changed to jump table to avoid incurring technical debt.

yellowbrick/text/postag.py
rebeccabilbro: See above response, but NLTK's implementation of Penn Treebank automatically converts all sentence closers to periods. For now, these tags only get used for Penn Treebank (SpaCy turns everything to `PUNCT`). I imagined that this may change as other tagsets got included? To ensure no one gets confused about this, I've added a note and changed the var name per your suggestion.

yellowbrick/text/postag.py
rebeccabilbro: See my note and examples above. Essentially, `plt.tight_layout` does not seem to have the desired effect. Not sure what to do, but this does seem likely related to #613

yellowbrick/text/postag.py
rebeccabilbro: 👍 Removed.

yellowbrick/text/postag.py
rebeccabilbro: 👍 Added to docstring.

yellowbrick/text/postag.py
rebeccabilbro: Got it; fixed.

docs/gallery.py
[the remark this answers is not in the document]
lwgray: @rebeccabilbro as you might recall I reopened #613 because after mpl updates I was still having problems with ConfusionMatrix labels being cutoff.  I think I might have found a partial solution.  I found that adding the keyword `bbox_inches='tight'` to poof fixed the cutoff.  Maybe it will work in this case, give it a try.

```
iris_cm = ConfusionMatrix(
    model, classes=classes,
    label_encoder={0: 'setosa', 1: 'versicolor', 2: 'virginica'}
)

iris_cm.fit(X_train, y_train)
iris_cm.score(X_test, y_test)

i=iris_cm.poof(bbox_inches='tight')
```

Before `bbox_inches='tight'`
image


After `bbox_inches='tight'`
image

docs/api/text/postag.rst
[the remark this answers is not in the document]
bbengfort: Ah, sorry for the confusion - but thanks for adding the universal tags example! I've realized I was a little unclear by what I meant; I was referring to the explicit creation of the list of list of token, tag tuples as reference data to what's expected by the visualizer. The secondary examples I was referring to were more something like:

```python
.. code-block:: python

    import nltk

    text = [
        "The woods are lovely, dark and deep,",
        "But I have promises to keep,", 
        "And miles to go before I sleep,", 
        "And miles to go before I sleep".,
    ]

    # Tokenize and tag text
    tagged_stanzas = [nltk.pos_tag(nltk.word_tokenize(sent)) for sent in text]

    # Create the visualizer 
    viz = PostagVisualizer()
    viz.fit(tagged_stanzas)
    viz.poof()
```

So I guess what I was saying was that I do really like the hardcoded example to show what kind of data was expected, but perhaps afterward we could create some code snippets to show how this is probably used in real-life, e.g. directly on strings with nltk and spacy -- but not in `.. plot::` directive so that the example code didn't execute and require extra dependencies for our documentation. 

Just wanted to clarify my comments - I'll leave it to you if you want to leave it like it is, that's perfectly fine!

yellowbrick/text/postag.py
bbengfort: Ah, that's too bad that didn't work. @lwgray came up with a solution - but I don't think that it should block this PR; @lwgray would you be willing to try your idea on the gallery when we merge this in?

docs/gallery.py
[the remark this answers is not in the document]
bbengfort: @lwgray thanks for the update -- and I'm glad this is a solution to the confusion matrix as well. My suggestion is to create another PR (after this is merged) to address #613 using this solution in both `ConfusionMatrix` as well as `PosTagVisualizer` - would you be interested in doing that?

docs/api/text/postag.rst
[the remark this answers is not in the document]
bbengfort: Ah, sorry for the confusion - but thanks for adding the universal tags example! I've realized I was a little unclear by what I meant; I was referring to the explicit creation of the list of list of token, tag tuples as reference data to what's expected by the visualizer. The secondary examples I was referring to were more something like:

```python
.. code-block:: python

    import nltk

    text = [
        "The woods are lovely, dark and deep,",
        "But I have promises to keep,", 
        "And miles to go before I sleep,", 
        "And miles to go before I sleep".,
    ]

    # Tokenize and tag text
    tagged_stanzas = [nltk.pos_tag(nltk.word_tokenize(sent)) for sent in text]

    # Create the visualizer 
    viz = PostagVisualizer()
    viz.fit(tagged_stanzas)
    viz.poof()
```

So I guess what I was saying was that I do really like the hardcoded example to show what kind of data was expected, but perhaps afterward we could create some code snippets to show how this is probably used in real-life, e.g. directly on strings with nltk and spacy -- but not in `.. plot::` directive so that the example code didn't execute and require extra dependencies for our documentation. 

Just wanted to clarify my comments - I'll leave it to you if you want to leave it like it is, that's perfectly fine!

tests/requirements.txt
```diff
@@ -15,5 +15,6 @@ requests>=2.18.3
 
 # Optional Testing Dependencies
 nltk>=3.2
+# spacy>=2.0.18
```
bbengfort: Ok, I think this is a good choice - but if you create a follow on issue for extensions to this visualizer, could we add a note to it to revisit this? At that point in the future, hopefully we'll have seen some usage and we can evaluate whether spaCy is critical or not. Maybe we've also created more spaCy-specific visualizations as well or we can use that as an opportunity to review if there are any to be created! :-)

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,370 @@
...
+        Specify a colormap to color the parts-of-speech.
     kwargs : dict
-        Pass any additional keyword arguments to the super class.
-    cmap : dict
-        ANSII colormap
+        Pass any additional keyword arguments to the PosTagVisualizer.
+
+    Attributes
+    ----------
+    pos_tag_counts_: dict
```
bbengfort: 🎉

yellowbrick/text/postag.py
```diff
@@ -20,125 +20,370 @@
...
+                    elif tag in self.punct_tags:
+                        self.pos_tag_counts_["punctuation"] += 1
+                    elif tag == "TO":
+                        self.pos_tag_counts_["infinitive"] += 1
+                    elif tag == "UH":
+                        self.pos_tag_counts_["interjection"] += 1
+                    elif tag == "SYM":
+                        self.pos_tag_counts_["symbol"] += 1
+                    else:
+                        self.pos_tag_counts_["other"] += 1
```
bbengfort: I suppose I should have also pointed out these case statements as a target for jump table conversion as well - but it shouldn't block this PR - but perhaps could be mentioned in a follow-up issue if one is created?

docs/gallery.py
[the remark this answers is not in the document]
rebeccabilbro: This seems to work &mdash; thanks @lwgray! Just a heads up that in `gallery.py`, the call to `poof` happens in a specialized function meant to work for all the visualizers. As a workaround, we could add optional kwargs to that function `def savefig(viz, name, gallery=GALLERY, **kwargs):` so that we can do:

```python
    oz = PosTagVisualizer(ax=newfig())
    oz.fit(tagged_stanzas)
    savefig(oz, "postag", bbox_inches="tight")
```

docs/api/text/postag.rst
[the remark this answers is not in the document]
rebeccabilbro: I'm sorry, I misunderstood what you were asking for. I'll add this to the follow-up task?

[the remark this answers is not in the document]
bbengfort: It's completely my fault! And that would be great!

docs/gallery.py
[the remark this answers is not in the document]
lwgray: thanks @rebeccabilbro for pointing this out and writing up an example...I had pointed this out to @bbengfort earlier.

PR #770, opened 2019-02-22
Fixes Label cutoff in saved images of PosTag and ConfusionMatrix Viz

lwgray:
Addresses concerns in #768 

For the gallery, visualizer produced images are saved to disk.  We found
in particular that the saved images for PosTagVisualizer and ConfusionMatrix were being
produced with labels cutoff (see below).  This PR remedies this with bbox_inches
being set to tight (see below).

postag_gallery

image

image

image


bbengfort:
@lwgray I'm sorry that it's taken so long for us to get to this PR - this was a very difficult one since modifying `poof()` had implications for our visual pipelines and our fig/axes management and I wanted to think deeply about it. I think your solution is very good - and I wanted to come back to it, particularly because of the discussion in #826. 

Right now, I would be willing to merge this as it is - but we'd have to remember that these two `poof` modifications exist as we continue to deal with the figure/axes API decisions moving into v1. 

Another option would be to modify the [base class visualizer poof method](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/base.py#L187) as follows:

```python
class Visualizer(BaseEstimator):

    def poof(self, outpath=None, clear_figure=False, tight=False, **kwargs):
        ...
        # Finalize the figure
        self.finalize()

        if tight and "bbox_inches" not in kwargs:
            kwargs["bbox_inches"] = kwargs.get("bbox_inches", "tight")

        if outpath is not None:
             ...      
```

This would allow us to optionally specify `tight=True` for poof in PosTag and ConfusionMatrix, e.g. for the gallery and the docs, but we wouldn't have to remember that we made those changes generally. We would then support users who had an issue saying to pass `tight=True` to `poof()`. 

Honestly, either way is fine with me - do you have any preferences?

bbengfort:
@lwgray also, I've merged master to check to make sure that there are no other conflicts since I let this one go for a while (again, I'm so very sorry) ... please do a pull in your feature branch before making any changes!

lwgray:
@bbengfort Just merge as is; we will just have to remember.

PR #772, opened 2019-02-26
Add plot directive to CVScores docs

pdamodaran:
The following changes were made:

* deleted the Python script previously used to generate the images
* deleted the original image files
* updated the `.rst` file with the new plot directive, including adding the `:context: close-figs` and `:alt:` options

The second image does not match the original


lwgray:
@pdamodaran This is a such a wonderful PR but before I go any further, could you explain the changes in jointplot and PCA files?

PR #776, opened 2019-03-02
Addresses Issue #741 - [elbow visualizer] Error in distortion score calculation #762

pdamodaran:
As reported by @maximilian1064 and confirmed by @zjpoh, the current version of the elbow visualizer was incorrectly calculating the distortion score.

The line distortion += distances.mean() needed to be replaced with distortion += distances.sum()

This pull request makes the fix and it also updates the expected results of the tests in test_elbow.py based on the fix made to elbow.py.


docs/README.md
```diff
@@ -68,3 +68,17 @@ The primary section is wrapped up with a discussion about how to interpret the v
 
 There are several other places where you can list your visualizer, but to ensure it is included in the documentation it *must be listed in the TOC of the local index*. Find the `index.rst` file in your subdirectory and add your rst file (without the `.rst` extension) to the `..toctree::` directive. This will ensure your documentation is included when it is built.
 
+## Troubleshooting
```
rebeccabilbro: Thanks for capturing this strange behavior you observed @pdamodaran! After thinking about it a bit more, I'm thinking this is better suited to a bug report (in case it has something to do with the matplotlib backend). Would you mind opening up a bug report?

Here in the README let's add here the (below) instructions for how to regenerate the gallery images, so that it will be easier for future contributors to figure it out! (FYI - I tried to push these changes to your branch but it looks like you've maybe disabled the feature that allows maintainers to push to your PR branches? Might be a useful thing to enable in the future!)

## Generating the Gallery

In v1.0, we have adopted Matplotlib's [plot directive](https://matplotlib.org/devel/plot_directive.html) which means that the majority of the images generated for the documentation are generated automatically. One exception is the gallery; the images for the gallery must still be generated manually. 

If you have contributed a new visualizer as described in the above section, please also add it to the gallery, both to `docs/gallery.py` and to `docs/gallery.rst`. (Make sure you have already installed Yellowbrick in editable mode, from the top level directory: `pip install -e .`)

If you want to regenerate a single image (e.g. the elbow curve plot), you can do so as follows:

```bash
$ python docs/gallery.py elbow
```

If you want to regenerate them all (_note: this takes a long time!_)

```bash
$ python docs/gallery.py all
```

PR #777, opened 2019-03-02
Added plot directive for all target visualizers

pdamodaran:
Modified all target visualizers to use the plot directive

BalancedBinningReference
ClassBalance
FeatureCorrelation

Original images and python scripts have been deleted.


docs/api/target/feature_correlation.rst
```diff
@@ -49,36 +52,39 @@ See `scikit-learn documentation <http://scikit-learn.org/stable/modules/generate
...
     feature_names = np.array(data['feature_names'])
     X_pd = pd.DataFrame(X, columns=feature_names)
 
     feature_to_plot = ['alcohol', 'ash', 'hue', 'proline', 'total_phenols']
 
+
     visualizer = FeatureCorrelation(method='mutual_info-classification',
                                     feature_names=feature_to_plot, sort=True)
-    visualizer.fit(X_pd, y, random_state=0)
+    visualizer.fit(X_pd, y, random_state=42)
```
lwgray: for my understanding @Kautumn06 do we want to change random_state values where they are already set?  random_state=42 does bring unity across the docs

pdamodaran: @Kautumn06 suggested that we should change to random_state=42 for consistency.

pdamodaran:
@lwgray - I have updated the rst files with your suggested changes.  Thanks for your feedback!

docs/api/target/feature_correlation.rst
Kautumn06: Hi @pdamodaran, In #775 I recommended setting the `random_state=42` when the parameter had not previous been set to a specific value. However, in this case `random_state` had already been set to `0` so it should not be changed.

pdamodaran: @Kautumn06 - I have changed this back to random_state = 0

PR #786, opened 2019-03-15
Add plot directive to Residuals documentation.

navarretedaniel:
# Residuals (ResidualsPlot)

## Overview
The following changes were made to address the plot directive project referenced in #687:

1. Deleted `residuals.py` script previously used to generate the images

2. Deleted the original image files (`residuals.png` and `residuals_no_hist.png`).

3. Updated `residuals.rst` with the new plot directive, including adding the `:context: close-figs and :alt:` options

## Results
`random_state=42` has been added in `X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=42)`.  As a result, the following images are now reproducible:

- [x] `residuals.png`

- [x]  `residuals_no_hist.png`


docs/api/regressor/residuals.rst
```diff
@@ -3,58 +3,46 @@
 Residuals Plot
 ==============
 
-Residuals, in the context of regression models, are the difference between the observed value of the target variable (y) and the predicted value (ŷ), e.g. the error of the prediction. The residuals plot shows the difference between residuals on the vertical axis and the dependent variable on the horizontal axis, allowing you to detect regions within the target that may be susceptible to more or less error.
+Residuals, in the context of regression models, are the difference between the observed value of the target variable (y) and the predicted value (ŷ), i.e. the error of the prediction. The residuals plot shows the difference between residuals on the vertical axis and the dependent variable on the horizontal axis, allowing you to detect regions within the target that may be susceptible to more or less error.
```
lwgray: great catch, i.e (in other words). is often confused with e.g.(for example)

docs/api/regressor/residuals.rst
```diff
@@ -3,58 +3,46 @@
...
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=42)
 
     # Instantiate the linear model and visualizer
-    ridge = Ridge()
-    visualizer = ResidualsPlot(ridge)
+    model = Ridge()
+    visualizer = ResidualsPlot(model)
 
-    visualizer.fit(X_train, y_train)  # Fit the training data to the model
+    visualizer.fit(X_train, y_train)  # Fit the training data to the visualizer
```
lwgray: here you are fitting to the model.  `visualizer.fit(X_train, y_train)` actually calls sklearn estimator(Ridge) fit - @Kautumn06 correct me if I am wrong

rebeccabilbro: Interesting question! I suppose it does both, since [`fit` calls `score`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/regressor/residuals.py#L463), which [then calls `draw`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/regressor/residuals.py#L498)! In any case, I think it's okay to leave this as is

Issue #797, opened 2019-03-29
normalize ='minabs' not working for ParallelCoordinates


haleemason:
**Describe the bug**
When using `visualizer = ParallelCoordinates(normalize ='minabs')` the normalize parameter  ='minabs' is not working.

**To Reproduce**

```python
from yellowbrick.features import ParallelCoordinates

# Load data set (don't need to do again, included for completeness)
X, y = load_occupancy()

# Specify the features of interest and the classes of the target
features = [
    "temperature", "relative humidity", "light", "CO2", "humidity"
]
classes = ["unoccupied", "occupied"]

# Instantiate the visualizer
visualizer = ParallelCoordinates(
    classes=classes, features=features,
    normalize='minabs', # normalize options: minmax, minabs, standard, l1, or l2
    sample=0.05, shuffle=True, size=(1080, 720)
)

# Fit the visualizer and display it
visualizer.fit_transform(X, y)
visualizer.poof()
```

**Dataset**
Did you use a specific dataset to produce the bug? Where can we access it?
`python -m yellowbrick.download`
`X, y = load_occupancy()`

**Expected behavior**
I was expecting the visualizer plot to be normalized using the `minabs` method. 

**Traceback**

```
---------------------------------------------------------------------------
YellowbrickValueError                     Traceback (most recent call last)
<ipython-input-21-c85f2a8aee8c> in <module>()
      3     classes=classes, features=features,
      4     normalize='minabs', # This time we'll specify a normalizer
----> 5     sample=0.05, shuffle=True, size=(1080, 720)
      6 )
      7 

~/repos/dev/yellowbrick/yellowbrick/features/pcoords.py in __init__(self, ax, features, classes, normalize, sample, random_state, shuffle, color, colormap, alpha, fast, vlines, vlines_kwds, **kwargs)
    260             raise YellowbrickValueError(
    261                 "'{}' is an unrecognized normalization method"
--> 262                 .format(normalize)
    263             )
    264 

YellowbrickValueError: 'minabs' is an unrecognized normalization method
<Figure size 1080x720 with 0 Axes>
```

**Desktop (please complete the following information):**
 - OS: `macOS`
 - Python Version:  `[Python 3.6.5 |Anaconda, Inc.| (default, Apr 26 2018, 08:42:37)`
`[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin`
[...]

PR #799, opened 2019-03-30
Update pcoords.rst

pswaldia:
In the documentation of Parallel Coordinates, one of the possible value for normalize parameter is mentioned as 'minabs' , which is not being implemented in yellowbrick.  Current supported options are 'minmax', 'maxabs', 'standard', 'l1'  and 'l2' according to _pcoords.py_ . This PR corrects the documentation in refrence to issue #797. The reason for non inclusion of 'minabs' is the non availability of _MinAbsScaler_ in _sklearn.preprocessing_ .


rebeccabilbro:
Thanks so much for the quick response here and on #797 @pswaldia!  @haleemason's got us covered & the fix is thankfully merged in now, so going to close this PR. If you ever get a hankering to implement `minabs`; let us know!!

Issue #437, opened 2018-05-18
Improvements to the Manifold Visualizer


bbengfort:
The following improvements can be made to the manifold visualizer:

- [ ] describe the number of features the embedding represents (e.g. the number of columns in X) and optionally the number of points in the dataset 
- [ ] create a documentation example for clustering 
- [ ] add `frameon=True` to legend 
- [ ] create documentation example using the standard scalar 
- [ ] get the visual tests to work on windows/decrease tolerance 
- [ ] decrease tolerance of `test_manifold_pandas` (not sure why the tolerance is so high)
- [ ] investigate the `n_neighbors` attribute; see below.

jimmyshah:
Hi! I wanted to work on the frameon feature? Is that still up for grabs?

lwgray:
hi @jimmyshah Thanks for your intrest in contributing to yellowbrick.  however we rarely assign issues.  But, You are more that welcome to open a PR for this addition.  I look forward to seeing you contributing.

jimmyshah:
Thank you for the quick response! Sounds good to me, I'm excited to contribute!

jimmyshah:
Hello again! I was interested in working on the n_neighbors attribute issue, is that one still available?

lwgray:
hi @jimmyshah Let me again Thank you for your  last contribution to yellowbrick. In regard to your currest ask, we rarely assign issues.  But, You are more that welcome to open a PR for this addition. I look forward to seeing your next contribution.

rebeccabilbro:
Hey there @jimmyshah and welcome back! The maintainers don't generally assign issues to contributors (we're only 5-6 people so that would be a bit too much management overhead for us), so the best strategy is to check if there's already an [open PR addressing the issue](https://github.com/DistrictDataLabs/yellowbrick/pulls) (which in this case, there isn't... yet! 😉) Look forward to seeing your next PR!

jimmyshah:
Thank you for the assistance @rebeccabilbro and @lwgray, I'm happy to contribute! I'll make a PR request shortly, after I've had a chance to make some code changes 🙂

bbengfort:
@naresh-bachwani -- please see this issue as you work on #930; @rebeccabilbro pointed out in #927 that we might be able to close this as you work through it.

bbengfort:
@naresh-bachwani have we managed to make progress on this? Can we close the issue?

PR #809, opened 2019-04-10
Fixes part of #437

jimmyshah:
Adds frameon=true from #437 , I closed the previous PR


jimmyshah:
I think I'm having some trouble understanding the problem with my submission. Would it be possible to receive some feedback?

rebeccabilbro:
Hey there @jimmyshah and thanks for checking out Yellowbrick and for helping us work through our (ever growing 😅) backlog of issues! 

Here's what's happening with your PR; you've globally modified the `manual_legend` utility, which is used by many of our visualizers. As a result, your change has broken many of those existing visualizers (for instance, `ParallelCoordinates`, [which already passes in a frameon argument](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/features/pcoords.py#L517)). 

The request in #437 is to do something similar with `Manifold` as is currently done in `ParallelCoordinates` (e.g. passing the `frameon` kwarg [into the call to `manual_legend`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/features/manifold.py#L403), which already knows how to consume any `**legend_kwargs` that get passed in), not to modify the `manual_legend` utility function directly. 

Note that when you make this change, it should fix all the CI errors, but we may need to work through some new broken image tests for `Manifold` &mdash; if you want to get a jump on figuring those out, you can read up on our visual image tests [here](https://www.scikit-yb.org/en/develop/contributing/developing_visualizers.html#testing) and [here](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/README.md#image-comparison-tests)!

Apologies for the confusion, and hope this is helpful!

jimmyshah:
Thank you so much for the elaborate response @rebeccabilbro !  It was really helpful and I understand the issue more clearly now. I'm going to try working on this again in a bit.

jimmyshah:
@rebeccabilbro Thank you again for your all of your help! It looks like my commit is passing the Travis CI check according to the site, but it is still showing it as being queued here. Is there something else that I need to do?

jimmyshah:
@rebeccabilbro This is so wonderful to hear! Thank you so much, and I look forward to future contributions. I can use the feature branch convention going forward as well, my apologies about that.

PR #807, opened 2019-04-05
Addressing issue #589: Adding alpha params to PCA

naresh-bachwani:
The following changes were made to PCA referenced #589 :
1. Added new parameter alpha to `pca.py `to specify the opacity.
2. Modified `test_pca.py `

The result is attached below:
Screenshot (144)

Issue #589, opened 2018-08-26
Add alpha (opacity) param to PCADecomposition


rebeccabilbro:
**Describe the solution you'd like**
Once #569 is done, add an `alpha` (opacity/translucency) param to the PCADecomposition visualizer. See also #558 and #574. 

Note that enabling the user to pass in an `alpha` value and decrease the opacity of points plotted in a matplotlib scatterplot will also decrease the opacity of the colors in the plot's legend. The workaround is to use the [`manual_legend` util](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/draw.py#L29), as we've done in the other scatterplot-based visualizers that have `alpha` params.

PR #811, opened 2019-04-11
Addressing issue #589: Adding alpha params to PCA

naresh-bachwani:
The following changes were made to PCA referenced #589 :  
**This PR was partially reviewed in #807** 
1. Added new parameter alpha to `pca.py `to specify the opacity.
2. Modified `test_pca.py ` to add a test for alpha
The result is attached below:
Screenshot (144)


naresh-bachwani:
Hello @lwgray, 
I have made the changes that you suggested in #807.

tests/test_features/test_pca.py
```diff
@@ -188,3 +191,26 @@ def test_scale_true_3d_execption(self):
...
+    def test_alpha_param(self, mock_sca):
+        """
+        Test that the user can supply an alpha param on instantiation
+        """
+        # Instantiate a prediction error plot, provide custom alpha
+        params = {'alpha': 0.3, 'proj_dim': 2, 'random_state': 9932}
+        visualizer = PCADecomposition(**params).fit(self.dataset.X)
+        pca_array = visualizer.transform(self.dataset.X)
+        alpha=0.3
+        assert visualizer.alpha == alpha
```
lwgray: 0.3 doesn't need to be assigned to alpha.  You should test directly ``assert visualizer.alpha == 0.3``

tests/test_features/test_pca.py
```diff
@@ -188,3 +191,26 @@ def test_scale_true_3d_execption(self):
         with pytest.raises(ValueError, match=e):
             pca = PCADecomposition(**params)
             pca.fit(X)
+    
+    @mock.patch('yellowbrick.features.pca.plt.sca', autospec=True)
```
lwgray: I am unsure about this test, I will wait on a review from @rebeccabilbro

yellowbrick/features/pca.py
```diff
@@ -99,6 +103,7 @@ def __init__(self,
                  proj_features=False,
                  color=None,
                  colormap=palettes.DEFAULT_SEQUENCE,
+                 alpha = 1,
```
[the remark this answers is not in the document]
rebeccabilbro: Tiny thing, but the spacing is off here. We shouldn't have spaces before and after the equals sign for keyword arguments.

yellowbrick/features/pca.py
```diff
@@ -118,7 +123,7 @@ def __init__(self,
             [('scale', StandardScaler(with_std=self.scale)),
              ('pca', PCA(self.proj_dim, random_state=random_state))]
         )
-
+        self.alpha=alpha
```
rebeccabilbro: Another spacing thing here (should be `self.alpha = alpha`)

yellowbrick/features/pca.py
```diff
@@ -280,7 +290,7 @@ def pca_decomposition(X, y=None, ax=None, features=None, scale=True,
     visualizer = PCADecomposition(
         ax=ax, features=features, scale=scale, proj_dim=proj_dim,
         proj_features=proj_features, color=color, colormap=colormap,
-        random_state=random_state, **kwargs
+        alpha=alpha, random_state=random_state,**kwargs
```
rebeccabilbro: It looks like the spaces were modified here? We should reinsert the space after the last comma and before `**kwargs`

tests/test_features/test_pca.py
rebeccabilbro: Yes, the test looks good! Patching pyplot's set current axes method ([`sca`](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.sca.html) ) ensures that when we make the mock on line 207 (which is used to called `scatter` and retrieve the `alpha`), matplotlib doesn't complain about it.

naresh-bachwani:
I changed the baseline images. The test passes on my local computer but is failing here!

rebeccabilbro:
Ok, thanks @naresh-bachwani &mdash; we may just need to tweak the comparison tolerances. We'll have a look this weekend and get back to you!

naresh-bachwani:
Ok! Thanks!

lwgray:
@naresh-bachwani i advise that you increase the tolerance for the tests that didn’t pass

naresh-bachwani:
I will increase the tolerance but I had a question.

All these tests pass on my local computer so how can I know in advance if my tests would pass after making a push? 

Capture1
Capture

lwgray:
what OS are you using locally?

naresh-bachwani:
Windows10

lwgray:
@naresh-bachwani That might be the reason for the difference .  We have known that the test run differently in Windows than in Linux ( the environment our test run in).  There isnt a way to know ahead of time if test will run successfully on Travis.  The goal is to make them pass on Travis despite what is happening locally.

naresh-bachwani:
Thanks for the help!

rebeccabilbro:
> All these tests pass on my local computer so how can I know in advance if my tests would pass after making a push?

@naresh-bachwani &mdash; yes, this unfortunately is one of the challenges of doing visual comparison tests. Different operating systems have *slightly* different versions of fonts, colors, etc, which can cause our tests to pass locally (since the images match on our local OSs) but fail mysteriously in the CI. Our CI solution is to use both Travis and AppVeyor, which allows us to see what will happen in both a linux and windows virtual machine. 

Generally my strategy is to run the tests locally, copy and commit any new baseline images, and then push to the PR branch to trigger the CI builds. If any builds fail, the first thing I check is if there are any `random_state` parameters I can set in the tests to make them run more deterministically (e.g. [here, where we set it to 9932](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_features/test_pca.py#L80)). Most scikit-learn estimators seem to have them. 

If the `random_state` has already been set, I'll inspect the Travis or AppVeyor build report to see what the RMS error of the test failure was, and round up to set the `tol` for the `self.assert_images_similar` call in the failing test. Occasionally it takes a few times, but it looks like you've figured out good tolerance values in this case to ensure the tests will pass on a range of operating systems! Nice work and thanks for working through this!

naresh-bachwani:
Hello @rebeccabilbro,
Thanks for your help! I had a question. How can I find the random state? Is it through some loop till RMSE decreases below threshold value?

rebeccabilbro:
> Hello @rebeccabilbro,
Thanks for your help! I had a question. How can I find the random state? Is it through some loop till RMSE decreases below threshold value?

Great question @naresh-bachwani! The `random_state` is actually a scikit-learn thing ([here's where it's defined for pca, for example](https://github.com/scikit-learn/scikit-learn/blob/7b136e9/sklearn/decomposition/pca.py#L193)) that's just a seed for a random number generator. By default, `random_state` picks a value at random to seed the relevant algorithm(s) with. But, if you set the `random_state` to a specific integer value, it will always start with the same seed, which is very helpful for the tests (less so for actual machine learning though ;) ). Generally people just pick an arbitrary number; 42 is popular because many OSS contributors are also [Adams fans](https://en.wikipedia.org/wiki/Phrases_from_The_Hitchhiker%27s_Guide_to_the_Galaxy#The_number_42) :D

naresh-bachwani:
Ok so 9932 is just an arbitrary number?

rebeccabilbro:
> Ok so 9932 is just an arbitrary number?

Yep!

PR #813, opened 2019-04-12
Add elbow detection using the "kneedle" method to Elbow Visualizer

### LINKED ISSUE #764: Add elbow detection using the "kneedle" method to Elbow Visualizer
It would be nice if we could (optionally) annotate the "best" k in the elbow method by detecting the "knee" of the curve using the "kneedle" method described in: https://github.com/arvkevi/kneed. This feature could be added either by adding `kneed` as an _optional_ dependency or by reimplementing part of the kneedle code in Yellowbrick.

pswaldia:
### I. Are you merging from a feature branch into develop?
Yes
### II. Summarize your PR 
This PR fixes #764 that aimed to include a feature to annotate the optimal value of 'k' using _kneedle_ method described in https://github.com/arvkevi/kneed . 

I have solved the issue in the following way:   
1. By using  `kneed`  library which can be added as an optional dependancy.
2. By plotting a vertical line (black dashed)  showing optimal K value.    

### III. Include a sample plot
plot

In the above plot k=7 is the optimal K.

### IV. List any TODOs or questions
Still to do:
- Upon getting feedback from the maintainers , it can be decided on whether it is to be added as an optional feature or regular feature.          

- Is the legend required or not. ? It can be removed if not required.


lwgray:
@pswaldia Welcome and Thanks for opening a PR.  We are wading through our backlog of issues and will get to your PR asap.

pswaldia:
@lwgray Sure. Thanks for the fast reply.

rebeccabilbro:
Hey there @pswaldia and thanks for jumping in to address #764! 

After taking a look through `kneed`, I think this will be a *very* useful functionality to integrate into our `KElbowVisualizer`! So useful, in fact, that I am going to propose that instead of using `kneed` as an optional dependency, we instead reimplement the code in [`knee_locator.py`](https://github.com/arvkevi/kneed/blob/master/kneed/knee_locator.py) and integrate it *directly* into our [`KElbowVisualizer`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/cluster/elbow.py#L118) (we should also make sure to give @arvkevi a big shout-out in our docs!). Since `kneed`'s dependencies are all already dependencies of YB, and since the code required to generate the elbows is only ~100 lines (not including tests), I'm hoping this won't be too heavy of a lift.

Full disclosure: My rationale for reimplementation over optional dependency is also partly in the interest of the long-term maintainability of our codebase.  It will be **much** easier on our end to maintain the code if it's part of YB. Basically, from the maintainer perspective, fewer dependencies == fewer headaches down the line :D 

Let us know if you're game to try the reimplementation @pswaldia!

pswaldia:
@rebeccabilbro Thanks for the response. I am ready to reimplement the knee_locater functionality in yb. I'll be updating  the pull request soon.
Thanks.

arvkevi:
This is so cool to see `kneed` even mentioned by yb 😄 
The reimplementation rationale makes sense, case in point, I'm planning a release soon that will implement some [significant changes](https://github.com/arvkevi/kneed/pull/26) (although they wouldn't affect this PR). 

I'm not an expert on OSS licenses, but if you go the reimplementation route it might be best to include the license and copyright attribution? Only mentioning this based on what I've picked up on in my OS adventures.

Thanks for the shoutout 👍

bbengfort:
@arvkevi thanks for chiming in - I was going to reach out to you about this, but I'm super excited that you're onboard to have us include `knee_locater.py` in our source. Let me know if you'd like to chat more directly over email or elsewhere but here is what I propose:

You've licensed `kneed` under a very permissive BSD-3 license. Yellowbrick is also licensed in a permissive fashion, using the Apache 2 license. Because our license is _slightly_ more restrictive than yours, I believe that our license will cover yours in an acceptable fashion. Further to meet the obligations of your license we propose to add the following preamble to:

- `yellowbrick/utils/kneed.py` 
- `tests/test_utils/test_kneed.py`

as follows:

```python
# yellowbrick.utils.kneed
# A port of the knee-point detection package, kneed.
#
# Author:   Kevin Arvai
# Author:   Pradeep Singh 
# Created:  Mon Apr 15 09:43:18 2019 -0400
#
# Copyright (C) 2017 Kevin Arvai
# All rights reserved.
# Redistribution and use in source and binary forms, with or without modification, 
# are permitted provided that the following conditions are met:
# 
# 1. Redistributions of source code must retain the above copyright notice, this list
# of conditions and the following disclaimer.
#
# 2. Redistributions in binary form must reproduce the above copyright notice, this 
# list of conditions and the following disclaimer in the documentation and/or other 
# materials provided with the distribution.
#
# 3. Neither the name of the copyright holder nor the names of its contributors may 
# be used to endorse or promote products derived from this software without specific
# prior written permission.
#
# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
# ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS 
# OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
# NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN
# IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
#
# ID: kneed.py [] pswaldia@no-reply.github.com $

"""
This package contains a port of the knee-point detection pacakge, kneed by Kevin Arvai and
hosted at https://github.com/arvkevi/kneed. This port is maintained with permission by the 
Yellowbrick contributors.
"""
```

We will also link to you and say thank you in our documentation when we get to that point. Thoughts about the above proposal?

bbengfort:
@pswaldia thank you for taking on this challenge! I will be reviewing your PR when you push the kneedle code and we can work together to make sure this is implemented correctly. As noted above, I think the best way to port the code would be to include the `knee_locator.py` file with the preamble above in `yellowbrick/utils/kneed.py` -- the file rename gives a stronger nod to @arvkevi. We can then place `test_sample.py` into `tests/test_utils/test_kneed.py` with the preamble modified for that file. 

Note that the two tests that rely on the `DataGenerator` can probably be simplified as we don't have the same requirement to model the original paper. We can either simply move the `noisy_gaussian` function over, or simply omit those tests.

Finally, to answer your question about the legend, I think that it is helpful, but with more information, e.g. remove the `score` label and instead simply have a label that says `optimal k=7, score=18232` or something like that. 

Adding the knee annotation should be an optional kwarg, wtih default=True so our users can turn it off if they don't want to use it. 

Let me know if you have any other questions! And thank you again so much for contributing to Yellowbrick!

pswaldia:
@bbengfort I have pushed some commits implementing changes directed by you. Before moving to tests I would like to recieve feedback regarding the changes.
**Changes done:**

1. Added _kneed.py_ to _yellowbrick/utils/_ .
2. Added the knee annotation as an optional kwarg, wtih default=True so our users can turn it off if they don't want to use it. ( Please let me know if the name of the variable chosen is alright.)
3. Changes been made to legend. ( I am not sure how to include independent texts to the legend without any symbols. Would love to recieve feedback on that.)

```python
from sklearn.cluster import KMeans
from yellowbrick.cluster import KElbowVisualizer

model = KMeans()

visualizer = KElbowVisualizer(model, k=(4,12)) #default knee=True

visualizer.fit(X)    # Fit the data to the visualizer
visualizer.poof()    # Draw/show/poof the data
```

plt2

```python
from sklearn.cluster import KMeans
from yellowbrick.cluster import KElbowVisualizer

model = KMeans()

visualizer = KElbowVisualizer(model, k=(4,12), knee=False) #setting knee=False 

visualizer.fit(X)    # Fit the data to the visualizer
visualizer.poof()    # Draw/show/poof the data
```


plot1

Thanks.

arvkevi:
@bbengfort this is great, I'm excited to see this feature hit master 😄

yellowbrick/cluster/elbow.py
```diff
@@ -247,6 +252,9 @@ def fit(self, X, y=None, **kwargs):
 
         self.k_scores_ = []
         self.k_timers_ = []
+        self.kneedle=None
```
bbengfort: Does the `KneeLocator` store any state that we may need to keep around? If so, we should store it in a variable `self._knee_locator` (just to make things obvious to other YB contributors). If not, there is probably no reason to keep this object on the visualizer, simply make it a variable local to `fit()` which will be discarded when we're done fitting.

yellowbrick/cluster/elbow.py
```diff
@@ -247,6 +252,9 @@ def fit(self, X, y=None, **kwargs):
...
+        self.knee_value=None
+        self.score=None
```
bbengfort: This is great, thank you so much for storing these on the visualizer! I think users will be interested in directly accessing them. I've got two requests:

1. We should make these "learned" attributes and suffix them with an underscore, e.g. `self.knee_value_` and `self.score_`; also it is probably more informative if we name them `self.elbow_value_` and `self.elbow_score_` if that's ok -- see discussion above. 

2. We should document these in the docstring of the class. Also, it seems that `k_scores_`  and `k_timers_` are not documented, so would you mind also documenting them? See [icdm.py Line 123](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/cluster/icdm.py#L123) for an example of how to document the learned attributes. 

Finally (more on this later), these properties should only exist iff `self.locate_elbow==True`.

yellowbrick/cluster/elbow.py
```diff
@@ -262,6 +270,9 @@ def fit(self, X, y=None, **kwargs):
                 self.scoring_metric(X, self.estimator.labels_)
             )
 
+        self.kneedle=KneeLocator(self.k_values_,self.k_scores_,curve='convex',direction='decreasing')
+        self.knee_value=self.kneedle.find_knee()[0]
+        self.score=self.k_scores_[self.k_values_.index(self.knee_value)]
```
bbengfort: So one reason a user might make `locate_elbow=False` is if the `KneeLocator` is throwing an exception of some kind or is taking an unreasonably long time to run. These lines of code run no matter the `locate_elbow` flag - so might I recommend the following?

```python
        if self.locate_elbow:
            knee_locator = KneeLocator(...)
            self.elbow_value_ = knee_locator.find_knee()[0]
            self.elbow_score_ = self.k_scores_[self.k_values_.index(self.elbow_value_)]
        else:
            self.elbow_value_ = None
            self.elbow_score_ = None
```

This also gives information to developers about the learned state of the visualizer.

yellowbrick/cluster/elbow.py
```diff
@@ -262,6 +270,9 @@ def fit(self, X, y=None, **kwargs):
                 self.scoring_metric(X, self.estimator.labels_)
             )
 
+        self.kneedle=KneeLocator(self.k_values_,self.k_scores_,curve='convex',direction='decreasing')
+        self.knee_value=self.kneedle.find_knee()[0]
```
bbengfort: After taking a look at the `KneeLocator` code, I noticed that `self.knee, self.norm_knee, self.knee_x` are all set when the locator is initialized at `__init__`() therefore there is no need to call `find_knee()` again (which is potentially an expensive operation). You can just do `self.elbow_value_ = knee_locator.knee`. 

Also below you use `self.k_values_.index`, you may be able to use `locator.knee_x` instead, but I'm not 100% sure about that.

yellowbrick/cluster/elbow.py
```diff
@@ -262,6 +270,9 @@ def fit(self, X, y=None, **kwargs):
                 self.scoring_metric(X, self.estimator.labels_)
             )
 
+        self.kneedle=KneeLocator(self.k_values_,self.k_scores_,curve='convex',direction='decreasing')
```
bbengfort: I'm a bit concerned about the `curve` and `direction` parameters of the `KneeLocator` - depending on the `metric` parameter, the curve may be concave/increasing or convex/decreasing (and other, unimplemented metrics may have a different effect). 

This leads me to believe that we need to do one of the following:

- Allow the user to specify the expected curve/direction
- Set the curve/direction based on the metric choice 
- Try multiple locators with different curve characteristics if we can't find an elbow 
- Normalize the curve to always be convex/decreasing no matter the metric
- Some combination of one or more of the above

Potentially the second bullet point is the easiest option, and we can go that route and open an issue to explore this further. What do you think? Are there any other options?

yellowbrick/cluster/elbow.py
```diff
@@ -271,8 +282,10 @@ def draw(self):
         Draw the elbow curve for the specified scores and values of K.
         """
         # Plot the silhouette score against k
-        self.ax.plot(self.k_values_, self.k_scores_, marker="D", label="score")
-
+        self.ax.plot(self.k_values_, self.k_scores_, marker="D")
+        if self.knee:
+            self.ax.axvline(self.knee_value,c='black',linestyle='--')
```
bbengfort: For the color, would you mind using [`yellowbrick.style. palettes.LINE_COLOR`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/style/palettes.py#L45)?

yellowbrick/cluster/elbow.py
```diff
@@ -271,8 +282,10 @@ def draw(self):
...
+            self.ax.legend(['Score={}'.format(round(self.score,3)),'Optimal k={}'.format(self.knee_value)],loc='best')
```
bbengfort: To create a text-only legend, you could use [`manual_legend`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/draw.py#L29). However, I don't think that's necessary, instead you could do:

```python
elbow_label = "$elbow at k={}, score={0.2f}$".format(self.elbow_value_, self.elbow_score_)"
self.ax.axvline(self.elbow_value_, c=LINE_COLOR, linestyle="--", label=elbow_label)
```

Then later, in `finalize()` you can simply call `self.ax.legend(loc='best')` (which is the best place to put the legend drawing, or the manual legend drawing if you decide to go that route).

bbengfort: Note also the `$` in the label - that renders the label as latexmath, which might be nicer, it might not be though.

yellowbrick/cluster/elbow.py
```diff
@@ -281,6 +294,7 @@ def draw(self):
                 c='g', marker="o", linestyle="--", alpha=0.75,
             )
 
+
         return self.ax
 
     def finalize(self):
```
bbengfort: Recommend that you move the legend drawing to finalize!

yellowbrick/utils/kneed.py
```diff
@@ -0,0 +1,208 @@
...
+# NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN
+# IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+# ID: kneed.py [] pswaldia@no-reply.github.com $
+
+"""
+This package contains a port of the knee-point detection package, kneed, by 
+Kevin Arvai and hosted at https://github.com/arvkevi/kneed. This port is maintained 
+with permission by the Yellowbrick contributors.
+"""
```
bbengfort: @pswaldia thank you so much for including everything I asked for and making sure that the text was correctly broken at the right spacing! 

@arvkevi do you have any thoughts/comments on this? We want to make sure you're satisfied before we include this port!

yellowbrick/utils/kneed.py
```diff
@@ -0,0 +1,208 @@
...
+        :type x: list or array.
+        :param y: y values.
+        :type y: list or array.
+        :param S: Sensitivity, original paper suggests default of 1.0
+        :type S: float
+        :param curve: If 'concave', algorithm will detect knees. If 'convex', it
+            will detect elbows.
+        :type curve: string
+        :param direction: one of {"increasing", "decreasing"}
+        :type direction: string
```
bbengfort: Would it be possible to change the docstring to our docstring format to make sure it's documented correctly? If this is too much effort - no worries, I can always do it. Basically, this would involve:

1. Moving the docstring from under `__init__` to under the `class KneeLocator(object):` definition. 
2. Adding a description to the class 
3. Changing the parameters to the numpydoc style 

Also, we should probably add this to the documentation by including it as an RST file under the utils API!

yellowbrick/utils/kneed.py
```diff
@@ -0,0 +1,208 @@
...
+        value, and the x value where the knee is located.
+        :returns: tuple(knee, norm_knee, knee_x)
+        :rtype: (float, float, int)
+        )
+        """
+        if not self.xmx_idx.size:
+            warnings.warn("No local maxima found in the distance curve\n"
+                          "The line is probably not polynomial, try plotting\n"
+                          "the distance curve with plt.plot(knee.xd, knee.yd)\n"
+                          "Also check that you aren't mistakenly setting the curve argument", RuntimeWarning)
```
bbengfort: Unfortunately, I think we may have to do something about this warning, otherwise, it could be very confusing to our users. 

First, it is probably preferable to raise an exception here - that way we can catch the exception in the `KElbowVisualizer` and either keep drawing without the detected elbow, issue an elbow-specific warning, or advise the user to set `locate_elbow=False`. I think this warning will be issued given the convex/decreasing or concave/increasing issue that I mentioned above. So we may want to do things differently depending on if the user has the ability to control these params or something has gone wrong with what we expected for the metric. 

If we do go with a warning, instead of a RuntimeWarning, could we please issue a [`YellowbrickWarning `](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/exceptions.py#L99)?

yellowbrick/utils/kneed.py
```diff
@@ -0,0 +1,208 @@
...
+            for j in between_local_mx:
+                if j in self.xmn_idx:
+                    # reached a minima, x indices are unique
+                    # only need to check if j is a min
+                    if self.yd[j + 1] > self.yd[j]:
+                        self.Tmx[mxmx_i] = 0
+                        knee_x = None  # reset x where yd crossed Tmx
+                    elif self.yd[j + 1] <= self.yd[j]:
+                        warnings.warn("If this is a minima, "
+                                      "how would you ever get here:", RuntimeWarning)
```
bbengfort: Perhaps we should raise an exception here instead? If not, replacing with the `YellowbrickWarning` would also be appreciated (and a bit more detail about what's happening that's causing the warning).

yellowbrick/utils/kneed.py
```diff
@@ -0,0 +1,208 @@
...
+        # local minima
+        self.xmn_idx = argrelextrema(self.yd, np.less)[0]
+        self.xmn = self.xd[self.xmn_idx]
+        self.ymn = self.yd[self.xmn_idx]
+
+        # Step 5: Calculate thresholds
+        self.Tmx = self.__threshold(self.ymx)
+
+        # Step 6: find knee
+        self.knee, self.norm_knee, self.knee_x = self.find_knee()
```
bbengfort: Note that `find_knee()` is called in `__init__` as mentioned above, therefore there is no need to call it again in the visualizer. It may be interesting to refactor this into the `fit()`, `predict()` API to make it like an estimator; perhaps we could add that as an issue for a later revision.

yellowbrick/utils/kneed.py
```diff
@@ -0,0 +1,208 @@
...
+        import matplotlib.pyplot as plt
+
+        plt.figure(figsize=(8, 8))
+        plt.plot(self.x, self.y)
+        plt.vlines(self.knee, plt.ylim()[0], plt.ylim()[1])
+
+    # Niceties for users working with elbows rather than knees
+
+    @property
+    def elbow(self):
```
bbengfort: Ha! Excellent, we could use elbow instead of knee if you wanted! Although, I'm amused every time I see `self.eblow_value_ = knee_locator.knee`.

yellowbrick/utils/kneed.py
```diff
@@ -0,0 +1,208 @@
...
+
+        plt.vlines(self.norm_knee, plt.ylim()[0], plt.ylim()[1])
+
+    def plot_knee(self, ):
+        """Plot the curve and the knee, if it exists"""
+        import matplotlib.pyplot as plt
+
+        plt.figure(figsize=(8, 8))
+        plt.plot(self.x, self.y)
+        plt.vlines(self.knee, plt.ylim()[0], plt.ylim()[1])
```
bbengfort: These plot functions are probably not necessary, but let's leave them in. Additionally, I'd like to make an issue to refactor this so that these functions take an `ax` argument and plot directly to it, rather than importing `plt` and creating a new figure.

yellowbrick/utils/kneed.py
arvkevi: @pswaldia great work porting the code over.
Thanks for keeping me in the loop @bbengfort this looking great, I'm OK with all of these suggestions.
I'm just glad people are actually reviewing the source for accuracy and styling 😉

yellowbrick/utils/kneed.py
arvkevi: whew, this is very, very legacy... the warning started as a sanity check long ago.

bbengfort: We are running into it because we can be a bit all over the place with convex/increasing or concave/decreasing depending on the metric we're using -- and if the clustering is terrible (e.g. bad features, wrong algorithm, no actual clusters) then things get really wild at that point. Any advice or thoughts you have would be welcome!

yellowbrick/cluster/elbow.py
```diff
@@ -170,6 +171,9 @@ class KElbowVisualizer(ClusteringScoreVisualizer):
         Display the fitting time per k to evaluate the amount of time required
         to train the clustering model.
 
+    knee : bool, default=True
+        Display the vertical line corresponding to the optimal value of k.
```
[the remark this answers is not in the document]
pswaldia: Yes , this is definitely a great parameter name. I'll replace that with the suggested one. 
Thanks.

yellowbrick/cluster/elbow.py
```diff
@@ -219,6 +223,7 @@ def __init__(self, model, ax=None, k=10,
         # Store the arguments
         self.scoring_metric = KELBOW_SCOREMAP[metric]
         self.timings = timings
+        self.knee=knee
```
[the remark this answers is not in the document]
pswaldia: Yeah! Sure.

yellowbrick/cluster/elbow.py
pswaldia: @bbengfort Thanks so much for your kind words.

1. I was not aware of that naming convention of 'learned' attributes in yb but I found that in every visualizer class. I would change them as sugggested by you. Thanks.
2. I will document those attributes. They are important for other contributors too. Thanks.

And as suggested it makes sense for these properties to exist iff `self.locate_elbow==True`. I will make sure that happens.

yellowbrick/cluster/elbow.py
pswaldia: @bbengfort I was also concerned about the same , it didn't feel right to me to pass these two parameters manually. Regarding first bullet point , I think it would not be convenient for the users (new to ML) to specify the expected curve and direction. I think for now 2nd point is the most viable option given we have only three metrics for now. And if new metrics are to be included in future we can modify the same. I'll update this soon. Thanks.

yellowbrick/cluster/elbow.py
pswaldia: I'll use that.

yellowbrick/cluster/elbow.py
pswaldia: I'll definitely go deeper into this. And would do the changes as suggested. Thanks for great info.

yellowbrick/utils/kneed.py
pswaldia: @arvkevi thanks for your kind words.

yellowbrick/utils/kneed.py
pswaldia: I was about to do the same thing but thought would be great to get a feedback from you. I will move that docstring as directed. And will add to the documentation. Thanks.

yellowbrick/cluster/elbow.py
pswaldia: Oh! sorry, that's definetly a good pointer. I missed that. I'll update that too.

yellowbrick/cluster/elbow.py
pswaldia: @bbengfort Thanks for such in-depth review pointing out these small mistakes. Greatly appreciated. Yes you are right we can use `knee_locator.knee_x` instead of accessing the index of the knee value.

yellowbrick/utils/kneed.py
```diff
@@ -0,0 +1,208 @@
...
+"""
+import numpy as np
+from scipy import interpolate
+from scipy.signal import argrelextrema
+import warnings
+
+
+class KneeLocator(object):
+
+    def __init__(self, x, y, S=1.0, curve='concave', direction='increasing'):
```
pswaldia: In `__init__() ` curve parameter and direction parameter have default value `concave` and `increasing` which do not match with our choice of default metric `distortion` . So do you approve changing this also or its just fine?

yellowbrick/cluster/elbow.py
pswaldia: @bbengfort For the moment I don't think there's any need to include `knee locator` object in the visualizer. We can make it local to `fit()`. Also we have to name it ` elbow_locator `  right?

yellowbrick/utils/kneed.py
pswaldia: @bbengfort I am trying to get familiar with the exceptions handling. Once done I'll accommodate these changes in a PR. 
Thanks.

yellowbrick/utils/kneed.py
pswaldia: @bbengfort I am trying to become more and more familiar with yb codebase and the style of writing. It will be very exciting to refactor this into fit(), predict() api. I would start work on that once this feature works.

yellowbrick/cluster/elbow.py
bbengfort: I agree, also after more consideration; the `KneeLocator` object holds all the data, which is something we don't want to do generally, so I think it's good that we're not storing it on the Visualizer. I'm fine with either `knee_locator` or `elbow_locator` -- your choice!

yellowbrick/cluster/elbow.py
bbengfort: I appreciate it! The learned attributes are more a scikit-learn thing than a Yellowbrick thing. If you're interested in learning more, check out the [sklearn developer guide](https://scikit-learn.org/stable/developers/contributing.html#apis-of-scikit-learn-objects). I guess they call them "estimated attributes" there. 

A learned/estimated attribute is any data that is created when `fit()` is called - e.g. the `coef_` of a linear model. They only exist because data has been passed into the estimator. Because of this, the convention is to omit them from the class and only set them in `fit()` -- this also is what we use to check if an estimator is fitted or not. 

Because you're making the `KneeLocator` a local variable, not storing it on the visualizer, it is not a learned attribute. However, `elbow_value_`, `elbow_score_`, `k_scores_`, and `k_timers_` are all learned attributes because they do not/cannot exist until `fit()`.

yellowbrick/cluster/elbow.py
bbengfort: No need to apologize, we don't expect or require perfect code in a PR - that's what these reviews are for, so we can discuss how to make the code as excellent as possible.

yellowbrick/utils/kneed.py
bbengfort: Ok, great - but could we do that in a follow-up PR so that we have the original port in a commit?

yellowbrick/utils/kneed.py
bbengfort: Let's stick with the defaults that were defined in the original code; you mentioned that you might refactor this code later - we can discuss what we want to do with the defaults then!

pswaldia:
Hello @bbengfort I have made changes suggested by you. 
- In `elbow.py`

1. Learned attributes are documented.
2. Parameters/Attributes names are changed to make them more understandanble.
3. Set the `curve_nature` and `curve direction`according to the metric we have. It was pretty straighforward.
4. Modified legend in mathLatEx style and sent the legend drawing to `finalize()`.
5. Warn the user and advise them to set `locate_elbow=False` when no `knee` point deteced.

- In `kneed.py`

1.  Changed the docstring to match with `yellowbrick` style.
2. Modified the warning message and used `YellowBrickWarning` instead of `RunTimeWarning`. Warning message for 2nd warning has not been changed , it will be made more meaningful after going through the entire implementation of `kneed`.

Demo:
1.
```python
from sklearn.cluster import KMeans
from yellowbrick.cluster import KElbowVisualizer

model = KMeans()

visualizer = KElbowVisualizer(model, k=(4,12))  #default knee=True, metric='distortion'

visualizer.fit(X)    # Fit the data to the visualizer
visualizer.poof()    # Draw/show/poof the data
```
dist

2. 
```python
from sklearn.cluster import KMeans
from yellowbrick.cluster import KElbowVisualizer

model = KMeans()

visualizer = KElbowVisualizer(model,metric="calinski_harabaz", k=(4,12))  #default knee=True, metric=''

visualizer.fit(X)    # Fit the data to the visualizer
visualizer.poof()    # Draw/show/poof the data
```
calinski

3. `yellowBrickWarning` will be displayed when `knee` point is not detected. Curve will be drawn without elbow point and warning will advice user to pass `locate_elbow=False` to remove warning.Note: **The below plot has a elbow point but just for demonstration purpose the `knee_` value has been set to None.**
```python
from sklearn.cluster import KMeans
from yellowbrick.cluster import KElbowVisualizer

model = KMeans()

visualizer = KElbowVisualizer(model,metric="calinski_harabaz", k=(4,12))  #default knee=True, metric=''

visualizer.fit(X)    # Fit the data to the visualizer
visualizer.poof()    # Draw/show/poof the data
```
warn

yellowbrick/cluster/elbow.py
```diff
@@ -260,7 +293,25 @@ def fit(self, X, y=None, **kwargs):
...
-            )
+            )    
+
+        if self.locate_elbow:
+            if self.metric == 'distortion':
+                self.curve_nature = 'convex'
+                self.curve_direction = 'decreasing'
+            elif self.metric=='silhouette' or self.metric=='calinski_harabaz':
+                self.curve_nature = 'concave'
+                self.curve_direction = 'increasing'
```
bbengfort: This is not a big deal and won't block merging of this code, but I wanted to make a recommendation because I'm hoping you will be a regular contributor to Yellowbrick, and this is both a code style thing and an user-facing API thing. 

I mentioned that learned attributes, suffixed with an underscore, are only set in `fit()` (nice work with those attrs, by the way!) In YB/sklearn terms, the other type of attributes are _hyperparameters_ - the properties that the user passes into `__init__` - unlike other python programming, sklearn and YB treat these specially. All other attributes of the class should be methods, e.g. `fit()` or they should be marked as private, e.g. prefixed with an `_`. 

Based on this, technically (again not a big deal) curve direction and nature should be `self._curve_direction`  and `self._curve_nature`.  

There is another little item here, though, and that's that the curve nature and direction should depend only on the metric. What happens if the user changes these values? (nothing in the case of your code) It could become confusing for the user. 

Therefore I would propose that we simply keep these as local variables when instantiating the `KneeLocator`. Simply removing the `self.` is fine - but I also wanted to show you how you might see this in other YB code:

```python
locator_kwargs = {
     'distortion': {'curve_nature': 'convex', 'curve_direction': 'decreasing'},
     'silhouette': {'curve_nature': 'concave', 'curve_direction': 'increasing'},
     'calinski_harabaz': {'curve_nature': 'concave', 'curve_direction': 'increasing'},
}.get(self.metric, {})

elbow_locator = KneeLoator(self.k_values_, self.k_scores, **locator_kwargs)
```

This is a jump table pattern and we often use it over if/elif so that it's easy to make modifications on a per-metric basis.

yellowbrick/cluster/elbow.py
```diff
@@ -260,7 +293,25 @@ def fit(self, X, y=None, **kwargs):
...
+            )    
+
+        if self.locate_elbow:
+            if self.metric == 'distortion':
+                self.curve_nature = 'convex'
+                self.curve_direction = 'decreasing'
+            elif self.metric=='silhouette' or self.metric=='calinski_harabaz':
+                self.curve_nature = 'concave'
+                self.curve_direction = 'increasing'    
+            self.elbow_locator = KneeLocator(self.k_values_,self.k_scores_,curve_nature=self.curve_nature,curve_direction=self.curve_direction)
```
bbengfort: I thought we agreed that this would be a local variable and not stored on the class? I still think that's the right way to go.

yellowbrick/cluster/elbow.py
```diff
@@ -170,10 +173,31 @@ class KElbowVisualizer(ClusteringScoreVisualizer):
...
+
+    k_timers_ : array of shape (n,) where n is no. of k values
+        The time taken to fit n KMeans model corresponding to each k value.
+
+    elbow_value_ : integer
+        The optimal value of k.
+
+    elbow_score_ : float
+        The silhouette score corresponding to the optimal value of k.         
+
```
bbengfort: Very nice, thank you!

yellowbrick/cluster/elbow.py
```diff
@@ -260,7 +293,25 @@ def fit(self, X, y=None, **kwargs):
...
+            elif self.metric=='silhouette' or self.metric=='calinski_harabaz':
+                self.curve_nature = 'concave'
+                self.curve_direction = 'increasing'    
+            self.elbow_locator = KneeLocator(self.k_values_,self.k_scores_,curve_nature=self.curve_nature,curve_direction=self.curve_direction)
+            self.elbow_value_ = self.elbow_locator.knee
+            if self.elbow_value_ == None:
+                warning_message=\
+                "No 'knee' or 'elbow' point detected, " \
+                "pass `locate_elbow=False` to remove the warning"   
+                warnings.warn(warning_message,YellowbrickWarning)
```
bbengfort: Very nice, thank you for changing this warning into a much more understandable signal to the user!

yellowbrick/cluster/elbow.py
```diff
@@ -271,8 +322,11 @@ def draw(self):
         Draw the elbow curve for the specified scores and values of K.
         """
         # Plot the silhouette score against k
-        self.ax.plot(self.k_values_, self.k_scores_, marker="D", label="score")
-
+        self.ax.plot(self.k_values_, self.k_scores_, marker="D")
+        if self.locate_elbow and self.elbow_value_!=None:
+            elbow_label = "$elbow\ at\ k={}, score={}$".format(self.elbow_value_, np.round(self.elbow_score_,3))
```
bbengfort: Again, not a big deal in this case but my preference is `"{:0.3f}".format(self.elbow_score_)` rather than `np.round` - I think this might be more expected by other developers. It won't be a blocker for this PR though.

yellowbrick/cluster/elbow.py
```diff
@@ -299,6 +355,10 @@ def finalize(self):
         # Set the x and y labels
         self.ax.set_xlabel('k')
         self.ax.set_ylabel(metric.lower())
+        
+        #set the legend if locate_elbow=True
+        if self.locate_elbow and self.elbow_value_!=None:
+            self.ax.legend(loc='best', fontsize='medium')
```
bbengfort: Nice!

yellowbrick/utils/kneed.py
```diff
@@ -0,0 +1,242 @@
...
+
+    curve_direction : string, default: 'increasing'
+       A string that determines tha increasing or decreasing nature of the elbow curve in 
+       which "knee" or "elbow" is to be found.
+    
+    Notes
+    -----
+    The KneeLocator is implemented using the "knee point detection algorithm" which can be read at
+    `<https://www1.icsi.berkeley.edu/~barath/papers/kneedle-simplex11.pdf>`
+
```
bbengfort: Perfect!

yellowbrick/cluster/elbow.py
pswaldia: This jump table pattern is excellent. I always wondered what `**kwargs` used for , today I learnt how they can be used. It's lot better over the if/else , looks tidy too.

pswaldia:
@bbengfort  I am done with those minor yet important changes suggested. 
And yes that testing strategy seems good to me too. (Although I am not that experienced writing tests, it will be first one for me, but i am really excited.). I will also look into the documentation part. 
Thanks.

bbengfort:
Thanks for doing that @pswaldia - again, it was no big deal at all, but I do appreciate you making the changes; you're ensuring that this PR is absolutely top notch! 

A couple of tips for docs and tests follow:

**Documentation**

- Make sure you have installed the documentation requirements `pip install -r docs/requirements.txt` (preferably inside a virtual environment)
- To build the docs, change into the `docs` directory and run `make html`, this should create a folder called `_build/html` inside of the `docs` directory. Note this only works in the `docs` directory! Let me know if there are any errors.
- If you're on a Mac or a Linux machine you can `open _build/html/index.html` on the command line, otherwise, open a browser and go to File -> Open and select that file. 
- Navigate to Visualizers and API -> Clustering Visualizers ->  Elbow Method 
- To edit the docs, edit the reStructuredText file at `docs/api/cluster/elbow.rst`

Let me know if there are any errors building the documentation or if the docs look weird.

**Tests**

- Make sure you have the test requirements installed `pip install -r tests/requirements.txt` (preferably inside a virtual environment)
- Run the tests with `pytest` ... this will take a while. Note that the tests must be run from the root project directory.
- Run the elbow visualizer tests with `pytest tests/test_cluster/test_elbow.py` -- this is probably what you'll want to do most when testing. 
- You'll be writing/updating tests in `TestElbowVisualizer`, note that the method name must start with `test_` and that the method name and the single line docstring are used in the test reporting.

I suggest the following:

For any tests that are currently failing, update them with `locate_elbow=False`; if those tests pass, let's merge this PR in and get to the tests and the updates to the docs in another PR. If those tests don't pass, then we'll have to sort something out.

pswaldia:
@bbengfort Regarding the documentation it was a smooth sail but I did noticed some wierd things that did not match with the code written.
Screenshot 2019-04-20 at 02 24 36

What I expected was `locate_elbow : bool, default: True` but it was not so. This was same with every visualizer not only KElbowVisualizer that we updated. Is it okay to have this bcoz in the official documentation available on net its fine. ?

I changed the documentation a bit so please have a look at it and suggest changes if any. I will be completing the tests today

Thanks..

saurabhdaalia:
> @bbengfort Regarding the documentation it was a smooth sail but I did noticed some wierd things that did not match with the code written.
> Screenshot 2019-04-20 at 02 24 36
> 
> What I expected was `locate_elbow : bool, default: True` but it was not so. This was same with every visualizer not only KElbowVisualizer that we updated. Is it okay to have this bcoz in the official documentation available on net its fine. ?
> 
> I changed the documentation a bit so please have a look at it and suggest changes if any. I will be completing the tests today
> 
> Thanks..

@pswaldia , I am facing the same issue that you mentioned here.  This seems to be an environment issue I feel. Deactivating virtualenv and then running make command works fine.

pswaldia:
Thanks @bbengfort  I will be working on writing tests. Though it will not be that fast as I will be busy with my exams. But I will open a pull request doing the required. Thanks again. It was really nice having this pull request merged.

arvkevi:
Nice work @pswaldia 😃

pswaldia:
Thanks sir @arvkevi for your kind words.

bbengfort:
@pswaldia it is important for you to be able to focus on your exams, therefore I think there are two options here:

1. If you create a PR - I can fill in the test stubs to make sure we don't lose the thread while you're taking your exam. 
2. We can set the default `locate_elbow=False` 

I'm fine with either, but option 1  requires a bit more git wrangling for you (e.g. pull from upstream develop, push into your fork, delete your old feature branch, and create a new feature branch). Let me know what you'd prefer!

pswaldia:
@bbengfort we need to complete the tests sooner or the later , let's go with option 1.  I will open a new PR soon.

arvkevi:
@bbengfort there was a [bug](https://github.com/arvkevi/kneed/issues/31) in `knee_locator.py`.
The algorithm was reported the last (most recent) knee point detected instead of the first. This caused the sensitivity parameter `S` to have an insignificant effect. 

There are *no* changes to the API. The refactored code now reports the first knee and `S` now has a significant impact on the knee point detected. See [code snippet and plot in the README](https://github.com/arvkevi/kneed#sensitivity-parameter-s). 
Also included in the 0.4+ release:
1. removed an additional `for` loop 
2. improved variable naming
3. `KneeLocator` saves `all_knees`  
4. more tests...

I'm happy to open a new issue and/or PR and modify the yb source code myself. I see you are nearing a v1.0 release 🎉 so you may want to forge ahead and put this on the backburner. Either way, I wanted to bring it to your attention.

Kautumn06:
Hi @arvkevi &mdash; thank you so much for reaching out! Your timing is perfect because yesterday while reviewing PR #891 I noticed that number being returned for `k` appeared to be one less than the optimal value. I tested the `KElbowVisualizer` again using `make_blobs ` to create a synthetic dataset with 8 clusters and it also showed the optimal value as one less than what I was expecting:
make-blobs

I had planned on reaching out to you today to see if perhaps there was a bug in our port, so your timing couldn't be better. We greatly appreciate your help opening a PR to fix this; we are all pretty focused on our v1.0 release, but this is a great bug to squash before it goes out!

Kautumn06:
Hi @arvkevi &mdash; thank you so much for reaching out! Your timing is perfect because yesterday while reviewing PR #891 I noticed that number being returned for `k` appeared to be one less than the optimal value. I tested the `KElbowVisualizer` again using `make_blobs ` to create a synthetic dataset with 8 clusters and it also showed the optimal value as one less than what I was expecting:
make-blobs

I had planned on reaching out to you today to see if perhaps there was a bug in our port, so your timing couldn't be better. We greatly appreciate your help opening a PR to fix this; we are all pretty focused on our v1.0 release, but this is a great bug to squash before it goes out!

Kautumn06:
Hi @arvkevi &mdash; thank you so much for reaching out! Your timing is perfect because yesterday while reviewing PR #891 I noticed that number being returned for `k` appeared to be one less than the optimal value. I tested the `KElbowVisualizer` again using `make_blobs ` to create a synthetic dataset with 8 clusters and it also showed the optimal value as one less than what I was expecting:
make-blobs

I had planned on reaching out to you today to see if perhaps there was a bug in our port, so your timing couldn't be better. We greatly appreciate your help opening a PR to fix this; we are all pretty focused on our v1.0 release, but this is a great bug to squash before it goes out!

Kautumn06:
Hi @arvkevi &mdash; thank you so much for reaching out! Your timing is perfect because yesterday while reviewing PR #891 I noticed that number being returned for `k` appeared to be one less than the optimal value. I tested the `KElbowVisualizer` again using `make_blobs ` to create a synthetic dataset with 8 clusters and it also showed the optimal value as one less than what I was expecting:
make-blobs

I had planned on reaching out to you today to see if perhaps there was a bug in our port, so your timing couldn't be better. We greatly appreciate your help opening a PR to fix this; we are all pretty focused on our v1.0 release, but this is a great bug to squash before it goes out!

PR #837, opened 2019-05-06
Add color customization to SilhouetteVisualizer

mgarod:
### III. Summarize your PR (HINT: See CHECKLIST below!)
This PR addresses `Add ability to define cluster colors and improve color selection methodology.` which is a portion of #516

1. Made SilhouetteVisualizer use `resolve_color()` instead of `color_palette()`
1. Change SilhouetteVisualizer __init__ params to specify `colormap=` and `colors=` (also changed from `color` to `colors` to match `resolve_color()` args.
1. Add 2 new tests for colormap and colors changes
1. Add/change documentation for these args
1. Update changelog

### IV. Include a sample plot
`SilhouetteVisualizer(model, colormap='gnuplot')`
gnuplot
`SilhouetteVisualizer(model, colors=['red', 'green', 'blue', 'indigo', 'cyan', 'lavender'])`
rgbicl

- [ ]     The function `resolve_colors()` will throw warning if both `colormap` and `colors` are present. By default, `SilhouetteVisualizer` sets `colormap='set1'`. If a user sets `colors`, it therefore will throw the warning. This can be avoided by explicitly setting `colormap=None`. Should this be changed? 
- [ ]      Should we add references to matplotlib's colormap reference or named colors reference?
- [ ]      Should we add documentation for the use of hex RGB strings? (e.g. `'#00ff00'`)


mgarod:
The tests have failed because this PR is now using matplotlib's version of the `Set1` color map, and no longer yellowbrick's `set1` color map found in `styles/palettes.py`.

Need clarification if this is a matter of simply adjusting the tests to match the new logic, or if this is a breaking change that should be corrected.

pdamodaran:
@mgarod - thanks for taking this on!  I will be including a couple of comments that will help answer some of your questions regarding this PR.

yellowbrick/cluster/silhouette.py
```diff
@@ -97,13 +104,15 @@ class SilhouetteVisualizer(ClusteringScoreVisualizer):
     >>> model.poof()
     """
 
-    def __init__(self, model, ax=None, **kwargs):
+    def __init__(self, model, ax=None, colormap='set1', colors=None, **kwargs):
```
pdamodaran: Matplotlib allows for users to specify parameters such as colormap and colors through kwargs - therefore - the addition you included above will not be required.  We can then use the kwargs below to read in the user defined parameters and use defaults if not specified.

yellowbrick/cluster/silhouette.py
```diff
@@ -97,13 +104,15 @@ class SilhouetteVisualizer(ClusteringScoreVisualizer):
...
         super(SilhouetteVisualizer, self).__init__(model, ax=ax, **kwargs)
 
         # Visual Properties
-        # TODO: Fix the color handling
-        self.colormap = kwargs.get('colormap', 'set1')
-        self.color = kwargs.get('color', None)
+        # Use colors if it is given, otherwise attempt to find a colormap in
+        # styles.PALETTES. If not found, default to None. The colormap may
+        # yet still be found in resolve_colors
+        self.colormap = colormap
```
pdamodaran: change this to self.colormap = kwargs.get('colormap', 'set1')

What this does is sets the colormap to the user defined colormap, otherwise sets the colormap to set1 if the user does not provide a colormap.

yellowbrick/cluster/silhouette.py
```diff
@@ -97,13 +104,15 @@ class SilhouetteVisualizer(ClusteringScoreVisualizer):
...
+        self.colors = colors if colors else PALETTES.get(colormap, None)
```
pdamodaran: change this to self.colors = kwargs.get('colors', None)

Ideally the user shouldn't provide both colormap and colors but if they do they will get a warning which is the expected behavior.

The reason you were getting a warning when you tested the updated code is because colormap was set to 'set1' and colors parameter was getting set using the colors in colormap.

mgarod: Thanks for reviewing @pdamodaran 
The reason I had changed the way `self.colors` was set was because I needed a way to get `PALETTES.get(colormap, None)` in there somewhere.

If we reach a point where `self.colormap = 'set1'` and `'self.colors = None`, when we reach [resolve_colors()](https://github.com/DistrictDataLabs/yellowbrick/blob/d2634e5899e3048cd423f814bdd6d141f5397ca9/yellowbrick/style/colors.py#L92) it will attempt `cm.get_cmap('set1')` and it will fail because `'set1'` does not exist. There does exist `'Set1'` with identical colors (I think) but the colors are in different orders and will break all tests.

The func `resolve_colors()` needs a way to be able to pull colormaps from `PALETTES`.

pdamodaran:
@rebeccabilbro - your suggestions make sense to me.  @mgarod - basically the changes we are proposing is to use colors unless the user explicitly passes colormap through kwargs.  The reason behind this is that we are visualizing different clusters and so a distinct list of colors would be more suitable.  Because resolve_colors() takes care of the choice of colors if the user does not pass in anything there is no longer a need for us to set a default color value.  I will add some more comments under your changes.  Let us know if you have any more questions!

yellowbrick/cluster/silhouette.py
pdamodaran: based on @rebeccabilbro's suggestion change the above to:

def __init__(self, model, ax=None, colors=None, **kwargs):

mgarod:
Pushed new changes. In summary:

* `colors` is now the primary color setting option.
* `colormap` is still available, and will override `colors`
* Added test case where `colors='cool'` which is when the colors var (normally a list of strings) is instead a matplotlib colormap name.

Questions:
* Should the docstring for `colormap` be removed if it is not really the first choice for setting color anymore?

yellowbrick/cluster/silhouette.py
```diff
@@ -66,6 +66,15 @@ class SilhouetteVisualizer(ClusteringScoreVisualizer):
         The axes to plot the figure on. If None is passed in the current axes
         will be used (or generated if required).
 
+    colormap : str, default: None
+        The name of the matplotlib color map to use for each cluster group. Overrides colors if both are specified.
+
```
rebeccabilbro: I think we can safely remove `colormap` since that's no longer an attr option?

pdamodaran: @rebeccabilbro - that makes sense since we are using colors.
@mgarod - I was reviewing your changes and they looked good.  I was working on fixing the merge conflicts - so I will make this change as well and push out the changes and merge it in.

docs/changelog.rst
```diff
@@ -38,7 +38,7 @@ Minor Changes:
     - Updated the example notebooks and created a quickstart notebook.
     - Fixed broken links in the documentation.
     - Add legend to SilhouetteVisualizer for the average score axis
-    - Move SilhouetteVisualizer cluster identity labels from the middle to the y-axis.
```
rebeccabilbro: It looks like this line somehow disappeared @pdamodaran &mdash; can you double check?

pdamodaran: The line was indeed missing in the latest commit - I will add it in.

PR #839, opened 2019-05-07
Add legend for silhouette score

mgarod:
### III. Summarize your PR (HINT: See CHECKLIST below!)
This PR addresses `Add a legend/annotation that describes the average clustering coefficient (e.g. label the red axvline)` which is a portion of #516

1. Add label to red Silhouette Score vertical axis
1. Add legend in `finalize()`
1. Update changelog

### IV. Include a sample plot
legend

From `tests/test_cluster/test_silhouette.py` (NOTE: Tests specifically exclude legend as intended)
test_integrated_kmeans_silhouette

- [ ]     The legend uses `best` placement to draw onto the plot. This is potentially slow, and the term "best" is subjective (see images above).

> `The 'best' option can be quite slow for plots with large amounts of data. Your plotting speed may benefit from providing a specific location.` https://matplotlib.org/api/_as_gen/matplotlib.pyplot.legend.html


pdamodaran:
@mgarod - thanks so much for contributing at the PyCon sprints!  Your changes produce the intended effect on the visualizer.  @rebeccabilbro - would you also mind taking a look?  Also, regarding the question posed by @mgarod regarding loc="best" for the position of the legend, what are your thoughts regarding allowing the user to pass in the location using kwargs, but have "best" be the default if it wasn't specified?

rebeccabilbro:
>@rebeccabilbro - would you also mind taking a look? Also, regarding the question posed by @mgarod regarding loc="best" for the position of the legend, what are your thoughts regarding allowing the user to pass in the location using kwargs, but have "best" be the default if it wasn't specified?

Great question @mgarod @pdamodaran &mdash; indeed `loc='best'` **could** potentially get us into trouble with some of our visualizers that do store more data on the class, however in this case, it looks like we store only the number of samples (not the full vector for each sample) and the cluster count, so I think we should be safe, right?

pdamodaran:
@mgarod - your changes look good!  I will update your branch and should all the tests pass - I will merge your changes in.  Thanks for your work on this!

PR #844, opened 2019-05-07
- Update Example notebook to use develop branch download API

yanigisawa:
1. Update the examples.ipynb notebook to use `from yellowbrick.datasets import load_credit, load_concrete, load_occupancy, load_mushroom, load_occupancy` instead of the legacy load_data method
2. Inverted the boolean to set replace=True if the user wants to overwrite the previously downloaded files.

The following warnings are generated after I ran the `develop` version of this code. Do we want to suppress warnings, or try and resolve these prior to merging?
- `'c' argument looks like a single numeric RGB or RGBA sequence, which should be avoided as value-mapping will have precedence in case its length matches with 'x' & 'y'.  Please use a 2-D array with a single row if you really want to specify the same RGB or RGBA value for all points.`
- `/Users/jalexander/.virtualenvs/yellowbrick-qVNOLk1S/lib/python3.7/site-packages/scipy/stats/morestats.py:1653: UserWarning: p-value may not be accurate for N > 5000.
  warnings.warn("p-value may not be accurate for N > 5000.")`   
- `/Users/jalexander/.virtualenvs/yellowbrick-qVNOLk1S/lib/python3.7/site-packages/sklearn/preprocessing/data.py:645: DataConversionWarning: Data with input dtype int64 were all converted to float64 by StandardScaler.
  return self.partial_fit(X, y)`
- `/Users/jalexander/.virtualenvs/yellowbrick-qVNOLk1S/lib/python3.7/site-packages/sklearn/model_selection/_split.py:2053: FutureWarning: You should specify a value for 'cv' instead of relying on the default value. The default value will change from 3 to 5 in version 0.22.
  warnings.warn(CV_WARNING, FutureWarning) `


Kautumn06:
Hi @yanigisawa, thank you for submitting a pull request! Once I've finished looking over your changes, I'll let you know if I have any suggestions or questions. Thanks again!

Kautumn06:
Hey there @yanigisawa — I'm going to go ahead and close this PR out per the discussion above. However, if you ever decide you want to jump back in, please feel free to open a new PR from a [feature branch](https://www.scikit-yb.org/en/develop/contributing/getting_started.html#branching-convention). Thank you for checking out Yellowbrick and I hope you'll try to contribute again in the future!

Issue #771, opened 2019-02-22
Improvements to PosTagVisualizer


rebeccabilbro:
In #768 we rebooted the part-of-speech tag visualizer to make it more practical for NLP machine learning tasks. Here are a few ideas for follow-up features that could be added:

- [x] optional ordering of the plot by tag count (similar to a frequency distribution). _Note: this should NOT be the default behavior since it could prevent users from being able to compare across plots or across corpora_
- [x] optional stacked barchart for each category/class if `y` is supplied
- [ ] option for NER tags with entity recognition (or maybe this should be another visualizer with a shared base class?)
- [x] optional parsing of raw text (only if nltk or spacy is installed)

ajaysub110:
Hello,
Just a query. For the stacked barchart, do you want the number of elements in each class, for each part of speech?
Thanks
Ajay

lwgray:
Hi @ajaysub110 Thanks for reaching out.  We're a bit busy at the moment & will get back to you soon.

dubesar:
Hello,
Can I know what does parsing the raw text here means? 
Does this is the same as processing the raw text? Like using tokenizer for processing the raw text?

rebeccabilbro:
Hi @dubesar and thanks for checking out Yellowbrick! 

The `PosTagVisualizer` currently expects corpora whose documents have already been transformed from their raw state into a (document) list of (paragraphs) lists of (sentence) lists of (token, tag) tuples; for example:

```python
tagged_stanzas = [                                               # document-level
    [                                                            # paragraph-level
        [                                                        # sentence-level
            ('Whose', 'JJ'),('woods', 'NNS'),('these', 'DT'),    # token-tag-level
            ('are', 'VBP'),('I', 'PRP'),('think', 'VBP'),('I', 'PRP'),
            ('know', 'VBP'),('.', '.')
            ],
        [
            ('His', 'PRP$'),('house', 'NN'),('is', 'VBZ'),('in', 'IN'),
            ('the', 'DT'),('village', 'NN'),('though', 'IN'),(';', ':'),
            ('He', 'PRP'),('will', 'MD'),('not', 'RB'),('see', 'VB'),
            ('me', 'PRP'),('stopping', 'VBG'), ('here', 'RB'),('To', 'TO'),
            ('watch', 'VB'),('his', 'PRP$'),('woods', 'NNS'),('fill', 'VB'),
[...]

Issue #525, opened 2018-07-25
Stacked bar plots helper function


zjpoh:
Create a helper function in the `yellowbrick.draw` module that can create a stacked bar chart from a 2D array on the specified axes. This function should have the following basic signature:

```python
def bar_stack(data, ax=None, labels=None, ticks=None, colors=None, **kwargs):
    """
    An advanced bar chart plotting utility that can draw bar and stacked bar charts from
    data, wrapping calls to the specified matplotlib.Axes object. 

    Parameters
    ----------
    data : 2D array-like 
        The data associated with the bar chart where the columns represent each bar
        and the rows represent each stack in the bar chart. A single bar chart would 
        be a 2D array with only one row, a bar chart with three stacks per bar would 
        have a shape of (3, b).  

    ax : matplotlib.Axes, default: None
        The axes object to draw the barplot on, uses plt.gca() if not specified.

    labels : list of str, default: None
        The labels for each row in the bar stack, used to create a legend. 

    ticks : list of str, default: None
        The labels for each bar, added to the x-axis for a vertical plot, or the y-axis
        for a horizontal plot.  

    colors : array-like, default: None
        Specify the colors of each bar, each row in the stack, or every segment. 

    kwargs : dict 
        Additional keyword arguments to pass to ``ax.bar``. 
    """
```

This is just a base signature and should be further refined as we dig into the work. Once this helper function is created, the following visualizers should be updated to use it:

- [ ] `PosTagVisualizer`
- [ ] `FeatureImportances`
- [ ] `ClassPredictionError`

(@DistrictDataLabs/team-oz-maintainers please feel free to add other visualizers above as needed)

**Describe the solution you'd like**
A helper function to create stacked bar plot. (See #510 and #847, #771 )

**Is your feature request related to a problem? Please describe.**
This can be useful for `ClassPredictionError`, `FeatureImportances`, and future stacked bar plot.

lwgray:
@zjpoh thank you for submitting this feature request.

PR #847, opened 2019-05-14
Addressing issue #771: Adding stack barchart to postag

naresh-bachwani:
The following changes were made to the postag visualizer    
1.      Added optional stack feature which plots stack barchart, given y is provided. 
2.      Also deals with #856.
The code to obtain below plot:

```python
corpus = load_corpus('data/hobbies')
docs = corpus.data
labels = corpus.target
tagged_stanzas = [nltk.pos_tag(nltk.word_tokenize(sent)) for sent in docs]
tag = [tagged_stanzas]

viz = PosTagVisualizer(stack=True,colormap='gnuplot')
viz.fit(tag,labels)
viz.poof()
```
The stacked barchart
Screenshot (145)

### TODOs
1.      Add test for stack barplot
2.      Arrange the stack plot by feature.


yellowbrick/text/postag.py
```diff
@@ -139,16 +144,34 @@ def fit(self, X, y=None, **kwargs):
...
         """
-        # TODO: add support for other tagsets?
-        if self.tagset == "penn_treebank":
-            self.pos_tag_counts_ = self._penn_tag_map()
-            self._handle_treebank(X)
-
-        elif self.tagset == "universal":
-            self.pos_tag_counts_ = self._uni_tag_map()
-            self._handle_universal(X)
+        if(self.stack == True and y == None):
```
bbengfort: Normally we wouldn't be so nitpicky, but we're looking forward to working with you on more PRs and so I wanted to mention some [style guide](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) recommendations that we use in Yellowbrick to make our code perform better and protect it from future changes. In this case, we would probably write:

```python
if self.stack and y is None:
```

Because `self.stack` is a bool and the `is` comparator makes sense a bit more sense for `None`.  Note that in some cases we may also do `if self.stack and y` - which only checks for the truthiness of y, particularly failing if `y` is `[]` -- however, since it is not necessarily appropriate to do a `len(X) == len(y)` check here (there are other tools that make sure `X` and `y` are correctly aligned), the `y is None` check is probably best.

yellowbrick/text/postag.py
```diff
@@ -73,6 +74,8 @@ class PosTagVisualizer(TextVisualizer):
     frequency: bool {True, False}, default: False
         If set to True, part-of-speech tags will be plotted according to frequency,
         from most to least frequent.
+    stack : bool {True, False}, default : False
+        If set to True, stacked barplot will be plotted.
```
bbengfort: We might want to mention that `y` is required and what the stack will be - e.g. "Plot the PosTag frequency chart as a per-class stacked bar chart. Note that fit() requires `y` for this visualization." or something like that.

yellowbrick/text/postag.py
```diff
@@ -139,16 +144,34 @@ def fit(self, X, y=None, **kwargs):
...
+                labeled.append((X[0][i],y[i]))
+        self.label_ = set(map(lambda x:x[1], labeled))
+        
+        if(self.stack == True):
+            X = [[[y[0]] for y in labeled if y[1]==x] for x in self.label_]
+        else:
+            X = [X]
+        
+        self.last = None
+        self.stack_count = 0
```
bbengfort: Lines 150-162 are an interesting solution to the problem of how to count part of speech tags by class. I especially like the novel idea of drawing each bar one at a time, growing the bar size on each pass because the `self._handle` methods are cumulative. 

Unfortunately, this does bring up an issue - and that is that there is now a slight mismatch between the visual stacks and the data that the user might introspect, namely `self.pos_tag_counts_` -- which only represents the total chart and not the different stacks. One thing that we want to be conscious of is that the computations we make for the visualization might be relevant to the user for diagnostic purposes. 

I propose that if the user wanted a break down of the part of speech distribution by class (e.g. by passing `stack=True`) then they would also want to be able to access that data in `self.pos_tag_counts_`. E.g. these two cases:

```python
# stack is False 
self.pos_tag_counts_ == {
    "NN": 42, 
    "NNP": 12,
    "VBG: 27, 
    ...
}

# stack is True 
self.pos_tag_counts_ == {
     "books": {  
        "NN": 42, 
        "NNP": 12,
        "VBG: 27, 
        ...
    },
    "cooking": {  
        "NN": 112, 
        "NNP": 72,
        "VBG: 249, 
        ...
    },
}
```

The amount of work you're doing in `fit()` could be minimized by computing the unique labels in `y` using `np.unique` (otherwise `None`), which `self._*_tag_map()` would take advantage of if present. The `self._handle_*` methods could also take `y` as input, and they could construct the correct learned output as above. Then draw would only have to know about `self.pos_tag_counts_` as before without having to track the stack count or the last stack marker. 

Although a relatively small change, I believe this would have a very big impact on the usability of the visualizer. Would you be willing to try to make the changes I suggested above?

yellowbrick/text/postag.py
```diff
@@ -359,6 +418,13 @@ def finalize(self, **kwargs):
                 "{} part-of-speech tags".format(self.tagset_names[self.tagset])
             )
         self.ax.set_ylabel("Count")
+        if(self.stack == True):
+            colors = resolve_colors(
+                    n_colors=len(self.label_), 
+                    colormap=self.colormap,
+                    colors=self.colors
+                    )
+            manual_legend(self, self.label_, colors, loc='best')
```
bbengfort: Great use of the `manual_legend`!

yellowbrick/text/postag.py
```diff
@@ -417,11 +485,11 @@ def postag(
     # Instantiate the visualizer
     visualizer = PosTagVisualizer(
         ax=ax, tagset=tagset, colors=colors, colormap=colormap,
-        frequency=frequency, **kwargs
+        frequency=frequency, stack=stack, **kwargs
     )
 
     # Fit and transform the visualizer (calls draw)
-    visualizer.fit(X, **kwargs)
+    visualizer.fit(X, y=y, **kwargs)
```
bbengfort: Thank you so much for updating the quick method as well!

yellowbrick/text/postag.py
```diff
@@ -139,16 +144,34 @@ def fit(self, X, y=None, **kwargs):
...
-            self.pos_tag_counts_ = self._uni_tag_map()
-            self._handle_universal(X)
+        if(self.stack == True and y == None):
+            raise YellowbrickValueError("Specify y for stack=True")
+        
+        labeled = []
+        if(y!=None):
+            for i in range(len(y)):
+                labeled.append((X[0][i],y[i]))
+        self.label_ = set(map(lambda x:x[1], labeled))
```
bbengfort: I think it's important that we keep track of the unique labels in `y` if they're passed in - but I would like to make this learned attribute consistent with the other classification visualizers. Would you mind calling this `self.labels_`?

yellowbrick/text/postag.py
naresh-bachwani: Hello @bbengfort, I will love to make these changes. I will get back to you by tomorrow with all the amendments.
My interpretation of the above is as follows. You suggest making a new function _*_tag_map() which returns a dictionary having the pos_tag_count for each label(if stack is True). And then use this returned dictionary in draw() to generate the stacked barplot. 
Is this what you mean? If not then please let me know about what needs to be done!

bbengfort: Hi Naresh, please take your time - there is no rush! 

And thank you for asking for clarification! I do not mean to create a new function, I'm simply using a shorthand to identify both `_uni_tag_map` and `_penn_tag_map` - same with the `_handle_treebank` and `_handle_universal` methods. 

The `_uni_tag_map` and `_pen_tag_map` should return a different datastructure depending on `self.stack` and `self.labels_`, which is saved on `self.pos_tag_counts_`. 

`_handle_treebank` and `_handle_universal` will have to accept an addtional, optional `y=None` argument, and will modify `self.pos_tag_counts_` accordingly. Then `draw()` will use `self.pos_tag_counts_`.

naresh-bachwani: Thanks!!

naresh-bachwani:
Hello @bbengfort, 
I made the amendments and added a new feature to obtain frequency sorted stacked barplots. I also added a few more tests for stacked barchart with and without frequency mode. 
I had a question regarding Travis test failing. I made no changes in **RFECV** and **Class Balance** yet those tests are failing. I am perplexed about the situation and can find no specific reasons for it.

rebeccabilbro:
> I made the amendments and added a new feature to obtain frequency sorted stacked barplots. I also added a few more tests for stacked barchart with and without frequency mode.
I had a question regarding Travis test failing. I made no changes in RFECV and Class Balance yet those tests are failing. I am perplexed about the situation and can find no specific reasons for it.

Hey there @naresh-bachwani &mdash;  not to worry! The errors are not related to your changes, but to a new update to Matplotlib that is currently impacting our automated tests. [We're on it and have a temporary fix in the works that should be merged in shortly](https://github.com/DistrictDataLabs/yellowbrick/pull/854), after which we'll pull from `develop` into your branch, which will trigger the tests to re-run! Stay tuned!

naresh-bachwani:
Thanks for helping:)

tests/test_text/test_postag.py
```diff
@@ -258,3 +258,37 @@ def test_spacy_tagged(self):
...
+        check_nltk_data()
+
+        _, ax = plt.subplots()
+        tagged_docs = list(get_tagged_docs(sonnets))
+
+        visualizer = PosTagVisualizer(stack=True, frequency=True, ax=ax)
+        visualizer.fit(tagged_docs, y=['a','b','c'])
+        visualizer.ax.grid(False)
+
+        self.assert_images_similar(ax=ax)
```
bbengfort: These tests look very good!

yellowbrick/text/postag.py
```diff
@@ -49,15 +51,12 @@ class PosTagVisualizer(TextVisualizer):
     only about a word’s definition, but also its use in context (for
     example the words “ship” and “shop” can be either a verb or a noun,
     depending on the context).
-
```
bbengfort: Can I ask why these white spaces were removed? They seem important to the readability of the docstring - do you have some setting on your code editor that is removing them automatically? If so, would you mind turning that setting off and replacing these blank lines?

yellowbrick/text/postag.py
```diff
@@ -114,251 +115,381 @@ def __init__(
...
-        # TODO: add support for other tagsets?
+        if self.stack and y is None:
+            raise YellowbrickValueError("Specify y for stack=True")
+        
+        self.labels_=np.unique(y)
+        
         if self.tagset == "penn_treebank":
             self.pos_tag_counts_ = self._penn_tag_map()
-            self._handle_treebank(X)
-
```
bbengfort: Same with these newlines as well!

yellowbrick/text/postag.py
```diff
@@ -114,251 +115,381 @@ def __init__(
...
+                            "non-English": 0,
+                            "interjection": 0,
+                            "list": 0,
+                            "symbol": 0,
+                            "other": 0,
+                        }
+        if self.stack:
+            penn_tag_map={}
+            for i in self.labels_:
+                penn_tag_map[i]=default_dic.copy()
```
bbengfort: Good use of `copy()` - that's very important! 

There is no need to make this change - but one thing I would think about looking at what's going on in `_penn_tag_map` and `_uni_tag_map` now is that there is some duplicate code that didn't exist before. If we wanted to add another tagset, we would have to duplicate this again. 

The only thing that is really different between these two functions is the names in the dictionary. Having the two functions before ensured that we were not passing around the same mutable object (e.g. the same reason you used `copy` here). 

One goal of YB is to make sure that the code is understandable to new contributors. One way to refactor the code might have been to add constants with the list of tags, then create a helper function that was called from the internal tag constructor functions. This would have made it easy to add new tags and also make the code a bit more readable to others:

```python
PENN_TAGS = [
    "noun", "verb", "adjective", "adverb", "preposition", "determiner", "pronoun",
    "conjunction", "infinitive", "wh- word", "modal", "possessive", "existential",
    "punctuation", "digit", "non-English", "interjection", "list", "symbol", "other", 
]

UNIVERSAL_TAGS = [...]

class PosTagVisualizer(TextVisualizer):

    def self._penn_tag_map(self):
        """
        Returns ...
        """
        return self._make_tag_map(PENN_TAGS)

    def self._uni_tag_map(self):
        """
        Returns ...
        """
        return self._make_tag_map(UNIVERSAL_TAGS)

    def _make_tag_map(self, tagset):
        """
        Returns a map of the tagset to a counter unless stack=True then returns
        a map of labels to a map of tagset to counters. 
        """
        zeros = [0] * len(tagset) # ensures the dict only contains a zero counter per tag
        if self.stack:
            return {
                label: dict(zip(tagset, zeros))
                for label in self.labels_ 
            }
        return dict(zip(tagset, zeros))
```

yellowbrick/text/postag.py
```diff
@@ -114,251 +115,381 @@ def __init__(
...
+                            if tag == "SPACE":
+                                continue
+                            self.pos_tag_counts_[y][jump.get(tag, "other")] += 1
+            else:
+                for tagged_doc in X:
+                    for tagged_sent in tagged_doc:
+                        for _, tag in tagged_sent:
+                            if tag == "SPACE":
+                                continue
+                            self.pos_tag_counts_[jump.get(tag, "other")] += 1
```
bbengfort: I really appreciate your efforts at ensuring the control flow logic works correctly! The issue here is that the control flow happens too high in the logic -- leading to a lot of repeated code. Consider the following control flow logic instead:

```python
for idx, tagged_doc in enumerate(X):
    for tagged_sent in tagged_doc:
        for _, tag in tagged_sent:

            if tag == "SPACE":
                continue

            # get counter correct counter from arguments 
            counter = self.pos_tag_counts_[y[idx]] if self.stack else self.pos_tag_counts_ 
            counter[jump.get(tag, "other")] += 1
```

Note that this results in the same number of loops as the original code and your modification and is much more readable at the cost of potentially tracking an unnecessary `idx` in the case of `self.stack is False`. This cost is more than worth it though, as it makes it easier to modify this code in the future and reduces the number of bugs.

yellowbrick/text/postag.py
```diff
@@ -114,251 +115,381 @@ def __init__(
...
+                    }
+        if self.stack:
+            uni_tag_map={}
+            for i in self.labels_:
+                uni_tag_map[i]=default_dic.copy()
+            return uni_tag_map
+        else:
+            return default_dic
+
+    def _handle_universal(self, X,y):
```
bbengfort: The signature here should be `def _handle_universal(self, X, y=None)` to match the signature of `fit()`. Also would you mind adding the `y` parameter to the docstring under X?

yellowbrick/text/postag.py
```diff
@@ -114,251 +115,381 @@ def __init__(
...
+                            self.pos_tag_counts_[y][jump.get(tag, "other")] += 1
+            else:
+                for tagged_doc in X:
+                    for tagged_sent in tagged_doc:
+                        for _, tag in tagged_sent:
+                            if tag == "SPACE":
+                                continue
+                            self.pos_tag_counts_[jump.get(tag, "other")] += 1
+
+    def _handle_treebank(self, X,y):
```
bbengfort: Same comment as above with the signature `(self, X, y=None)` and the docstring!

yellowbrick/text/postag.py
```diff
@@ -114,251 +115,381 @@ def __init__(
...
+                            self.pos_tag_counts_[y]["punctuation"] += 1
+                        elif tag == "TO":
+                            self.pos_tag_counts_[y]["infinitive"] += 1
+                        elif tag == "UH":
+                            self.pos_tag_counts_[y]["interjection"] += 1
+                        elif tag == "SYM":
+                            self.pos_tag_counts_[y]["symbol"] += 1
+                        else:
+                            self.pos_tag_counts_[y]["other"] += 1
+        else:
```
bbengfort: Would you mind also applying my comments about control flow logic from the other handle method here as well? As you can see, with a complex function like this one, where the choice really only is a huge switch, ensuring that the control flow happens as low as possible becomes increasingly important! 

As in the first function, you should be able to accomplish what you're looking for with only 2 additions and several modifications.

yellowbrick/text/postag.py
```diff
@@ -408,7 +540,6 @@ def postag(
         from most to least frequent.
     kwargs : dict
         Pass any additional keyword arguments to the PosTagVisualizer.
-
```
bbengfort: Since we're at the bottom of the code - just a gentle reminder to modify your editor settings so lines aren't removed by accident!

yellowbrick/text/postag.py
```diff
@@ -114,251 +115,381 @@ def __init__(
...
+            colors = resolve_colors(
+                    n_colors=len(self.labels_), 
+                    colormap=self.colormap,
+                    colors=self.colors
+                    )
+            manual_legend(self, self.labels_, colors, loc='best')
+            self.set_title(
+                "PosTag plot for {}-token corpus".format(
+                    (sum([sum(i.values()) for i in self.pos_tag_counts_.values()]))
+                )
```
bbengfort: As in `draw()` the control flow logic is at an appropriate level, but there is still a lot of duplicate code. Would you mind either refactoring to minimize the amount of duplicated code or to specialize the functionality for each case and comment which case is which?

yellowbrick/text/postag.py
```diff
@@ -114,251 +115,381 @@ def __init__(
...
             Axes on which the PosTagVisualizer was drawn.
         """
-        colors = resolve_colors(
-            n_colors=len(self.pos_tag_counts_),
-            colormap=self.colormap,
-            colors=self.colors,
-        )
-
-        if self.frequency:
+        if self.stack:
```
bbengfort: In this case, the control logic is probably at an appropriate place, but there is a lot of duplicated code here as well. 

There seem to be four cases to consider:

1. stack/frequency
2. stack/no-frequency
3. no-stack/frequency
4. no-stack/no-frequency 

What code do each of these cases share? What are their primary differences? Do you think you could take another look to see if you can specialize each case (and comment that they exist and which case is which) or to reduce the duplicate code? I'm happy to discuss it further if you'd like - I know I have a number of tricks up my sleeve for handling this and would be happy to chat about it!

yellowbrick/text/postag.py
naresh-bachwani: I think that can be a problem with the editor because I didn't change the docstring for fit(). I will find a setting and get it right!

yellowbrick/text/postag.py
naresh-bachwani: I will make it a point to avoid building duplicate code. I understand that it makes the code look more complex for the new developers.

yellowbrick/text/postag.py
```diff
@@ -150,70 +169,34 @@ def fit(self, X, y=None, **kwargs):
...
-                        "other": 0,
-                    }
-        if self.stack:
-            uni_tag_map={}
-            for i in self.labels_:
-                uni_tag_map[i]=default_dic.copy()
-            return uni_tag_map
-        else:
-            return default_dic
+        self.pos_tags = UNIVERSAL_TAGS
```
bbengfort: Is it necessary to store `pos_tags` on the visualizer? (e.g. are you using this property elsewhere?) This seems like a bit of a duplicate with `self.tagset` and the keys on the `pos_tag_counts_`.

yellowbrick/text/postag.py
```diff
@@ -242,132 +225,83 @@ def _handle_universal(self, X,y):
...
+                    elif tag in self.punct_tags:
+                        counter["punctuation"] += 1
+                    elif tag == "TO":
+                        counter["infinitive"] += 1
+                    elif tag == "UH":
+                        counter["interjection"] += 1
+                    elif tag == "SYM":
+                        counter["symbol"] += 1
+                    else:
+                        counter["other"] += 1
```
bbengfort: Good updates!

yellowbrick/text/postag.py
naresh-bachwani: Yup! It is duplicate but I used it at certain places to de-duplicate the code. Like in `finalize()` it leads to common xticks for both `frequency=True` and `frequency=False`. It is not used extensively and can be removed if you suggest!!

bbengfort: No worries - if you're using it to deduplicate code then that's fine with me! My preference, however, would be to make it `self._pos_tags` -- the sklearn API suggests the following for estimator attributes:

- no underscore affix (`self.foo`): a hyperparameter passed in by the user in `__init__`
- underscore suffix  (`self.foo_`): a learned parameter that only exists on the class after calling `fit()` 
- underscore prefix (`self._foo`): any other required class property that is not a method

yellowbrick/text/postag.py
bbengfort: Bump!

yellowbrick/text/postag.py
bbengfort: Bump!

yellowbrick/text/postag.py
```diff
@@ -114,252 +131,279 @@ def __init__(
...
         colors = resolve_colors(
-            n_colors=len(self.pos_tag_counts_),
-            colormap=self.colormap,
-            colors=self.colors,
-        )
-
+                    n_colors=len(self.pos_tag_counts_), 
+                    colormap=self.colormap,
+                    colors=self.colors
+                    )
```
bbengfort: It appears something went wrong with the indent here?

yellowbrick/text/postag.py
```diff
@@ -114,252 +131,279 @@ def __init__(
...
+        
+        pos_tag_sum = copy.deepcopy(self.pos_tag_counts_)
```
bbengfort: Why is this deep copy required? If it's necessary, I'm all for it - but we do want to be conscious of how much memory we're using when creating tools in a machine learning context -- people might be working with very large datasets.

yellowbrick/text/postag.py
```diff
@@ -114,252 +131,279 @@ def __init__(
...
+        try:
+            #Stores the tag-wise sum of each label
+            pos_tag_sum = sum((Counter(i) for i in self.pos_tag_counts_.values())
+                            , Counter())
+        except TypeError:
+            pass
```
bbengfort: This is interesting - I do like the EAFP (easier to ask forgiveness than permission) over the LYBL (look before you leap) style of more pythonic programming. I suppose the idea here is that this is only used for `stack=True` and an exception will be raised if the type is not correct in the `stack=False` case? 

Although this line of code will likely change when you create the stack helper function -- I would like to point out two things:

1. When doing something like this - it is important to add comments! We want to make sure that another contributor and future you (in 3 months time) can understand what's happening in this code! 
2. Line 323 represents a safety flaw - although doing nothing is what you're going for here, this try/except could capture a different type of error, burying it and making bugs hard to diagnose. If you did `pos_tag_sum=None` instead of `pass` - it would at least allow `None` type errors later in the program which would guide us to this line of code.

yellowbrick/text/postag.py
```diff
@@ -114,252 +131,279 @@ def __init__(
...
-
-        self.ax.bar(
-            range(len(self.pos_tag_counts_)),
-            list(self.pos_tag_counts_.values()),
-            color=colors,
-        )
+            self.pos_tags = sorted(
+                        pos_tag_sum, key=pos_tag_sum.get, 
+                        reverse=True
+                        )
```
bbengfort: Indent is off here?

bbengfort:
@naresh-bachwani spec for the stacked bar chart helper function is in #525

yellowbrick/text/postag.py
naresh-bachwani: Deep copy is used as pos_tag_sum may be a nested list. So basic idea was to take an extra measure in case I changed the inner values. But in this I am not using it so I will go on to save it as `copy()` and not `deepcopy()`

yellowbrick/text/postag.py
naresh-bachwani: Thanks for providing me with regulations!

naresh-bachwani:
@bbengfort  ready for review!

yellowbrick/text/postag.py
```diff
@@ -157,10 +155,11 @@ def fit(self, X, y=None, **kwargs):
         self : instance
             Returns the instance of the transformer/visualizer
         """
-        if self.stack and y is None:
-            raise YellowbrickValueError("Specify y for stack=True")
-        
-        self.labels_=np.unique(y)
+        self.labels_=['documents']
```
bbengfort: Ok, this makes sense to me - but we should make a note of it in the docstring when we describe `self.labels_` - nice rewrite for the exception handling section below!

yellowbrick/text/postag.py
```diff
@@ -232,10 +230,14 @@ def _handle_universal(self, X,y):
                     for _, tag in tagged_sent:
                         if tag == "SPACE":
                             continue
-                        counter = self.pos_tag_counts_[y[idx]] if self.stack else self.pos_tag_counts_
+                        if self.stack:
+                            counter = self.pos_tag_counts_[y[idx]] 
+                        else: 
+                            counter = self.pos_tag_counts_['documents']
```
bbengfort: no need to change, but consider the following as a stylistic thing you might like:

```python
label = y[idx] if self.stack else 'documents'
counter = self.pos_tag_counts[label]
```

yellowbrick/text/postag.py
```diff
@@ -232,10 +230,14 @@ def _handle_universal(self, X,y):
...
+                        
                         counter[jump.get(tag, "other")] += 1
 
-    def _handle_treebank(self, X,y):
+    def _handle_treebank(self, X, y=None):
```
bbengfort: Thank you for making these updates!

bbengfort: It's not very important here - but let's please be sure that we are also updating the docstrings with all the parameters. Since this is library code, all of our docstrings serve as documentation that users might see with `help()` or built in the docs.

yellowbrick/text/postag.py
```diff
@@ -308,59 +314,40 @@ def draw(self, **kwargs):
...
-        try:
-            #Stores the tag-wise sum of each label
-            pos_tag_sum = sum((Counter(i) for i in self.pos_tag_counts_.values())
-                            , Counter())
-        except TypeError:
-            pass
+        # Converts nested dict to nested list
+        pos_tag_counts = np.array([list(i.values()) for i in self.pos_tag_counts_.values()])
+        # stores sum of nested list column wise
+        pos_tag_sum = np.sum(pos_tag_counts, axis=0)
```
bbengfort: Awesome!

bbengfort:
@naresh-bachwani I know there were some minor documentation todos - but we can sort those out in the next PR (when you add the stacked bar plot helper) ... if you don't mind making a note of them. I've updated this branch with develop and will merge it in when the tests pass.

PR #859, opened 2019-05-24
Added utf8 encoding option to the open() function

pHequals7:
New to creating pull requests,
but noticed that your load_corpus() function did not have the encoding option of utf8. so added that :)

### DIFF
--- a/docs/api/text/corpus.py
+++ b/docs/api/text/corpus.py
@@ -31,7 +31,7 @@ def load_corpus(path):
             files.append(os.path.join(path, cat, name))
             target.append(cat)
 
-            with open(os.path.join(path, cat, name), 'r') as f:
+            with open(os.path.join(path, cat, name), 'r',encoding = 'utf8') as f:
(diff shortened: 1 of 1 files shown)


lwgray:
Hi @pHequals7 

Thank you so much for checking out Yellowbrick and opening this PR.  We really appreciate the contribution.  However, the load_corpus function you updated is being deprecated because of our impending release (v1.0) which  incorporates a brand new datasets module. You can learn more about it here: http://www.scikit-yb.org/en/develop/api/datasets.html “ (edited) 

If you are up for it and interested in doing so; you can open a new PR and just remove those files `docs/api/text/corpus.py`  and `docs/api/text/corpus.rst` (please double check that we aren't referencing either of those pages from any other page in the docs that might result in a broken link) that would be great.  If you aren't interested that is fine also, we can open up an issue for it to be fixed. 

Again we are so very grateful for you checking out Yellowbrick.  Have a wonderful day!

Issue #690, opened 2019-01-19
Some YB tests fail with conda Python 3.7 & Matplotlib 3.x


amueller:
I'm trying to run the tests.
I get 133 failures, many of them
```
E   AssertionError: 'Qt5Agg' != 'agg'
```

Does running the tests require setting a specific backend?

ndanielsen:
@amueller great question and apologies that the tests didn't pass on the first shot. 

The yb uses the [`agg` backend ](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/base.py#L72)

Sometimes it needs to be set in the specific matplotlib config usually located at  `/.matplotlib/matplotlibrc`

Let me know if this helps with getting the tests passing!

amueller:
I actually run "make test" which I thought was the sure-fire way to do it right ;)
Usually I would use pytest. The Makefile actually uses ``python setup.py test``.

Re config: according to ``matplotlib.get_configdir()``  it's in ``.config/matplotlib`` and it's empty.

Running ``pytest tests`` in the root YB directory yields the same errors.

amueller:
I'm not on the same machine right now, but have the same problem on a different machine.
Also has an empty matplotlibrc, is using ubuntu 18.04.1, conda with python 3.7.1, matplotlib 3.0.2

rebeccabilbro:
Ah, thank you very much for the additional details @amueller! I think I've figured out what the issue is.

Apparently, this is [a unique Anaconda bug](https://github.com/ContinuumIO/anaconda-issues/issues/10219#issuecomment-433712810) that occurs when using conda Python 3.7 and Matplotlib 3.x ([which updated the way that backends are set](https://matplotlib.org/users/whats_new.html#improved-default-backend-selection)). 

Fortunately, this bug should only affect YB tests for you, and for contributing (which I hope you're interested in!) if you don't mind using vanilla Python 3.x, you'll be able to avoid them entirely.  Alternatively, you can manually set the backend using the workaround described by @nehaljwani. 

In the meantime, I've reported the behavior you observed on that Anaconda issue, which we'll track closely. Our tests are currently passing on Travis with pure Python, and on AppVeyor with Windows and miniconda. As soon as the Anaconda issue is resolved, we can add miniconda to our Travis test matrix, too (assuming it’s available) so that we can keep an eye on it moving forward!

amueller:
[...]

PR #862, opened 2019-05-25
Handle mpl 3.1.0 default backend that is inconsistent with 'agg'

nickpowersys:
### I. Fix issue reported in #854 and backend configuration tests in #690

Based on looking at the backend test that has been in place, I found that it does not match the backend that is used by the visualizer. This is because the backend test has been checking the backend that would be used if use('agg') was not in place. However, use('agg') is having the intended effect of modifying the backend as needed.

I verified that the backend used within the instantiation of the image comparison visualizer is 'agg', by using assertions.

I have removed the existing test and created another test within test_meta.py.

Appveyor Miniconda will be addressed as needed in #823.

I am coordinating my progress with @ndanielsen


ndanielsen:
Thanks for putting this in. I'll take a closer look tomorrow or Monday. 😁

nickpowersys:
That is just a placeholder, allowing 3.1.0 back into the tests/requirements.txt. The full solution is more likely to be ready after the weekend.

nickpowersys:
**Background**: The change that happened in mpl 3.1.0 affected the MacOS default backend so that it became 'MacOSX' instead of 'agg', leading to the backend assertion failures.

From matplotlib.org:
> Previous versions of matplotlib required a Framework build of python to work. The app type was updated to no longer require this, so the MacOSX backend should work with non-framework python.

**Proposed fix**: As described in API changes to matplotlib at https://matplotlib.org/api/api_changes.html:

> Test functions should use pytest.mark.backend(...), and the mark will be picked up by the matplotlib.testing.conftest.mpl_test_settings fixture.

It would allow contributors to have the backend set automatically for local and CI tests, whereas `mpl.use('agg')` within `tests/__init__.py` does not result in the backend being switched to 'agg'. (See related discussion of Travis build at the end.)

**Using the matplotlib API-related suggestion to use `pytest.mark.backend` within the code seems like the cleanest, simplest and most reliable option for supporting contributors and users overall, without imposing an additional burden on MacOS contributors. I can implement and validate it cross-platform if this is the most desirable approach. It would require an update to tests/requirements.txt. A user requirements.txt update is not required.**

**Tests/requirements.txt: MPL Releases**
The use of `pytest.mark.backend` started with mpl 2.1.0, released Oct. 2017.
https://github.com/matplotlib/matplotlib/commit/05f4b16c3e4e28500d9e87192b1a2e1a68ee2553

The yb tests/requirements.txt has matplotlib >= 1.5.1.

**Alternative fixes would result in less compatibility between the contributor and regular user experience**
- Both 1) creating a matplotlibrc within the repositories and 2) `use('agg')` in  `yellowbrick/__init__.py` would affect non-testing plots as well as testing plots. (See discussion of Travis custom build and use('agg') below). matplotlibrc files could be created after installation, but this would result in an additional file to track and manage.
- If contributors relied on os.environ['MPLBACKEND']='agg' through the command line or a dedicated script for local testing and development, it would require additional configuration steps. 

The environment variable approach is still a possibility, although it would be less convenient and possibly problematic for contributors on Mac to incorporate into a local workflow. Any doc additions that are needed would apply only to the MacOS for now.

**Impact of 3.1 on backend selection appears limited to MacOS**
On **Appveyor** within the DistrictDataLabs builds, while there was an image comparison failure after mpl 3.1.0 was released and installed within the build, and the backend assertion tests passed. Therefore, 'agg' is being used.

On **Travis**, within my own custom build, I
- allowed mpl 3.1.0 in tests/requirements.txt
- changed `use('agg')` in `tests/__init__.py` to `use('TkAgg')` to see if this changed the backend

All backend tests passed for this regular Python, Linux build.

This shows that `use('agg')` within `tests/__init__.py` has not been affecting the backend. `'Agg'` is still the backend is due to its being the default backend selected during the first set of imports.

https://travis-ci.org/nickpowersys/yellowbrick/jobs/535614512

nickpowersys:
The test_meta.py tests validate that conditions needed for ImageComparison tests cases are met. ImageComparison objects and visualizers are instantiated in the tests. This seems like the best place for a new backend test. This single test will fail if the backend has not been set correctly.

 I will also remove legends from tests that are failing and generate new baseline images from Linux.

nickpowersys:
This is the successful build on Travis under districtdatalabs. I'm not sure if the status "Expected - Waiting for status to be reported" is expected.
https://ci.appveyor.com/project/districtdatalabs/yellowbrick/builds/24864530

ndanielsen:
@nickpowersys great work on this. I'm going to test this on my constellation of systems and services tomorrow. =)

ndanielsen:
Spent some time testing this on:
- Windows 10 (conda) with 3.7
- MacOs (conda / pyenv) with 3.6/3.7
- Ubuntu (conda / pyenv) with 3.6/3.7

On Ubuntu (conda) and Mac with pyenv, I was getting an interesting error:
`matplotlib.testing.exceptions.ImageComparisonFailure: Image sizes do not match expected size: (550, 800, 3) actual size (1100, 1600, 3)`

The test failures were in these test files:
- test_target/test_feature_correlation.py
- test_text/test_freqdist.py

Outside of the test failure, the image being drawn would come up in a python instance popup and required to be closed before the testsuite would carry on.

On mac and ubunut os, putting both of these key:values together in a matplotlibrc file seemed solve the issue.:
```
backend: Agg
figure.figsize :  6.4, 4.8
```
Looking at how matplotlib tests, perhaps we should include some form of this in the `setup.py`:

https://github.com/matplotlib/matplotlib/blob/765d452bdd1125c45d2ac91c62f064b5616a72fd/setup.py#L225-L235

Any thoughts? 

Thanks again for following up on this. I think that we're getting closer. :1st_place_medal:

nickpowersys:
I was able to eliminate the popups on Mac pyenv by replacing poof with finalize in

- test_target/test_binning.py
- test_target/test_feature_correlation.py
- test_text/test_freqdist.py

This connects with the audit mentioned at 
https://github.com/DistrictDataLabs/yellowbrick/pull/705#issuecomment-459023224

I can test this change with Ubuntu conda and Windows conda next, and see if this also resolves the image size issue.  

If it works, then I could commit these changes to ensure they work on the builds. How does this sound?

ndanielsen:
@nickpowersys great work! Sounds like a good plan to me.

ndanielsen:
@nickpowersys thanks for your efforts of this. This PR represents a ton of work and another step towards making the yellowbrick contributor experience a lot more friendly and robust across platforms. 

I've run this locally on the mac and ubuntu machines without any test failures. I'm happy to approve this, and I'm going to leave the honor of merging it in to you. =)

ndanielsen:
@nickpowersys ooops, looks like the tooling is setup so that only maintainers can merge in. Forgive the unintentional slight and accept the intended honor. 💯

nickpowersys:
@ndanielsen Thanks so much for reviewing this in-depth to assure the contributor experience is excellent for all platforms, and suggestions for the PR! I will be in communication on the other PR for Appveyor next!

Issue #423, opened 2018-05-15
Windows Image Similarity Tests


ndanielsen:
A good suggestion came out from @jkeung at pycon2018, we should investigate creating separate baseline images for windows machines.

Recommend that we had a check in the base.py for if the machine is windows and then direct the test to checkout the windows folder in baseline images or otherwise, significantly increase the tolerance for them.

rebeccabilbro:
Here are the [actual images and diffs](https://drive.google.com/open?id=0B5pgYPYyUFpzS3FteGgzcjRRNlNSOWtkd1Njb3d1aGZJRy04) that result from running the tests on Windows with Python 3.6 and YB 0.8.

bbengfort:
Following along on this -- 

@drwaterman and I went through all the windows image similarity errors and recorded the RMSEs and inspected them. The values can be found here:

[YB XFails List](https://docs.google.com/spreadsheets/d/1coQakoVoxqev2UVLpOolrd6uNhi9aFhm2PWkkgBCjZA/edit?usp=sharing)

We couldn't find any similarity errors that weren't related to text. @drwaterman notes in #567:

> @bbengfort If you are interested in reducing the RMSE for the Windows tests, I did a little playing around and I think if you turn off text antialiasing you can get rid of some of the text shadowing that we saw in the windows diff files. I added the following to a couple of the tests:
>
> At imports: import yellowbrick.style.rcmod as yb_rcmod
> At the beginning of a testing method: yb_rcmod.set_aesthetic(rc={'text.antialiased': False})
>
> Then comparing the output images from a windows machine and a mac machine, most of the text shadows disappear.
>
> There may be further rcmod parameters that can do similar things for the other rendered elements, I only tested with the text.
>
> The reduction in RMSE wasn't huge so you might not want to spend the time to implement this change unless a fix can also be found for the other elements.

bbengfort:
I propose the following:

Phase 1: Increase tolerance if on Windows 
Phase 2: Implement backoff with tiered tolerances 

Backoff strategy:

1. before calling `assert_images_similar`, call `viz.finalize()` to ensure the entire viz is produced. 
2. have `assert_images_similar` accept a target threshold and a backoff threshold 
3. The backoff strategy will generate the image and test it, passing when the image comparision is <= the target threshold, and backing-off so long as the <= the backoff threshold; if > than backoff threshold or if all backoff images are exhausted, then the test fails. 
4. If we backoff, we can issue a test warning 

[...]

PR #864, opened 2019-05-27
Add: add a windows_tol for the visual unittest suite

ndanielsen:
This PR adds in adding another optional argument for the visual unit test suite specifically for windows environments. 

The optional arg is `windows_tol` to be used in this manner `self.assert_images_similar(ax=g, windows_tol=15.0)`

This additional option will give us greater confidence and a more stable testsuite for windows environments. Currently, many of the image comparison tests are set to `xfail` or expected fail as certain attributes like fonts are significantly different between mac and windows envs. 

This is a incremental change as part of a larger body of work around improving the contributor experience for window users.


ndanielsen:
Related issue is https://github.com/DistrictDataLabs/yellowbrick/issues/423

pdamodaran:
@ndanielsen - thanks for submitting this PR!  This will help make our windows testing more robust as we can better control the tolerance levels based on the OS.

I just have a couple of comments regarding the assignment of the variables that I will annotate inline.  Let me know what you think!

tests/base.py
```diff
@@ -201,6 +207,7 @@ def __init__(self, stack, visualizer=None, ax=None, tol=0.01, ext=".png",
 
         # Save other image comparison properties
         self.tol = tol
+        self.windows_tol = windows_tol
```
pdamodaran: My thought here is rather than adding a new class attribute, we could update self.tol using the logic you added on line 315, so change this to something like:

            if os.name == "nt" and windows_tol is not None:
            self.tol = windows_tol

tests/base.py
```diff
@@ -305,8 +312,13 @@ def compare(self):
...
             )
 
+        if os.name == "nt" and self.windows_tol is not None:
+            err_tol = self.windows_tol
+        else:
+            err_tol = self.tol
+
         # Perform the comparison
-        err = compare_images(expected, actual, self.tol, in_decorator=True)
+        err = compare_images(expected, actual, err_tol, in_decorator=True)
```
pdamodaran: If self.tol is updated in the init method, this change will not be needed.

tests/base.py
```diff
@@ -153,6 +156,9 @@ class ImageComparison(object):
         difference. The test fails if the average pixel difference is greater
         than this value.
 
+    windows_tol : float, default: 0.01
+        The tolerace (tol) parament for the windows operating system environment.
```
pdamodaran: did you mean "parameter"?

ndanielsen: Oops, fixed it!

tests/base.py
ndanielsen: great idea!

PR #865, opened 2019-05-27
Banner Image for Social Preview

bbengfort:
An update for the readme_imgs.py script to generate a single banner image for use on the GitHub social preview. This update places all images on a single figure with matplotlib and loads our logo into the middle of the image.

banner

The social banner dimensions need to be 1280x640 -- so this image is a little tall; working on cropping it now to make it fit.


bbengfort:
@rebeccabilbro it took a bit of fiddling with the `figsize` property to finally get things sorted - I think your addition of the padding and the tight layout got things very close! I did learn some interesting things that I think will be useful for the `GridVisualizer` - so it was definitely an important exercise! Thank you so much for doing the bulk of the original work to make it possible! 

The three original images aren't deleted, just modified -- we didn't control all random aspects in the visualization so they do change when you run the script repeatedly. I'm not sure what the 99% and 100% figures are referring to in the GitHub diff ... maybe GitHub has something similar to our image similarity? But the bottom line is that the `README.md` shouldn't be broken and the three original images can be regenerated on demand, in fact, this is the default when you run the script. You have to specify an option to generate the full banner image, and another option to replace the prediction error plot with the YB logo. 

As for replacing the README image - I'm fine with doing that if you'd like; I'm also fine with keeping it the way it is or replacing the three with the full image but no logo. Let me know your preference and I'll make the change!

bbengfort:
@ndanielsen @rebeccabilbro also check out the social preview image - I squashed it to the right aspect ratio but didn't update the script since the social preview is a limited use case.

PR #824, opened 2019-04-22
Additional tests for elbow detection in Elbow Visualizer

pswaldia:
This PR adds tests for a recently merged PR #813. 

I have added the following tests:   
1.  Added `tests/test_utils/test_kneed.py` that contains tests for `KneeLocator` class implemented in `kneed.py` 
2. Tests related to  `DataGenerator` have been removed since we do not require it in yellowbrick implementation of `kneedle` method

@bbengfort I ran the tests for this particular file and all tests passed. Please have a look .
Thanks.

PR #871, opened 2019-06-04
Add: add image similarity test for locate_elbow in the KElbowVisualiz…

ndanielsen:
Adding a basic image similar test for when `locate_elbow=True` in the scenario below.

```
        visualizer = KElbowVisualizer(
            KMeans(random_state=0), k=5,
            metric="calinski_harabaz", timings=False, locate_elbow=True
        )
```
The elbow score is already stripped out in the image similarity test suite, but it's a good thing to be sure of. =)


ndanielsen:
Related to this PR:

https://github.com/DistrictDataLabs/yellowbrick/pull/824

pdamodaran:
Hi @ndanielsen  - thanks for adding this test!  It is good to be sure as we add new features that we can ensure that the image comparison tests still work.  One thing I believe that this test will also benefit from is if you could add one additional test to check the self.elbow_value_ as none of the tests currently have it and this is a value that would get populated when `locate_elbow=True` is set.

ndanielsen:
@pdamodaran made requested changes. Great code review btw!

pdamodaran:
@ndanielsen - thanks for adding the PR to get the AppVeyor build to pass.  I will go ahead and merge this in.

PR #883, opened 2019-06-12
Analyze and address image comparisons for Travis Miniconda builds to fix #690

### LINKED ISSUE #690: Some YB tests fail with conda Python 3.7 & Matplotlib 3.x

nickpowersys:
Since image comparison tests are not yet addressed for Travis as they have been on Appveyor, this PR handles the image test failures and fixes #690. I will
- [x] Make Miniconda builds fully active in .travis.yml as long as image comparisons are passing
- [x] Assess the roughly 40 failing image comparison tests based on diffs with regular Python on Linux
- [x] Determine if there is significant overlap with the set of image comparisons on Appveyor with RMSE>0.1. Are there are any recurring patterns that could potentially be addressed through code across environments?
- [x] Recommend how to handle failing comparisons based on the analysis
- [x] Based on feedback, indicate the failing tests' status by resolving them primarily through configuration
- [x] Update docs by removing reference to past PyQt4 dependency affecting Anaconda on Linux


ndanielsen:
@nickpowersys thanks for starting this PR. Let me know when it's ready for review or you want to bounce some ideas off me and the other maintainers. 

Thanks again for tackling this critical stability and contributor experience piece of the yellowbrick road!

nickpowersys:
@ndanielsen Please let me know how this sounds. I think this will give an actionable base from which to address issues that are common across both platforms and Python distributions. The ultimate goal is not to create but to remove xfails.

By linking to issues that summarize the reason for failure and known affected tests, xfails can give a sense of how many tests can be resolved through a single fix.

Once the Travis conda builds are running, the many Appveyor xfails outside of these 40 tests that are still in place but that may have been resolved by the removal of labels and legends can be checked again and removed wherever possible in another PR.

I found that all of the failing Linux conda tests have either an existing xfail for win32, or a conditional tol just for win32, or a windows_tol.

About 2/3 of the failing tests are text-related, and can clearly be resolved through code. The code updates will resolve Appveyor xfails as well.

The text-related reasons apply to 31 plots in 40 failing tests for one of the following reasons:
- only one of multiple sub-plots has text removed (form of plots are same)
- there are labels on the plot
- there are labels to the right of the plot
- there are labels on secondary plots (form of plots different from main)

Non-text-related differences apply to 13 plots, such that a few elements or gridlines are positioned slightly differently.

There are some tests that have both text and non-text differences from the baselines.

I can
- create one issue for each of the four text reasons and one non-text reason above, referencing the specific tests to which each issue applies
- for the 40 failing tests, update or put in place xfails that reference the issue numbers in the xfail reasons; remove the win32 conditions from the xfails within the 40 tests
- if any of the failing tests has an if condition with a tol relating to win32, the condition will be removed, since the test has been confirmed to fail across platforms
- if there is no existing xfail but a windows_tol, it will be removed, since the test has been confirmed to fail across platforms

Before creating the issues, I'm putting the numbers for the Linux conda RMSE and existing tols here where they exist, just to show that the numbers are comparable.

The leading numbers are just indexes from the list of 40. All other tests not included here have existing xfails for Windows.

test_classifier/test_class_prediction_error
4 test_integration_class_prediction_error 9.499 on linconda, windows_tol=9.5 plot element differences

test_classifier/test_prcurve
7 test_binary_probability_decision Y - 4.522 on linconda, windows check has tol = 4.6 - element labels
8 test_custom_iso_f1_scores Y 4.358 on linconda, windows check has tol = 4.5 - element labels
9 test_multiclass_probability Y 4.729 on linconda, windows check has tol = 6.6 - element labels
10 test_quick_method Y 4.725 on linconda, windows check has tol = 5.8 - element labels

test_cluster/test_icdm
12 test_affinity_tsne_no_legend 1.557 on linconda, windows check has tol = 2.75 - labels on plot itself
13 test_kmeans_mds Y 4.461 on linconda, windows check has tol = 4.9 - labels on plot itself
14 test_quick_method Y 1.48 on linconda, windows check has tol 2.75 -labels on plot itself

test_features/test_importances	
15 test_integration_coef 15.617, windows_tol=16.2 horizontal lines
16 test_integration_feature_importances 12.772, windows_tol=13, horizontal lines
17 test_integration_quick_method 14.816, windows_tol=15.0, horizontal lines
18 test_multi_coefs_stacked 16.606, windows_tol=17.5, horizontal lines

test_features/test_jointplot
19 test_columns_none_x_y 2.471, windows_tol=2, plot elements (scatter) and vertical line slightly off
20 test_columns_single_int_index_numpy 2.787, windows_tol was applied at 5.5 - plot elements scatter and vertical line slightly off
21 test_columns_single_str_index_pandas 4.513, windows_tol at 5.5 - same as previous
22 test_columns_double_index_continous_y_hist 2.177, windows check has tol = 4.0, plot elements positioned slightly differently
23 test_columns_double_index_discrete_y_hist 0.157, windows check has tol = 4, labels on plots besides the main one
24 test_columns_double_int_index_numpy_no_y_hist, 0.157, windows check has tol = 4, labels on plots besides the main 

25  test_columns_double_str_index_pandas_no_y_hist, 2.177, windows check has tol = 4, TWO REASONS: plot elements slightly askew, AND labels on plots besides the main one
26 test_columns_none_x_hist, 0.157. windows check has tol = 4, labels on plots besides the main one

27 test_columns_none_x_y_hist, 2.85, windows check has tol = 3.5, TWO REASONS: plot elements slightly askew, AND labels on plots besides the main one
28, test_columns_single_int_index_numpy_hist, 5.469, windows check has tol = 0.5, TWO REASONS: plot elements, AND labels on plots besides the main one
29, test_columns_single_str_index_pandas_hist, 4.835, windows check has tol = 0.5, TWO REASONS

Thanks!

ndanielsen:
Apologies for the delay in responding to this. I'll take a look at this later today. =)

.travis.yml
```diff
@@ -1,20 +1,49 @@
...
 
 before_install:
-- sudo apt-get update
-- sudo apt-get build-dep python-scipy
+- sudo apt-get update;
+- if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+    MINICONDA_OS="Linux";  
+    sudo apt-get build-dep python-scipy;
+  else
+    MINICONDA_OS="MacOSX";
```
ndanielsen: can you put a comment here explaining why we're setting `MINICONDA_OS="MacOSX";`. 

It doesn't appear that we're testing `osx` yet. =)

ndanielsen:
@nickpowersys thank you for the detailed and highly analytical write up for this. 

```
About 2/3 of the failing tests are text-related, and can clearly be resolved through code. 
The code updates will resolve Appveyor xfails as well.
...
create one issue for each of the four text reasons and one non-text reason above, 
referencing the specific tests to which each issue applies
```
Feel free to tackle this first approach / issue that you identified that is mostly targeted at text related xfails. To make it easy to review and get quickly accepted in, as normal let's try to keep the PR (or continuation of this one) highly targeted and atomic in focus. 

After we sort out the easy fixes, we can take a deeper look into the non-text / drawing issues and figure out if those differences are worth the effort to fix and maintain.

.travis.yml
nickpowersys: The `MINICONDA_OS` variable is used to grab the OS-specific Miniconda installer file. I wanted to avoid hard-coding the file name where it could be avoided. I like it being explicit that Linux is not the only possible OS.

nickpowersys:
Although Github says "This branch is out-of-date with the base branch," there is no button that allows me to fix that. I'm not sure why there was a button before, but not now. Should I just do

```
git checkout develop
git merge upstream/develop
git checkout travisminiconda
git merge --no-ff develop
```

nickpowersys:
It seems like the most promising way to have a fix that scales well is to address text as Matplotlib does, based on local Freetype. Otherwise, any given fix for text will only apply to roughly 5-10 tests.

I can still indicate the reasons for the tests failing, but instead of having one issue for each of the 4 text-based reasons, I can have a single issue indicating all of the tests that have text-only reasons for failing. That could be addressed in a separate PR (definitely, not having to deal with text differences at all would apply both to these tests and new tests).

After this PR is resolved and the Appveyor xfail issue is resolved by removing the ones that no longer fail, would it make sense to ensure that any tests where tol is elevated or win_tol=# have been investigated, such that tol/win_tol means a decision has been made to leave them as-is? It seems like the number of tests where this applies but that have not been investigated yet should be manageable.

Whereas in the past, it was not easy to know why a test was failing on Windows, it seems that the same tests will nearly uniformly have the same driver on Linux conda environments, and so there should be a low effort to understand the reason for comparisons failing in general, as these tests show.

If the Freetype-based approach has some reason it will not work for builds and local testing, and a decision is made that the text-related issue linked from the xfail will not be fixed, then the xfail could be removed in favor of tol or win_tol. Do you think this would provide a good long-term convention where a contributor will be able to tell just from the code whether a test is either a "possible fix" (xfail) versus a "won't fix" (tol/win_tol)? I can follow through accordingly.

nickpowersys:
I can put a comment in .travis.yml about MacOSX, like "Obtain OS-specific
installer based on MINICONDA_OS; allow for Mac in future build matrix".

ndanielsen:
@nickpowersys I'm still mentally retracing your journey in this [comment](https://github.com/DistrictDataLabs/yellowbrick/pull/883#issuecomment-504094005) so please forgive if I get something a bit wrong.   

I like how the approach is shaping up with a potentially scalable solution (Freetype) to most of the image comparison tests along with a fall back with the image tolerances.

My only concern is that in the short-medium term with the current implementation in this PR that we'll lose the guard rails on the tests were the `tolerances` have been removed and now we are xfailing them on all environments. As this library is under active development from a variety of contributors, I think that we might be inviting bugs by allowing a gap in test coverage while a bigger, more comprehensive solution is in the works.  

Is there an intermediate solution to this where we can keep up the test coverage while keeping your excellent work in place? 

Also the intent of this PR is starting to diverge a bit. It would be great to get a smaller win here and start another PR that is specifically targeted on another victory (ie Freetype) in the making?

.travis.yml
ndanielsen: As we haven't gotten to `osx` testing yet, can we remove it for now? It might be confusing for other future contributors to see this artifact and get the impression that we are testing `osx` but aren't yet. 

Does that make sense?

ndanielsen:
Just want to reiterate that I really like where this effort is going and thank you for your efforts here. A lot of valuable things have been implemented here, and I want to get the good stuff merged in quickly to get other contributors to be able to leverage it while we put in place the game changers.

Does that make sense or have I been through too many agile workshops? =)

nickpowersys:
I'm not sure how much effort resolving the issue through Freetype may take, as it is definitely an unknown area, especially given that it would need to span multiple environments. The first priority I had was to ensure that 690 could be resolved for potential contributors. Identifying the set of tests affected by text and the possible impact of scalability is what I hoped to accomplish. But to address your feedback on the implementation so far, I hope the following sounds like a good solution.

Overall, the roughly 30 xfails for text being in place across all environments would cover both Windows 
for both Python distributions and Linux conda, across which (Windows generally, Linux conda specifically) there are verified failures for the image comparison tests. And Linux with PyPI Linux. Including it as well even though it does not have image comparison failures could undo non-image comparison coverage.

As an intermediate solution, I could simply add a condition in the xfails that the environment is not Linux with regular Python (using a function equivalent to the one for Windows conda I had). Then, non-image comparison test failures could occur within the Travis builds. Do you think that the non-image comparison tests will be sufficiently covered that way for these roughly 30 tests? I'm not as familiar with all of the types of tests, but I am guessing that they would be far less prone or not prone at all to environment-specific conditions as image comparisons.

If new visualizers and tests are added, the updates I have made here will not affect them, since there are no changes to the image comparison test in general, only to test-specific configurations through xfails or tols.

With the tolerances that I put for Linux conda and Windows, the failures were also verified across those environments.  If new elements were added to existing visualizers, then this should exceed an existing tol within the builds to trigger failures. If elements were removed, the tols could be checked and revised down.

nickpowersys:
I am un-drafting it to resolve the "this branch is out-of-date" as needed and update the affected tests with a uniform condition for the xfails as described, nothing the assertions in the tests to see if any might be affected by the environment.

If the condition needs to be modified for the PR, having the same condition in multiple places would make it easy to revise through a find-and-replace.

nickpowersys:
Please let me know if there is anything else that should be addressed. Note that I indicated in the Getting Started section about creating environments that conda developers should add the conda-forge channel for testing dependencies. This will only have any effect if they next install conda packages. If conda packages are preferred, then there is one more step that is needed to install the dependencies before activating:

`conda env create -f tests/requirements.txt -n yellowbrick python=[3.6 or 3.7]`

The question I have about this is, would it be preferred for them to test by pip installing (meaning they will have packages through PyPI), or to use conda packages?

If they were to use conda packages, they would have an experience like that of the typical conda user with conda packages, and their local testing and testing on the builds would be the same.

If they were to use pip packages, then they could generate baseline images that would reflect the PyPI Python environment.

docs/contributing/getting_started.rst
```diff
@@ -43,6 +43,10 @@ Once forked, use the following steps to get your development environment set up
 
         $ virtualenv venv
 
+    To develop with a conda environment, the conda-forge channel is needed to install some testing dependencies. The following command adds the channel with the highest priority::
+
+        $ conda config --add channels conda-forge
```
ndanielsen: this is awesome! thanks for adding this to the documentation!

ndanielsen:
@nickpowersys I'm going to review this in depth on Monday. Thanks for your efforts on this. On my first quick pass, it looks like test coverage has been maintained and we have ✅ s for the CI environment which is my first sniff test. 🥇

ndanielsen:
Pulling out some of the specific questions I see in your comments:

`Do you think that the non-image comparison tests will be sufficiently covered that way for these roughly 30 tests? `

My working assumption is that non-image comparison tests are working robustly across environments 
and operating systems. If that isn't the case, then we have some huge problems. =)

`The question I have about this is, would it be preferred for them to test by pip installing (meaning they will have packages through PyPI), or to use conda packages?`

This is a great question, and I don't have as excellent of an answer in response. By convention and habit,  I don't think that we have a **deep** conda user in the wheel house. For `conda contributors`, I think that the ideal situation is that `development` environment matches the `production` environment. So having `conda contributors` using pypi packages might introduce strange bugs in the end user experience.  

Obviously, the end goal is to get the `conda` and `pypi` image comparison tests running identically and I think that is the next steps (through whatever means) after we get this PR merged in.

Please let me know if I correctly interpreted your questions.

ndanielsen:
@nickpowersys I'm still working through this PR by verifying tests and implementation with a big emphasis on maintainability and not unnecessarily ceding any test coverage. 

To summarize for posterity, in this PR we are:
- Adding `miniconda` testing in travisCI in the linux environment
- Updating some documentation around installing conda
- Updating several of the image comparison tests to xFail if the os is not linux and the pypi is not the package manager
- Tweaking some of the tolerances

nickpowersys:
Thank you, you have answered my questions. I am not making any other commits as you conduct the review.

ndanielsen:
@nickpowersys apologies I got a little distracted on an appveyor issue that popped up:

https://github.com/DistrictDataLabs/yellowbrick/pull/894

Any thoughts on how to fix this? I spent a good chunk of time experimenting without success.

tests/base.py
```diff
@@ -29,10 +29,10 @@
...
 
 
-def is_winconda_env():
-    return (os.name == 'nt' and
-            os.path.exists(os.path.join(sys.prefix, 'conda-meta')))
-
+def linux_pypi():
+    linux_os = sys.platform.startswith('linux')
+    pypi = not os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
+    return linux_os and pypi
```
ndanielsen: see my comment about naming conventions below and how this might transform a bit.

Also I think it is appropriate to call this function and set the constant in the `Module Constants` below

nickpowersys:
I will be able to look at the Appveyor build a bit and incorporate your
suggested improvements for Travis Miniconda later today.

nickpowersys:
I looked at alternative `install` approaches to address issue #894, and the problem is due to a change to that Appveyor Miniconda37 image that has no readily available workaround. There is a fix that is a bit confusing at first glance and would only be a temporary approach for that reason.

Since the Appveyor `install` step creates a new conda environment and does not rely on the base Anaconda Python that comes with the Appveyor image, it is possible to just substitute the Miniconda36 image (and also reflect the matching `MINICONDA_VERSION: "4.5.4"`) as follows. The Miniconda job for 3.6 remains as-is. This revision for 3.7 works on Appveyor:

First,
```
# Work-around substituting Miniconda36 base for Miniconda37 due to LoadLibrary() issue #894
- PYTHON: "C:\\Miniconda36-x64"
  PYTHON_VERSION: "3.7"
  MINICONDA_VERSION: "4.5.4"
  PYTHON_ARCH: "64"
```

and next, specify the Python version explicitly as follows by adding the `python=` option:

`conda env create -f tests/requirements.txt -n yellowbrick python=%PYTHON_VERSION% &&`

I could submit another PR after this one if this is reasonable given the context.

I will finish the revisions you suggested and push them.

tests/base.py
nickpowersys: I see what you mean here - will make that change accordingly so the constant is imported instead of the function.

tests/test_base.py
```diff
@@ -175,7 +174,7 @@ def test_draw_visualizer_grid(self):
         self.assert_images_similar(grid)
 
     @pytest.mark.xfail(
-        sys.platform == 'win32', reason="images not close on windows"
+        not LINUX_PYPI, reason="text differences described in #892"
```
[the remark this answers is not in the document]
nickpowersys: Implemented in 07f253e.

tests/base.py
nickpowersys: Implemented in https://github.com/DistrictDataLabs/yellowbrick/pull/883/commits/2e22c3e3bb583f175941dff8721213fcca652f38

tests/test_base.py
```diff
@@ -155,7 +154,7 @@ class TestVisualizerGrid(VisualTestCase):
     """
 
     @pytest.mark.xfail(
-        sys.platform == 'win32', reason="images not close on windows"
+        not LINUX_PYPI, reason="text differences described in #892"
```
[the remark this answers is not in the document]
nickpowersys: Implemented in https://github.com/DistrictDataLabs/yellowbrick/commit/07f253e8c4d42639f9e657f7c3e9fbc11dd1a83e

nickpowersys:
Please let me know if I can do anything else for the PR.

tests/base.py
nickpowersys: https://github.com/DistrictDataLabs/yellowbrick/blob/07f253e8c4d42639f9e657f7c3e9fbc11dd1a83e/tests/base.py#L32-L35

nickpowersys: https://github.com/DistrictDataLabs/yellowbrick/blob/2e22c3e3bb583f175941dff8721213fcca652f38/tests/test_base.py#L29

ndanielsen: this is looking great!

tests/test_base.py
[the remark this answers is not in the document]
ndanielsen: awesome!

ndanielsen:
@nickpowersys thanks for the tremendous job on this. This has significantly increased the robustness of our test suite across many platforms. I'm very happy to approve this!

As for the appveyor issue that you analyzed in https://github.com/DistrictDataLabs/yellowbrick/pull/883#issuecomment-506615201, I think that we should wait for a less hacky / workaround fix to surface. As we're already testing conda in a linux environment as a result of this PR, we're in a great spot and much better off than we were before!

nickpowersys:
Thank you again for your much-appreciated feedback, @ndanielsen

PR #888, opened 2019-06-17
Addressing issue#697: Updated the threshold logic

### LINKED ISSUE #697: DecisionThreshold seems to compute the minimum number of thresholds across trials, not max
Presumably the following code should find the max number of thresholds found across all the trials so as not to ignore the values in the densest threshold list across all trials:
```
# Compute maximum number of uniform thresholds across all trials
n_thresholds = np.array([len(t['thresholds']) for t in trials]).min()
self.thresholds_ = np.linspace(0.0, 1.0, num=n_thresholds)
```
But this finds the minimum. So we may be losing information by bisecting with the sparsest number of thresholds rather than the densest.

naresh-bachwani:
This PR fixes #697. Although the method suggested in the issue is to use all the thresholds values, I used a different approach. 
**Suggested Method**
Update `self.threshold_ = np.sort(np.concat([t['thresholds'] for t in trials]))`. This leads to a few thousand values for self.threshold_. Hence this might make visualizer slower.

**My method**
n_threshold = max(n_threshold, 101)
This ensures accuracy up to two decimal places and will work faster. 

This is a small change and I can revert back to the suggested method if maintainers recommend so.

Possible Todo:
- [ ]  Add param for accuracy of threshold up to certain decimal points.


lwgray:
Hi @naresh-bachwani,  As always...Thank you immensely for your thoughtful contributions.  In this instance the change you introduced does not fix #697.  The problem is a bit more nuanced and will require us to rethink our approach.  Also this issue/PR isn’t vital for our progression towards v1 thus I will close it.  Maybe we can continue this discussion in the comments of #697.   I definitely would like input from @bbengfort and @rebeccabilbro

Issue #874, opened 2019-06-04
ProjectionVisualizer: unifying functionality of PCA and Manifold


bbengfort:
One of the basic high-dimensional visualization techniques that Yellowbrick makes use of is to decompose or project a high dimensional space into 2 or 3 dimensions to display the data as a scatter plot. Projections of this kind reduce the amount of space between points (decreasing sparsity) but can still give us some intuition of structures in the higher dimensionality. Currently, we have three primary decomposition methods that use this technique:

- Manifold: wraps a non-linear transformer from `sklearn.manifold` to produce embeddings
- PCA: uses linear principal component analysis to decompose to lower dimensionality 
- Text: TSNE and UMAP visualizers do the same as Manifold but with text-specific helpers 

These visualizers have a lot of shared functionality that can be combined to streamline these kinds of visualizations and make it easier to extend them (e.g. to add ICA, Fast PCA, etc. to the PCA decompositions, or to extend the text visualizers to use the manifold visualizations). 

I propose we create a `ProjectionVisualizer` base class or mixin that knows how to:

- Wrap a transformer to project `X` into `X'` of shape `(n_instances, 2)` or `(n_instances, 3)`
- Create a scatter plot for 2D or 3D plots (implemented in PCA)
- Identify the type of the target and add colors (implemented in Manifold)
- Subselect the features to use in `X` for the projection 

This shared functionality could then be easily used by PCA, Manifold, etc. 

The following notes about the class hierarchy:

- The `MultiFeatureVisualizer` produces a `self.features_` attribute on `fit()` which is useful in PCA for biplots and to understand the original feature set. 
- The `DataVisualizer` produces `self.classes_` from y and is supposed to "provide helper functionality related to target identification" but does not currently implement this yet (it is implemented on `Manifold`)
- `yellowbrick.contrib.ScatterVisualizer` _might_ be valuable to be moved to `yellowbrick.draw.scatter` and use as a mixin to handle part of these cases; though I don't necessarily want to confuse things too much.  
- The `JointPlot` visualizer would also benefit from the target color handling things from above. 

This implies that the `ProjectionVisualizer` is a `DataVisualizer` and that the `DataVisualizer` needs to be updated to handle the target identification stuff that is in `Manifold`. It also implies that `JointPlot` should be a `DataVisualizer` as well. 

[...]

Issue #318, opened 2018-03-03
Overhaul unittest and fixtures


bbengfort:
Remove dependency on `unittest.TestCase` by replacing assertions with `assert` statements. This also involves using `pytest.raises` and `pytest.skip` decorators. 

Use PyTest fixtures for datasets and other fixtures. 

~Add PEP checking to pytest (alongside pyflakes)~

- [x] rename `ThingTests` to `TestThing` 
- [x] assertions 
- [x] pytest.skip 
- [x] dataset fixtures 
- [ ] ~PEP8~

Alongside this we should also do a PY2 removal check, namely anywhere:

```python
try:
    from unittest import mock
except ImportError:
    import mock
```

should be replaced with just `from unittest import mock`

NealHumphrey:
@bbengfort what's the motivation for removing the unittest dependency?

bbengfort:
@NealHumphrey the short answer is that weird things happen when you mix and match pytest and unittest - my primary concern is test discovery. PyTest expects `TestThing` vs `ThingTests` (although I guess that is more of a nose thing than a unittest thing). Plus pragmas and marks (e.g. skip, xfail) behave slightly differently. I think it's better to standardize on one or the other. 

Because our feeling is that people are more comfortable with pytest and because of features like pytest-flakes, pytest-spec, etc. we've chosen to move to pytest. Pytest does play well with unittest so there is no rush on this, but the sooner it's standard the easier it will be to diagnose issues in the tests (of which we have a few).

rebeccabilbro:
Yeah @NealHumphrey, this is sort of a follow on from #291 -- after we discovered last year that [nose wasn't being maintained anymore](https://twitter.com/llanga/status/826144101833732097) and started converting all the tests to pytest.

NealHumphrey:
Adding a TODO:
- [ ] Update the [documentation](http://www.scikit-yb.org/en/latest/contributing.html#testing) for contributors

NealHumphrey:
As noted in my pull request, I had trouble running `pytest` locally including pyflakes and coverage. On Windows, haven't tried it on Mac yet. 

What I tried:
1) Create a new virtual environment (using conda) 
2) activate it `activate yb`
3) pip install pytest (b/c it's not in the requirments.txt, see TODO in previous comment), which installed 3.4.2 plus some dependencies, and `pip install --upgrade pyflakes`  (installed v 1.6.0)
[...]

PR #893, opened 2019-06-25
Addressing issue 874: Updated DataVisualizer to handle target_type_identification

naresh-bachwani:
This PR handles a subpart of #874. It updates the data visualizer to handle target_color_type. This makes use of the `_determine_target_color_type` function already implemented in `Manifold` visualizer. This visualizer is supposed to be the super class of Projection Visualizer proposed in #874. This visualizer currently misses the feature of target color type identification. 

### TODOs

- [x] Add test for target identification
- [ ] Discuss with mentors regarding the removal of duplicate code from its sub classes.


bbengfort:
@naresh-bachwani thanks for tackling this! Let me know when you're ready for a review.

naresh-bachwani:
@bbengfort it is ready for review!

tests/test_features/test_base.py
```diff
@@ -47,3 +48,57 @@ def test_subclass(self):
     #     visualizer = FeatureVisualizer()
     #     with self.assertRaises(NotImplementedError):
     #         visualizer.poof()
+    
+class DataVisualizerBaseTests(MultiFeatureVisualizer):
```
bbengfort: The test suite shouldn't be a `MultiFeatureVisualizer` right? It should just be an `object`.

bbengfort: Also, I think this might need to be `TestDataVisualizerBase` per [Conventions for Python test discovery](https://docs.pytest.org/en/3.0.0/goodpractices.html#test-discovery) -- the `Tests` suffix is for `unittest`, e.g. if you extended `VisualTestCase` -- but we're trying to move away from that per #318

yellowbrick/features/base.py
```diff
@@ -177,7 +186,7 @@ class DataVisualizer(MultiFeatureVisualizer):
     """
 
     def __init__(self, ax=None, features=None, classes=None, color=None,
-                 colormap=None, **kwargs):
+                 colormap=None, target=AUTO, **kwargs):
```
bbengfort: Don't forget to update the docstring with the new argument!

yellowbrick/features/base.py
```diff
@@ -22,8 +22,17 @@
...
+from yellowbrick.style import resolve_colors
+from yellowbrick.exceptions import YellowbrickValueError
+
 from sklearn.base import TransformerMixin
 
+AUTO = "auto"
+SINGLE = "single"
+DISCRETE = "discrete"
+CONTINUOUS = "continuous"
+
```
bbengfort: Have you used Python enums before? I'd like to explore their use here. It would allow us to more easily check if the target type string was valid by parsing the enum and group these constants together in a more accessible way for the user. 

https://docs.python.org/3/library/enum.html

yellowbrick/features/base.py
```diff
@@ -190,6 +199,7 @@ def __init__(self, ax=None, features=None, classes=None, color=None,
         # Visual Parameters
         self.color = color
         self.colormap = colormap
+        self.target=target
```
bbengfort: Likely we'll need some kind of validation that the target string supplied by the user is correct and raise a `YellowbrickValueError` on `__init__` if not. E.g. if the string is misspelled. Above I suggest an enum as one mechanism to do this, though you could use a set and some string manipulation as well if it is preferable.

yellowbrick/features/base.py
```diff
@@ -215,13 +225,58 @@ def fit(self, X, y=None, **kwargs):
...
-            # TODO: Is this the most efficient method?
-            self.classes_ = [str(label) for label in np.unique(y)]
+        self._determine_target_color_type(y)
+
+        # Compute classes and colors if target type is discrete
+        if self._target_color_type == DISCRETE:
+            # Store the classes for the legend if they're None.
+            if self.classes_ is None:
+                # TODO: Is this the most efficient method?
+                self.classes_ = [str(label) for label in np.unique(y)]
```
bbengfort: Nice! Way to store the classes only on a `DISCRETE` target type! I appreciate that!

yellowbrick/features/base.py
```diff
@@ -215,13 +225,58 @@ def fit(self, X, y=None, **kwargs):
...
+            
+# This fails in case of PCoords if subsampled target misses some values.
+            # Ensures that classes passed by user is equal to that in target
+#            if len(self.classes_)!=len(np.unique(y)):
+#                raise YellowbrickValueError("Number of unique target is not " 
+#                                            "equal to classes")
```
bbengfort: Besides this comment being formatted weirdly - I'm not sure I understand what's going on here or why this is commented? Was this code here before? If so, can you look at git blame to see why it was commented out? I'd like to fix it if possible.

yellowbrick/features/base.py
```diff
@@ -215,13 +225,58 @@ def fit(self, X, y=None, **kwargs):
...
+            color_values = resolve_colors(n_colors=len(self.classes_), 
+                                          colormap=self.colormap, colors=self.color)
+            self._colors = dict(zip(self.classes_, color_values))
+            
+        # Compute target range if colors are continuous
+        elif self._target_color_type == CONTINUOUS:
+            y = np.asarray(y)
+            self.range_ = (y.min(), y.max())
```
bbengfort: Awesome! We should probably add to the docstring the learned parameters of `self.range_` and `self.classes_` and when are on each based on the `_target_color_type`.

yellowbrick/features/base.py
```diff
@@ -215,13 +225,58 @@ def fit(self, X, y=None, **kwargs):
...
+            # Ensures that classes passed by user is equal to that in target
+#            if len(self.classes_)!=len(np.unique(y)):
+#                raise YellowbrickValueError("Number of unique target is not " 
+#                                            "equal to classes")
+            color_values = resolve_colors(n_colors=len(self.classes_), 
+                                          colormap=self.colormap, colors=self.color)
+            self._colors = dict(zip(self.classes_, color_values))
+            
+        # Compute target range if colors are continuous
+        elif self._target_color_type == CONTINUOUS:
```
bbengfort: We should probably also put a:

```python
else:
    raise YellowbrickTypeError("unknown target color type '{}'".format(self._target_color_type))
```

Just for future proofing. Plus I don't really like `if`/`elif` without an `else`, it seems like a code smell to me.

yellowbrick/features/base.py
bbengfort: We may also want to do a `self._colors` here as well and get the colormap for the scatterplot - does that make sense?

tests/test_features/test_base.py
```diff
@@ -47,3 +48,57 @@ def test_subclass(self):
...
+    
+    def test_determine_target_color_type(self):
```
bbengfort: I realize that the manifold tests had this as one long test; but now that we've broken it out into independent functionality, perhaps we could turn this into more tests:

1. Test single (both auto and specified)
2. Test discrete (both auto and specified)
3. Test continuous (both auto and specified) 
4. Test bad target types raise an exception 

That would allow us to test other things as well such as `classes_`, `range_` and `_colors` and be a bit more thorough with the testing process (e.g. use `make_classification` and `make_regression` instead of `np.random`).

yellowbrick/features/base.py
naresh-bachwani: I can add it but it will ultimately be duplicate for the exception in line 278.

yellowbrick/features/base.py
naresh-bachwani: Well, this was the functionality I added. It ensures that the number of classes is equal to the number of unique `target`. But it gave an error in case of ParallelCoord because it samples the target. Pcoords explicitly finds classes hence classes are equal to the unique unsampled target. Due to this, the number of classes becomes more than unique sampled y. So I seek suggestions from you on this. Should I keep or remove this? This may need updating pcoords.

tests/test_features/test_base.py
naresh-bachwani: My bad! Will correct it.

yellowbrick/features/base.py
naresh-bachwani: I have read the doc. I will try implementing it. Or do you wish to explore it together?

yellowbrick/features/base.py
naresh-bachwani: Do you mean something like `self._colors=colormap`? I plan to calculate the color based on target and colormap. Then we can pass it back as `self._colors`. Can we add this functionality(to handle ranges) to `resolve_colors`!

yellowbrick/features/base.py
naresh-bachwani: Will do it.

tests/test_features/test_base.py
naresh-bachwani: WIll do it.

yellowbrick/features/base.py
bbengfort: Go ahead and take a crack at it!

yellowbrick/features/base.py
bbengfort: Yeh, that's true - but it will be a little backup in case something goes really weird.

yellowbrick/features/base.py
bbengfort: Yes, though the user is likely to have supplied the colormap as a string, so you would convert it to an actual colormap and normalizer via the range, e.g. 

```
self._colors = get_cmap(self.cmap)
self._norm = mpl.colors.Normalize(vmin=self.range_[0], vmax=self.range_[1])
self.cbar = mpl.colorbar.ColorbarBase(self.cax, cmap=self._colors, norm=norm)
```

This has a lot of implications though, so maybe we should hold off until we get to the scatter prototype.

naresh-bachwani: I already updated my code with this functionality. But should we find cbar here? If yes then probably we should find legends too. Or we can keep these both to be handled by the subclass of `DataViz`

yellowbrick/features/base.py
bbengfort: Was just looking at parallel coordinates, which looks like it may need some updates as a result of this PR; can you actually list all of the subclasses of `DataVisualizer` for me so that we can make sure that things work as expected? For example, we'll have to ensure that we do:

```
super(ParallelCoordinates, self).__init__(ax, target=DISCRETE) 
```

In `ParallelCoordinates` and `RadViz` since there isn't really a continuous target identification yet; and for those that do, we'll have to add the params of the `DataVisualizer` there. 

As for this functionality, I think doing a verification that the classes supplied by the user match whats in `y` is a good thing; perhaps instead of an exception, we could simply issue a warning?

yellowbrick/features/base.py
bbengfort: Yeh, we can leave off the `cbar` until `finalize`.

yellowbrick/features/base.py
naresh-bachwani: I updated with something like this:
```python
norm = matplotlib.colors.Normalize(self.range_[0], self.range_[1])
cmap = matplotlib.cm.get_cmap(self.colormap)
color_values = cmap([norm(yi) for yi in y])
self._colors = color_values
```
It calculates color for each point. So should I let it be like this or update to return just colormap.

yellowbrick/features/base.py
naresh-bachwani: I updated with something like this:
```python
norm = matplotlib.colors.Normalize(self.range_[0], self.range_[1])
cmap = matplotlib.cm.get_cmap(self.colormap)
color_values = cmap([norm(yi) for yi in y])
self._colors = color_values
```
It calculates color for each point. So should I let it be like this or update to return just colormap.

bbengfort: Hmm, I would absolutely want to compute the color values with norm the way you did instead of leaving it to matplotlib, but I would want to do it in draw and not in fit and I wouldn't want to store the color values on the visualizer. Go ahead and just make `self._colors` the cmap for now and we'll use `self.range_` to compute the colorbar etc. in `draw` in `finalize`.

yellowbrick/features/base.py
naresh-bachwani: Following are the subclasses for DataVisualizer:

1. ParallelCoordinates
2. RadViz
3. contrib/missing/MissingDataVisualizer(It has only single target _type)
4. contrib/scatter(only discrete)

yellowbrick/contrib/missing/dispersion.py
```diff
@@ -69,7 +69,8 @@ class MissingValuesDispersion(MissingDataVisualizer):
     >>> visualizer.poof()
     """
 
-    def __init__(self, alpha=0.5, marker="|", classes=None, **kwargs):
+    def __init__(self, alpha=0.5, marker="|", classes=None, target="discrete",
```
bbengfort: See above comment RE `MissingValuesBar`!

yellowbrick/features/base.py
```diff
@@ -124,6 +130,12 @@ def fit(self, X, y=None, **fit_params):
 ## Data Visualizers
 ##########################################################################
 
+class target_type(Enum):
+    AUTO = 'auto'
+    SINGLE = 'single'
+    DISCRETE = 'discrete'
+    CONTINUOUS = 'continuous'
```
bbengfort: So do you like the `Enum`? What happens when you do `target_type("foo")`" From the below code it appears that some exception is raised; do you know what kind? 

Also, this should probably be `class TargetType(enum)` - so that we can do `TargetType.AUTO`.

yellowbrick/features/base.py
```diff
@@ -165,6 +177,16 @@ class DataVisualizer(MultiFeatureVisualizer):
         optional string or matplotlib cmap to colorize lines
         Use either color to colorize the lines on a per class basis or
         colormap to color them on a continuous scale.
+        
+    target : str, default: "auto"
```
bbengfort: Should we change this argument to `target_type` -- do you think that would be more understandable?

yellowbrick/features/base.py
```diff
@@ -190,7 +212,11 @@ def __init__(self, ax=None, features=None, classes=None, color=None,
         # Visual Parameters
         self.color = color
         self.colormap = colormap
-
+        try:
+            # Ensures that taregt is either Single, Discrete, Continuous or Auto
+            self.target = target_type(target)
+        except:
```
bbengfort: What kind of exception does this raise? 

Also what happens when you do `target_type(target_type.AUTO)`?

bbengfort: Also, small typo in the comment ;-)

yellowbrick/features/pcoords.py
```diff
@@ -250,7 +250,7 @@ def __init__(self,
                  **kwargs):
 
         super(ParallelCoordinates, self).__init__(
-            ax, features, classes, color, colormap, **kwargs
+            ax, features, classes, color, colormap, target="discrete", **kwargs
```
bbengfort: See above comment about `MissingValuesBar`.

yellowbrick/features/radviz.py
```diff
@@ -147,7 +147,7 @@ class RadialVisualizer(DataVisualizer):
     """
 
     def __init__(self, ax=None, features=None, classes=None, color=None,
-                 colormap=None, alpha=1.0, **kwargs):
+                 colormap=None, alpha=1.0, target="discrete", **kwargs):
```
bbengfort: See above comment about `MissingValuesBar`

yellowbrick/features/base.py
bbengfort: Ok - that's actually great and not too many of them. Thanks for compiling that list! Right now none of these care about the target type, or they expect discrete.

yellowbrick/features/radviz.py
bbengfort: Although this should default to `discrete` not `single`.

yellowbrick/features/pcoords.py
bbengfort: Although this should default to `discrete` not `single`.

tests/test_features/test_base.py
```diff
@@ -17,17 +17,51 @@
...
+def continuous(request):
+    """
+    Creates a random regressor fixture. 
+    """
+    X, y = make_regression(
+        n_samples=500, n_features=22, n_informative=8, random_state=2019
+    )
+
+    # Set a class attribute for digits
+    request.cls.continuous = TestDataset(X, y)
```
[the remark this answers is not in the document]
naresh-bachwani: So I will leave it as it is in this PR and we can pick move it to `conftest.py` in a new PR.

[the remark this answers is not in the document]
bbengfort: Yep, thank you! Also here is the note from pytest on fixture discovery:

https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions

yellowbrick/features/base.py
naresh-bachwani: Yup, the enums are a great way to refer to a group of constants having the same functionality. `target_type("foo")` raises `ValueError` stating that `"foo"` is not a part of the class.

I will change the name to TargetType.

yellowbrick/features/base.py
naresh-bachwani: It raises`ValueError`. 

`target_type(target_type.AUTO)`  and 'target_type("AUTO")` have same results.

I will correct the typo.

bbengfort: per your above comment, would you mind making this `except ValueError:`?

bbengfort: Awesome - if so, perhaps we should have the default be `TargetType.AUTO` instead of the string?

naresh-bachwani: We certainly can do so. But I think that it may confuse the user because a string type variable would have defaulted to some Enum. We can go with it if such examples are common.

yellowbrick/features/base.py
naresh-bachwani: Well, this is a matter of choice and conventions. Perhaps, you will know better about this. Just let me know and I will change this.

yellowbrick/features/base.py
bbengfort: It doesn't really matter; let's leave it as a string for now.

yellowbrick/features/base.py
bbengfort: I'm leaning toward `target_type` -- @rebeccabilbro want to way in? @DistrictDataLabs/team-oz-maintainers?

naresh-bachwani:
I have made the suggested changes. Now we should just wait for the `target` or `target_type` thing.

yellowbrick/features/base.py
rebeccabilbro: I think `target_type` would be much more intuitive to our users (who will probably expect `target == y`)!

yellowbrick/features/base.py
```diff
@@ -124,6 +130,12 @@ def fit(self, X, y=None, **fit_params):
 ## Data Visualizers
 ##########################################################################
 
+class Target_Type(Enum):
```
bbengfort: Minor typo, `Target_Type` should be `TargetType`, variable names are:

- all caps snake case for constants
- camel case for classes 
- all lower snake case for functions, methods, and other variables

bbengfort:
@naresh-bachwani sorry - made some changes during the merge on my end; so I overrode your merge -- I canceled the duplicate builds and as soon as the tests pass we'll merge!

PR #898, opened 2019-07-01
Avoid updating Travis conda package manager; issue with conda 4.7.5 dependency graph

nickpowersys:
### Travis Miniconda 3.7 affected

While working locally with Anaconda 3.7 on Linux with the latest conda version, 4.7.5, I found that the conda solver ran into this issue with conflicting dependencies, which prevents a solution:
https://github.com/conda/conda/issues/8844

The Travis installation script has been updating the base conda version before installing dependencies.

After I updated .travis.yml on Travis under this PR such that conda is not updated, the version that has been running on Travis is 4.6.14.

### Windows is not affected by the issue

I also investigated Miniconda on Windows, and there is apparently not the same problem: a full set of dependencies was determined and installed by conda 4.7.5 for both Miniconda 3.6 and 3.7.

On Appveyor Miniconda 3.7, although conda 4.7.5 is able to solve the dependency graph, there is still an issue with the Appveyor 3.7 configuration such that the build remains inactive.


nickpowersys:
The only commit that actually belongs in this should be the last, for .travis.yml: a66a739.

Somehow my local branch still reflects previous commits in the PR. I will go over the last section of  Contributing.md to see if I can resolve that.

nickpowersys:
Closing this as the source branch need to be changed and this cannot be done through the GUI.

PR #899, opened 2019-07-01
Skip conda update to avoid broken Travis builds; issue with conda 4.7.5 dependency graph

nickpowersys:
### Avoid updating conda package manager on Travis, due to issue observed on Travis Miniconda 3.7

While working locally with Anaconda 3.7 on Linux with the latest conda version, 4.7.5, I found that the conda solver ran into this issue with conflicting dependencies, which prevents a solution:
https://github.com/conda/conda/issues/8844

The Travis installation script has been updating the base conda version before installing dependencies.

After I updated .travis.yml on Travis under this PR such that conda is not updated, the version that has been running on Travis is 4.6.14.

### Windows is not affected by the issue

I also investigated Miniconda on Windows, and there is apparently not the same problem: a full set of dependencies was determined and installed by conda 4.7.5 for both Miniconda 3.6 and 3.7.

On Appveyor Miniconda 3.7, although conda 4.7.5 is able to solve the dependency graph, there is still an issue with the Appveyor 3.7 configuration such that the build remains inactive.


nickpowersys:
*I really only want a66a739 in the PR*, for the deletion of a single line in .travis.yml. I am not able to stop the other commits from recurring.

nickpowersys:
This would be preventive in that Travis Miniconda 3.7 builds are not breaking yet with the existing .travis.yml, even though the conda update step updates to 4.7.5.

nickpowersys:
I identified the problem, which was that I specified a minimum version of pip to be used with a conda environment (19.1.1) and it broke compatibility.

Issue #826, opened 2019-04-24
API Discussion for Figures and Axes


bbengfort:
This issue is to discuss the open letter regarding Yellowbrick's API roadmap. To summarize, we currently attempt to only manage matplotlib `Axes` objects so that visualizers can be embedded into more complex plots and reports. However, many of our visualizers are getting increasingly complex, requiring subplots of their own. For a complete discussion of the API issue please see:

https://www.scikit-yb.org/en/develop/api/figures.html

The questions at hand are:

1. Like Seaborn, should YB have two classes of visualizer, one that wraps an axes and one that wraps a figure?
2. Should we go all in on the AxesGrid toolkit and continue to restrict our use of the figure, will this method be supported in the long run?

tacaswell:
This is going to be a bit of a rambling answer (sorry).

This is a sticky API problem (that gets worse if you want to think about how to return "artists" that span multiple `Axes` (I want to change the color of / data in the residuals plot, I need one to be able to grab all of the parts of it) ).

> Generally, visualizers should behave as though they are a plot that as part of a larger figure.

:100: agree with this.


AxesGrid is not going anywhere and there is currently on going work to document and harden that part of the code base.

Riffing a bit, it sounds like the information that should go into the visualizers is "this area on this figure is for your use" which is currently being expressed as an `Axes` object that gets laundered through `make_axes_locatable`.  Taking in a `GridSpec` or and `Axes` and internally shimming them to be the same is probably a reasonable API (and also allow `Figure` which lets it assume it owns the whole thing).

Another axis of composition that maybe of interest is to put multiple models on the same (set of) axes.  In that case it may be worth having a compatibility layer with an api that is half dictionary / half GridSpec.  Something like

```python

def viz_method(data, fit, **style, axes_source):
    # make named Axes objects where we want them
    if not len(axes_source):
        # The position would be relative to the space given to the axes_source
        axes_source.add('data', (.1, .2, .8, .7))  # left, bottom, width, height
        # want to share the xaxis of data and residual to keep synched
        axes_source.add('residual', (.1, .1, .8, .1), share_x='data')
    # or rely on them existing
    else:
         pass
    # do the plotting
    d = axes_source['data'].plot(data, **style)
    f = axes_source['data'].plot(fit, **{**style, alpha=.5})
    r = axes_source['residual'].plot(fit - data, **style)

    # return the artists in a structured way
    return {'data': d, 'fit': f, 'residual': r}

```
[...]

PR #908, opened 2019-07-04
[WIP] Addressing issue 874: Added Projection Visualizer

naresh-bachwani:
This PR handles #874 . I updated the `tests/features/base.py`, which now contains ProjectionVisualizer. It unifies the functionality of PCA and Manifold. It contains the following features: 
1. It handles both discrete and continuous variables.
2. It handles both 2d and 3d plots.
3. It manually draws colorbar for 2d plots. 

### TODOs and questions

- [ ] Update to handle more features(like LabelEncoder)
- [x] Add docstring.
- [x] Create tests


yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+        self._cax = None
+
+    @property
+    def cax(self):
+        """
+        The axes of the colorbar, right of the scatterplot.
+        """
+        if self._cax is None:
+            raise AttributeError(
+                "this visualizer does not have a histogram for the X axis"
```
bbengfort: The attribute error message needs to be updated.

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+class ProjectionVisualizer(DataVisualizer):
+    
+    def __init__(self, transformer, ax=None, features=None, classes=None, color=None,
+             colormap=None, target_type="auto", proj_dim=2, alpha=0.75, **kwargs):
+        super(ProjectionVisualizer, self).__init__(ax=ax, features=features, 
+                                                     classes=classes, color=color,
+                                                     colormap=colormap, 
+                                                     target_type=target_type, **kwargs)
+        self.transformer=transformer
+        if proj_dim not in [2, 3]:
```
bbengfort: I don't actually like the argument `proj_dim` -- could we change it to just `projection`? Also I think I'd prefer the following options to be valid `{2, 3, "2D", "3D"}` (and note the use of the set for containment testing).

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
                 "could not determine target color type "
                 "from target='{}' to '{}'"
             ).format(self.target_type, self._target_color_type))
+
+
+##########################################################################
+## Projection Visualizers
+##########################################################################
+
+class ProjectionVisualizer(DataVisualizer):
```
bbengfort: Perhaps we should put this in `yellowbrick/features/projection.py`? `base.py` is getting a little full at this point!

bbengfort: Also to wrap this up, I'd like to make sure we have a complete docstring that describes what this base class does for other yellowbrick developers.

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+    
+    def __init__(self, transformer, ax=None, features=None, classes=None, color=None,
+             colormap=None, target_type="auto", proj_dim=2, alpha=0.75, **kwargs):
+        super(ProjectionVisualizer, self).__init__(ax=ax, features=features, 
+                                                     classes=classes, color=color,
+                                                     colormap=colormap, 
+                                                     target_type=target_type, **kwargs)
+        self.transformer=transformer
```
bbengfort: Instead of having base classes pass a transformer to `super` (which is a decent idea), I'd prefer the projection visualizer to assume that it is a transformer (e.g. implements the `fit()` and `transform()` methods. This will allow different subclasses to instantiate their transformers whenever they see fit and not just in `__init__` (e.g. they could do it in `fit()`. Manifold and PCA create their transformers very differently so I want to make sure that we can capture all possible use cases.

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+        if proj_dim not in [2, 3]:
+            raise YellowbrickValueError("Projection dimensions must be either 2 or 3")
+        self.proj_dim = proj_dim
+        self.alpha = alpha
+        self._cax = None
+
+    @property
+    def cax(self):
```
bbengfort: So we may have to think in more detail about how to best organize all of the related axes in the visualizer. I do like what you've done here, which is to specify the colorbar axes as a property but raise an exception if the visualizer does not have a colorbar. The question is do we do this for all possible axes in the visualizer, or do we leave it for the layout mechanism of the subclasses?

All projection visualizers will have:

1. A scatter plot 
2. An optional colorbar in the case of `_target_type == "CONTINUOUS"`
3. In the 3D case, an Axes3D object that replaces gca 

PCA will additionally have the heatmap, JointPlot will have the X and Y histograms. 

Perhaps we should have a `layout()` method that is called by `draw()` and can be overridden by subclasses? I like that you've added `layout_colorbar` - but this would make things a bit more general.

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+        """
+        Creates the layout for colorbar. The colorbar is added to the right of the 
+        scatterplot 
+        """
+        # Ensure matplotlib version compatibility
+        if make_axes_locatable is None:
+            raise YellowbrickValueError((
+                "heatmap requires matplotlib 2.0.2 or greater "
+                "please upgrade matplotlib or set heatmap=False on the visualizer"
+            ))
```
bbengfort: Please update the text of this exception.

bbengfort: Although, see https://github.com/DistrictDataLabs/yellowbrick/issues/826#issuecomment-509249546 -- we may just want to make matplotlib v2.0.2 the minimum version and we can take out these warnings.

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+
+        # Create the new axes for the colorbar
+        divider = make_axes_locatable(self.ax)
+        self._cax = divider.append_axes("right", size="5%", pad=0.3)
+        
+    def fit(self, X, y=None, **kwargs):
+        
+        self.fit_transform(X, y, **kwargs)
```
bbengfort: Is there a reason to call `fit_transform` here rather than just fitting the transformer? Also, I think we need a call to super to do target type identification, get the classes and feature names, etc. And docstrings for all the methods here!

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+        return self
+        
+    def fit_transform(self, X, y=None, **kwargs):
+
+        # Transformer will be  defined either in the subclasses or explicitly
+        self.transformed_features_ = self.transformer.fit_transform(X)
```
bbengfort: We don't want to save data on the visualizer wherever we can avoid it - imagine that this is a really large text corpus, we could easily blow up the memory of a users machine if we save data. Instead, let's just pass the transformed data to `draw()`. 

Also the `super` and the `fit()` should come before transform here.

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+        super(ProjectionVisualizer, self).fit(X, y, **kwargs)
+        return self.transformed_features_
```
bbengfort: +1 - the transform methods should always return the transformed data!

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+        # Transformer will be  defined either in the subclasses or explicitly
+        self.transformed_features_ = self.transformer.fit_transform(X)
+        super(ProjectionVisualizer, self).fit(X, y, **kwargs)
+        return self.transformed_features_    
+        
+    def transform(self, X, y=None, **kwargs):
+        try:
+            return self.transformer.transform(X)
+        except AttributeError as e:
+            raise AttributeError(str(e) + " try using fit_transform instead.")
```
bbengfort: I'd say that `transform()` is the entry point for draw. We should assume that `transform()` can be called multiple times on multiple data, just adding new scatter plot points every time. So here let's pass `X_prime` (the transformed `X` which has either 2 or 3 columns) and `y` to `draw()`

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+    def finalize(self):
+        
+        if self._target_color_type == TargetType.DISCRETE:
+            # Add the legend
+            manual_legend(self, self.classes_, list(self._colors.values()),
+                          frameon=True)
+
+        elif self._target_color_type == TargetType.CONTINUOUS:
+            if(self.proj_dim==3):
+                self.fig.colorbar(self._scatter, ax=self.ax)
```
bbengfort: We can't rely on there being a `self.fig` on the visualizer. In this code because that only happens if `proj_dim` == 3;  but generally since we won't have a `figure` object to work with.

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+    # Not sure about the function signature      
+    def draw(self, X=None, y=None):
+        X = self.transformed_features_
+        scatter_kwargs = self._determine_scatter_kwargs(y);
+        
+        if self.proj_dim == 2:
+            self._scatter = self.ax.scatter(X[:,0], X[:,1], **scatter_kwargs)
+
+        if self.proj_dim == 3:
+            self.fig = plt.figure()
```
bbengfort: Rather than creating a new figure, I'd prefer if we used `plt.gcf` (get current figure) - and also I don't think we should store the fig on the visualizer as discussed in #826.

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+            self.ax = self.fig.add_subplot(111, projection='3d')
```
bbengfort: What about the case where the user has passed in an `Axes` object? Let's refactor this to override the `Visualizer.ax` property and do appropriate checks as needed.

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+            if(self.proj_dim==3):
+                self.fig.colorbar(self._scatter, ax=self.ax)
+            
+            else:
+                self._layout_colorbar()
+                # Manually draw the colorbar.
+                self.cbar = mpl.colorbar.ColorbarBase(self.cax, cmap=self._colors, 
+                                                  norm=self._norm)
+
+    def _determine_scatter_kwargs(self, y=None):
```
bbengfort: Nice! Good way to encapsulate this functionality!

yellowbrick/features/base.py
```diff
@@ -309,3 +312,130 @@ def _determine_target_color_type(self, y):
...
+            return self.transformer.transform(X)
+        except AttributeError as e:
+            raise AttributeError(str(e) + " try using fit_transform instead.")
+
+    # Not sure about the function signature      
+    def draw(self, X=None, y=None):
+        X = self.transformed_features_
+        scatter_kwargs = self._determine_scatter_kwargs(y);
+        
+        if self.proj_dim == 2:
```
bbengfort: Should we divide this up into `_draw_2d` and `_draw_3d` methods to make the organization a bit easier?

yellowbrick/features/base.py
naresh-bachwani: Will do so!

yellowbrick/features/base.py
naresh-bachwani: I will write the docstring and move it to `projection.py`!

yellowbrick/features/base.py
naresh-bachwani: I completely understand your point related to Manifold! Initially, I had the same functionality as you mention here and wanted to discuss it.  
I think that having this parameter allows `ProjectionVisualizer` to be able to project any transformed data in 2D or 3D. By using this, It will act as a different Visualizer that projects when the transformer is given. Although here we can use something like `Visualizer.transformer=PCA`, having a parameter will seem more intuitive. What do you think about this? Let me know your thoughts. I know that this may cause problems with Manifold!

yellowbrick/features/base.py
naresh-bachwani: I will remove this!

yellowbrick/features/base.py
naresh-bachwani: I just need to change the name to `layout`, right? Just to be sure, in PCA we will have two colorbar and a heatmap for the continuous case, right?

yellowbrick/features/base.py
naresh-bachwani: The super `fit` calls draw(). So if I place the super `fit` before `transform` then It will call draw without having transformed data.
This is also one of the reasons I saved the data on the visualizer. The problem that arises here is that call to `draw()` is made by the `super` class. So, either we somehow pass the data to super class or save it on visualizer. Is there any other method, that you think will help us?

yellowbrick/features/base.py
naresh-bachwani: A problem here! The super class calls draw in `fit` method so I don't think that we can call this here. Let me know your thoughts!

yellowbrick/features/base.py
bbengfort: Having the transformer as an attribute assumes that the Visualizer _wraps_ a transformer, not that it _is a_ transformer; it's a subtle difference I know. Basically what I'm saying here is that the `ProjectionVisualizer` should just assume that the `transform()` method will correctly project the original data into either two or three dimensions as needed. In fact, it is likely that the `ProjectionVisualizer` does not need `fit`, `transform` or `fit_transform` unless there is a special reason to do so that I don't see quite yet.

yellowbrick/features/base.py
bbengfort: Yep, changing the name to `layout` is a good start! Why two colorbars in PCA? One for the data points and one for the heatmap?

naresh-bachwani: Yup!

yellowbrick/features/base.py
bbengfort: As I'm in the airport I'm having trouble with my Internet connection and can't easily check the source code - is there already a `projection.py`? (I'm hoping there is not).

yellowbrick/features/base.py
bbengfort: Before you reset the minimum matplotlib version, let's get @rebeccabilbro and @DistrictDataLabs/team-oz-maintainers to chime in on the major version deprecation. Thoughts, guys?

yellowbrick/features/base.py
naresh-bachwani: No, currently we do not have projection.py!

yellowbrick/features/base.py
bbengfort: Ok, so an estimator must be `fit()` before it can either `transform()` or `predict()` -- this is a pretty critical part of the sklearn workflow. It looks like we might have to remove the call to draw in the `DataVisualizer` -- this was just there as a convenience anyway. Would you mind looking and seeing what this will affect -- e.g. what subclasses of `DataVisualizer` rely on this call to `draw`? It should be pretty straight forward to fix:

```python
class MyVisualizer(DataVisualizer):

    def fit(self, X, y=None):
        super(MyVisualizer, self).fit(X, y)
        self.draw(X, y)
        return self
```

bbengfort: You know, this might be why `Manifold` didn't extend `DataVisualizer` ... so this will be a good thing to fix!

yellowbrick/features/base.py
bbengfort: ok awesome, I'm now thinking about the following organization:

```
yellowbrick
└── features
|   ├── __init__.py
|   └── projection
|   |   ├── __init__.py
|   |   ├── base.py
|   |   ├── manifold.py
|   |   ├── pca.py
|   ├── base.py
|   ├── decomposition.py
```

But let me think about this a little more.

yellowbrick/features/base.py
rebeccabilbro: I'm at peace with this; it'll make #826 a heck of a lot easier to move forward with.

yellowbrick/features/base.py
naresh-bachwani: We can do this, but I think that the method is quite small and understandable.

yellowbrick/features/base.py
naresh-bachwani: Hey @bbengfort, 
Can you elaborate on what is needed here? I understand that when user passes `Axes` we get axes lines on the plot. To deal with that we just need to clear the figure and redefine the `Axes`. Is that what you mean here?

yellowbrick/contrib/missing/base.py
```diff
@@ -60,8 +60,8 @@ def fit(self, X, y=None, **kwargs):
 
         self.y = y
 
-        super(MissingDataVisualizer, self).fit(X, y, **kwargs)
```
bbengfort: We probably do still want the call to `super` right? To ensure that the target type is identified and that the class and feature names are collected?

tests/test_features/test_base.py
```diff
@@ -140,8 +135,7 @@ def test_bad_target(self):
         with pytest.raises(YellowbrickValueError, match=msg):
             DataVisualizer(target_type="foo")
 
-    @patch.object(DataVisualizer, 'draw')
-    def test_classes(self, mock_draw):
```
bbengfort: Thanks for cleaning this up!

yellowbrick/features/base.py
```diff
@@ -276,10 +276,8 @@ def fit(self, X, y=None, **kwargs):
             raise YellowbrickValueError(
                     "unknown target color type '{}'".format(self._target_color_type))
 
-        # Draw the instances
-        self.draw(X, y, **kwargs)
-
         # Fit always returns self.
+        # NOTE: cannot call draw in fit to support data transformers
```
bbengfort: 👍

yellowbrick/features/pcoords.py
```diff
@@ -371,8 +371,9 @@ def fit(self, X, y=None, **kwargs):
         if self.normalize is not None:
             X = self.NORMALIZERS[self.normalize].fit_transform(X)
 
-        # the super method calls draw and returns self
-        return super(ParallelCoordinates, self).fit(X, y, **kwargs)
```
bbengfort: Although we're not returning `super` we probably do still want the call to super right?

yellowbrick/features/projection.py
```diff
@@ -0,0 +1,269 @@
...
+        # Ensure matplotlib version compatibility
+        if make_axes_locatable is None:
+            raise YellowbrickValueError((
+                "heatmap requires matplotlib 2.0.2 or greater "
+                "please upgrade matplotlib or set heatmap=False on the visualizer"
+            ))
+
+        # Create the new axes for the colorbar
+        divider = make_axes_locatable(self.ax)
+        self._cax = divider.append_axes("right", size="5%", pad=0.3)
```
bbengfort: I'm sure you're still working on it - but just wanted to make a note so we didn't miss it - the `cax` is only required if `_target_type == CONTINUOUS` and we also want to make sure that no `_cax` is already there (e.g. from a different layout call).

yellowbrick/features/pcoords.py
naresh-bachwani: Call to super was basically a duplicate code. So I just removed it. It was used only to call draw.

bbengfort: Ok, I see now that I've expanded that. In this case - though it is out of scope for this PR - we should have the super-classes handle this work and not ParallelCoordinates. Let's create an issue for this an handle it in a follow-up PR.

yellowbrick/features/base.py
bbengfort: So a couple of things here, maybe I'm misunderstanding - but there is a difference between `Axes` and `axis spines` (see [anatomy of a matplotlib plot](https://www.scikit-yb.org/en/latest/matplotlib.html#figures-and-axes)). The `Axes` has nothing to do with lines on the plot, it is a part of the figure that is drawn upon. 

Consider the following use case:

```python
fig, axes = plt.subplots(ncols=2, projection='3d', figsize=(18,6))

Manifold(ax=axes[0], projection='3d').fit_transform(X, y)
PCA(ax=axes[1], projection='3d').fit_transform(X, y)
```

In this code, the figure is composed of two subplots represented by two Axes objects. The user is asking the Manifold visualization to be drawn on the left axes and the PCA visualization to be drawn on the right axes. 

Because of this we can't just create a figure or an axes on our own - we have to respect what the user passes in if they do, and only create the axes if nothing has been supplied. 

Does that help clear up the confusion?

yellowbrick/features/base.py
bbengfort: Ok, in that case let's leave it as it is!

yellowbrick/features/base.py
naresh-bachwani: Ok! Great! But what if the user passes non-3d axes? We can't draw 3d plots on that! Also, I could not find methods to convert 2d to 3d axes. Will we raise an error in this case?

bbengfort: This is a very good point - we should see what happens! If matplotlib raises a sensible error, then we'll keep that error, or wrap it in a YellowbrickError. Otherwise matplotlib might just plot it in 2D, I'm not sure - so let's find out!

yellowbrick/features/projection.py
```diff
@@ -0,0 +1,303 @@
...
+            if projection in {'2D', '2d'}:
+                projection = 2
+            if projection in {'3D', '3d'}:
+                projection = 3
+        if projection not in {2, 3}:
+            raise YellowbrickValueError("Projection dimensions must be either 2 or 3")
+        self.projection = projection
+
+        if self.ax.name!='3d' and self.projection == 3:
+                warnings.warn("Uses third feature space as size. Pass 3d axes.",
```
bbengfort: I like very much putting the warning in `__init__` and I'm glad you found a way to test if the axes was 3D - I had no idea that axes had a name property! 

```suggestion
                warnings.warn("data projection to 3 dimensions requires a 3d axes to draw on.", 
```

yellowbrick/features/projection.py
bbengfort: I've been thinking about this, and I think we should make it `layout()` instead of `_layout()` so that users can organize the axes as required.

yellowbrick/features/projection.py
```diff
@@ -0,0 +1,303 @@
...
+            ))
+
+        if self._cax is not None or self._target_color_type != TargetType.CONTINUOUS:
+            self.ax
+            return
+        # Create the new axes for the colorbar
+        divider = make_axes_locatable(self.ax)
+        self._cax = divider.append_axes("right", size="5%", pad=0.3)
+        self._cax.set_yticks([])
+        self._cax.set_xticks([])
```
bbengfort: Why are we taking the ticks off the colorbar axes?

yellowbrick/features/projection.py
```diff
@@ -0,0 +1,303 @@
...
+
+        Returns
+        -------
+        Xprime : array-like of shape (n, 2)
+            Returns the 2-dimensional embedding of the instances.
+
+        """
+        return self.fit(X, y, **kwargs).transform(X, y, **kwargs)
+    
+    def draw(self, X, y=None):
```
bbengfort: Could we please change the name of `X` to `Xp` to reflect that the transformed data is passed to draw?

bbengfort:
@naresh-bachwani when you've made the changes and are ready - let me know and I'll do the final run through of the code with the test images, etc.

tests/test_features/test_projection.py
```diff
@@ -0,0 +1,209 @@
...
+        visualizer.transformer=self.transformer
+        msg = "Target needs to be label encoded."
+        with pytest.raises(YellowbrickValueError, match = msg):
+            visualizer.fit_transform(X, y)        
+        
+    def test_y_required_for_discrete_and_continuous(self):
+        X, y = self.discrete
+        visualizer = ProjectionVisualizer()
+        visualizer.transformer=self.transformer
+        visualizer.fit(X, y)
```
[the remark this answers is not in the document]
naresh-bachwani: It is the case when someone passes y in `fit()` but not in `transform()`. And as transform calls draw, which calculates scatter kwargs for discrete target it raises an error.

tests/test_features/test_projection.py
```diff
@@ -0,0 +1,209 @@
...
+        n_samples=500, n_features=22, n_informative=8, random_state=2019
+    )
+
+    # Set a class attribute for continuous data
+    request.cls.continuous = Dataset(X, y)
+
+@pytest.fixture(scope="function")
+def transformer(request):
+    """
+    Creates a PCA transformer. Acccepts the projection dimnesion as param and
```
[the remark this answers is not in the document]
naresh-bachwani: Sorry for the inconvenience. I will keep it in mind.
This function returns `none()` for a specific reason. This fixture has a scope 'function'. And I used it for the class. Which means that it is called for every function of the class. Some of the tests do not need this fixture and which is ultimately called and lead to error. Hence to handle this error I returned  `None`. 
But by your above suggestion, this will ultimately be removed.

tests/test_features/test_projection.py
```diff
@@ -0,0 +1,209 @@
...
+    returns transformer accordingly. If parame is not specified then None is returned.
```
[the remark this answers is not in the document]
naresh-bachwani: This will be removed.

tests/test_features/test_projection.py
```diff
@@ -0,0 +1,209 @@
...
+                                    ("pca", PCA(request.param, random_state=2019))])
+    # Set a class attribute for transformer
+    request.cls.transformer = pca_transformer
+
+
+@pytest.mark.usefixtures("discrete", "continuous", "transformer")
+class TestProjectionVisualizerBase(VisualTestCase):
+    
+    @pytest.mark.parametrize("transformer", [2], indirect=True)
+    def test_discrete_plot(self):
```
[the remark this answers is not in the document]
naresh-bachwani: Will do it!

tests/test_features/test_projection.py
```diff
@@ -0,0 +1,209 @@
...
+
+        X, y = self.discrete
+        classes = ["a", "b", "c", "d", "e"]
+        visualizer = ProjectionVisualizer(projection=2,
+                                          colormap="plasma", classes=classes)
+        visualizer.transformer=self.transformer
+        transform_array = visualizer.fit_transform(X, y)
```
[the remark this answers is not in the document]
naresh-bachwani: Gotcha! X_prime makes more sense. I will change it.

tests/test_features/test_projection.py
```diff
@@ -0,0 +1,209 @@
...
+            ("pca", PCA(n_components=15, random_state=1998)),
+            ("lle", LocallyLinearEmbedding(n_components=2, random_state=2019)),
+        ])
+
+        visualizer = ProjectionVisualizer(projection="2d")
+        visualizer.transformer = manifold
+        visualizer.fit_transform(X, y)
+        visualizer.finalize()
+        visualizer.cax.set_yticklabels([])
+        self.assert_images_similar(visualizer)
```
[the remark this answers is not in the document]
naresh-bachwani: Thats great! But I think that changing random state can fix it. Or maybe changing the transformer. Because it is working in one of test for continuous.

tests/test_features/test_projection.py
```diff
@@ -0,0 +1,209 @@
...
+        assert transform_array.shape == (self.discrete.X.shape[0], 2)
+        
+    def test_continuous_plot(self):
+
+        X, y = self.continuous
+        manifold = Pipeline([
+            ("pca", PCA(n_components=15, random_state=1998)),
+            ("lle", LocallyLinearEmbedding(n_components=2, random_state=2019)),
+        ])
```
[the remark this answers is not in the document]
naresh-bachwani: No. This was just to try it with just some other transformer. And ensure that it works with others too! And I think that this is the reason for the difference and changing it to transformer fixture will work fine.

tests/test_features/test_projection.py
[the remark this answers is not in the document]
naresh-bachwani: What do you think can be the reason of difference?

tests/test_features/test_projection.py
```diff
@@ -0,0 +1,209 @@
...
+
+        visualizer = ProjectionVisualizer(projection="3D")
+        visualizer.transformer = manifold
+        visualizer.fit_transform(X, y)
+        visualizer.finalize()
+        visualizer.cbar.set_ticks([])
+        self.assert_images_similar(visualizer)
+
+    @pytest.mark.parametrize("transformer", [2], indirect=True)
+    @mock.patch("yellowbrick.features.pca.plt.sca", autospec=True)
```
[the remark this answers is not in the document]
naresh-bachwani: I was unsure about this test and copied it from PCA. But I was going to update it. I will surely remove the unused features.

yellowbrick/features/projection.py
naresh-bachwani: I was getting ticks(extra) on the bottom and left of the axes. So I removed them.

yellowbrick/features/projection.py
naresh-bachwani: Sure!

naresh-bachwani:
Well I like your suggestion related to `fit()` and `transform()`. I will make the necessary changes and let you know!

yellowbrick/contrib/missing/base.py
naresh-bachwani: Yup! I think that dealing with this in #923 will make more sense.

tests/test_features/test_projection.py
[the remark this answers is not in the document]
bbengfort: I think it's just the windows renderer vs the gtk renderer - it doesn't seem to be a random state issue, but I'll keep looking for a place where there is randomness that we didn't fix.

tests/test_features/test_projection.py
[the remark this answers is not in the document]
bbengfort: Ah, I understand now - thanks!

yellowbrick/features/projection.py
bbengfort: Ok, I was worried this would move the scale on the legend, but it looks like it's fine:

Figure_1

I guess this is because we are manually creating the colorbar in `finalize()`; let's just make sure this remains true as we move to the other visualizers.

PR #914, opened 2019-07-06
Revert rank2d baseline images to pre 3.1.1 state

### LINKED ISSUE #912: Matplotlib 3.1.1 impacts some visual tests
Matplotlib 3.1.1 (released on July 1, 2019) broke some of our tests, particularly the Rank2D tests with very big RMSE (~40) and very different diff images. For now in in #907 I have regenerated the baseline images for Rank2D so that our tests will pass on Travis and Appveyor, but I have not changed the `tests/requirements.txt` (except to note this strange behavior). Running the tests with `matplotlib==3.1.0` will result in image comparison failures for Rank2D. We should investigate what the behavior is for other matplotlib versions and determine what has changed in 3.1.1 that changed the Rank2D images (but not, say, the Confusion Matrix or Classification Heatmap) so drastically.

rebeccabilbro:
This PR fixes #912, which identified a discrepancy with how Matplotlib was rendering the Rank2D images for our baseline test images. It has been determined that the plot-clipping is due to [a bug in 3.1.1 that will be fixed in 3.1.2](https://github.com/matplotlib/matplotlib/pull/14677). This PR updates our matplotlib `tests/requirements.txt` to avoid the 3.1.1 plot clipping bug, and reverts the Rankd baseline images to pre 3.1.1 state.

## Questions
 - Should we also update `requirements.txt` and `docs/requirements.txt`?


nickpowersys:
Would we want to also modify requirements.txt? While the scope is not large in terms of known affected tests, anyone plotting Rank2D as a user may not understand that the problem is not with yb and not know how to fix it with their package manager?

There could be a reason pro for keeping 3.1.1 that I am not aware of yet, such as a bug that it fixes that is in <3.1.1. However, we know at least that the images across non-3.1.1 releases are consistent with the exception of Rankd2D.

nickpowersys:
Oh, I realize this would only have an effect once the 1.0.goes out, so it could wait. Ha!

rebeccabilbro:
Good question @nickpowersys &mdash; @bbengfort want to weigh in? I know you've been thinking about multi-dependency support these days.

jklymak:
If yellowbrick were able to test off matplotlib master that would be a huge help for one of your test builds.  Of course matplotlib should have more comprehensive tests but if many downstream packages test our master, particularly ones that are doing lots of customization, it would be a huge help in catching bugs on use cases we didn’t anticipate.

nickpowersys:
A reason for not excluding 3.1.1 is that it could make it more difficult to achieve consistency across all other existing packages and package requirements someone has in their environment, not just for yb. Maybe another package excludes one or more non-3.1.1 versions, for example. The conda solver might not be able to solve when attempting to install yb.

bbengfort:
@nickpowersys I agree with your assessment, @rebeccabilbro decided on our advice to only update the docs/requirements.txt and tests/requirements.txt and to add a warning in RankD to users who are using mpl v 3.1.1 -- this way we don't create a more complicated dependency graph than we have to! 

@jklymak thanks for the suggestion, I've opened #918 to discuss it further!

PR #921, opened 2019-07-11
Enhance datasets documentation

bbengfort:
This PR fixes #441 - updating the docs with more detailed information about our datasets, and is one of the last steps of #683. Changes include:

1. Create individual pages for each dataset in the documentation
2. Extend the dataset documentation with a better API reference 
3. Add more detail information in tables about dataset contents 
4. Fix other documentation warnings and errors

### TODOs and questions

Still to do:

- [x] Enhance the datasets download and usage documentation
- [x] Finish all dataset detail pages 
- [x] Enhance the loading a text corpus page 
- [x] Add information about using datasets in tests for contributors

Questions for the @DistrictDataLabs/team-oz-maintainers:

- [x] How much detail do we want on each dataset page?
- [ ] ~Should we include examples or links to examples on each dataset page?~

### CHECKLIST

- [x] _Is the commit message formatted correctly?_
- [ ] _Have you noted the new functionality/bugfix in the release notes of the next release?_

- [ ] ~_Included a sample plot to visually illustrate your changes?_~
- [x] _Do all of your functions and methods have docstrings?_
- [ ] ~_Have you added/updated unit tests where appropriate?_~
- [ ] ~_Have you updated the baseline images if necessary?_~
- [x] _Have you run the unit tests using `pytest`?_
- [x] _Is your code style correct (are you using PEP8, pyflakes)?_
- [x] _Have you documented your new feature/functionality in the docs?_

- [x] _Have you built the docs using `make html`?_

PR #922, opened 2019-07-15
[WIP] Update model selection tutorial so figure windows do not block next visualizer in scripts

nickpowersys:
This relates to #924. Instead of implementing a plot directive due to the long build time, it involves updating an existing code directive for `visualize_model`.

The planned updates are

1. Address how, in `visualize_model`, a call to `poof()` for multiple visualizers requires significant manual work from the user such as saving each file manually. This would slow down model selection when using a .py script based on the code directive.

2. In the `score_model` code directive section
- imported LabelEncoder and OneHotEncoder, and all code needed to produce the F1 scores.
- The code can be copy-pasted and used in a notebook or script because it is fully self-contained.
- It remains a code directive section instead of a plot directive because it has `print` statements and a `parsed-literal` directive rather than plots.
- I used the 88-character black formatting line width for wrapping.

3. Although I pulled from the latest upstream develop branch, there were no files in the existing .png image paths referenced within the .rst file, so they must have been removed previously.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -111,14 +111,19 @@ diagnostics from the Yellowbrick library).
 
 .. code:: python
 
-    from sklearn.metrics import f1_score
     from sklearn.pipeline import Pipeline
+    from sklearn.preprocessing import OneHotEncoder, LabelEncoder
+    from sklearn.metrics import f1_score
     from sklearn.svm import LinearSVC, NuSVC, SVC
     from sklearn.neighbors import KNeighborsClassifier
-    from sklearn.preprocessing import OneHotEncoder, LabelEncoder
-    from sklearn.linear_model import LogisticRegressionCV, LogisticRegression, SGDClassifier
-    from sklearn.ensemble import BaggingClassifier, ExtraTreesClassifier, RandomForestClassifier
+    from sklearn.linear_model import (LogisticRegressionCV, LogisticRegression,
+                                      SGDClassifier)
+    from sklearn.ensemble import (BaggingClassifier, ExtraTreesClassifier,
+                                  RandomForestClassifier)
     
+    from yellowbrick.datasets import load_mushroom
+
+    X, y = load_mushroom()
         
     models = [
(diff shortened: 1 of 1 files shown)


lwgray:
@nickpowersys let's connect over Yellowbrick slack channel [@Larry Gray]

nickpowersys:
Before closing the PR, I think there may be overlapping of the plots with the current code. This was happening during `make html` when the line for subplots was not included as it is in the PR.

I could simply revise this to a code directive linked to the .png file, to ensure that someone could copy-paste either the score_model section or visualize_model section and run them to produce the results on the page shown now (and then revise them accordingly to suit their needs).

Of course, if this is a low priority, it can be closed and revisited if desired.

lwgray:
hi @nickpowersys Great catch! Could you post a screenshot of the suspected overlapping?

nickpowersys:
The overlapping occurs if there is a plot directive, unless there is the `_, ax = plt.subplots()` line. I have now tested the code in the existing code directive (which does not have that line) in a plain script and notebook as well.

With a script, a figure window is created, and waits to be closed before the next figure is shown. Each shows up as Figure 1. Closing each figure created by a non-notebook script means that figure is no longer available for comparison with the others, but manually saving each figure one at a time before closing it is extremely slow. Adding `_, ax = plt.subplots()` has no effect on this behavior. 

With a notebook, separate figures are automatically generated in the same output cell.

I could leave the `_, ax = plt.subplots()` line out and do the following to show how a script can be written to run smoothly and quickly.

- Specify the signature as:

`def visualize_model(X, y, estimator, in_notebook=True, **kwargs):`

- At the end of the function

```
# This function is either called in a notebook or a script
if in_notebook:
    # Display the image in the output cell
    visualizer.poof()
else:
    # Save the image to a file then close the figure
    visualizer.poof(estimator.__class__.__name__ + '.png')
    plt.close()
```

lwgray:
@nickpowersys Let's open a issue specifically for this phenomenon and wait until @bbengfort PR #921 is resolved as his PR is also working on this file and waiting would avoid a merge conflict

Issue #456, opened 2018-05-29
Implement black code formatting as pre-commit


bbengfort:
In order to avoid style-related comments when discussing code with contributors, I propose that we implement the [black](https://github.com/ambv/black) python code formatter as a pre-commit requirement for development. 

This will potentially have a large impact on our current code-base, so should be undertaken right before a release to ensure that we don't end up with large merge conflicts. Also for discussion:

- what styles do we want to customize with black? 
- is black in its beta stage stable enough to use with our project?
- do we set this up as a pre-commit in GitHub/Travis/locally?

bbengfort:
@ambv -- if you have any suggestions how we should go about this for our project or know of any open source libraries that have implemented black, we'd appreciate your thoughts!

bbengfort:
Awesome, thanks @ambv!

rebeccabilbro:
Note: Moving this from v1.0 to v1.1 &mdash; we've run `black` on all the files that are about to be merged in #957; this issue concerns implementing a pre-commit hook that automatically runs `black` ahead of `git push`

stefmolin:
It's been a while since this issue was updated. Is this still something you want to include? I'm happy to put together a pre-commit hook in a PR, if it's still of interest.

bbengfort:
@stefmolin that would be great. I guess it's been quite a while since we've thought about this. The precommit hook would require users to have black installed before they could commit right? Or would you only run black if it was available? It seems like we'd also need to update the contributor's guide documentation and add black to one of the requirements.txt files - possibly the docs/requirements.txt?

stefmolin:
@bbengfort - I'll be implementing this using [pre-commit](https://pre-commit.com/), which makes the setup and configuration of pre-commit hooks easy :smile: So, to answer your first two questions: a user won't need to install `black` at all actually, they will just need to install `pre-commit`, which will pull down `black` (and any other checks you want to run) for them.

As for the `requirements.txt` files, that is going to be a question of how strict you want to be. I'm a first-time contributor, so I don't know what kind of checks you have automated on PRs, but if you don't have `black` in there, you might want to require people to set this up on their end. Aside from installing `pre-commit`, people will have to run `pre-commit install` to install the hook the first time, so the process wouldn't be automated even if you include it in the `requirements.txt` file.

[...]

PR #925, opened 2019-07-16
Check if an estimator is already fitted before calling fit

### LINKED ISSUE #297: Check if an estimator is fitted and visualize without modification

rebeccabilbro:
This PR resolves #297, which requested that we add a feature that would allow users to pass in both fitted and unfitted models into YB visualizers.

1. Added an `is_fitted` method to `yellowbrick/utils/helpers.py` that checks if an estimator is fitted simply by calling `predict` on it and checking to see if a `NotFittedError` is raised from sklearn. _Note that this is the solution proposed to us [here](https://github.com/DistrictDataLabs/yellowbrick/issues/297#issuecomment-489187568) and also suggested on [stackoverflow](https://stackoverflow.com/a/39900933/6552250), though it remains unclear how it will perform with sklearn-style Estimators and Transformers from other 3rd party libraries like Keras, XGBoost, etc._
2. Updated the base class for `yellowbrick.classifier` to call `is_fitted` before calling `fit` (hehe &mdash; that was the easy one!)
3. Updated the regressors:
    - Updated `Alphas` to call `is_fitted` before calling `fit`
    - Updated `ResidualsPlot` to call `is_fitted` before calling `fit`
4. Updated the clusterers
    - Updated `ICDM` to call `is_fitted` before calling `fit`
    - Updated `SilhouetteVisualizer` to call `is_fitted` before calling `fit`
5. Updated the `FeatureImportances` visualizer to call `is_fitted` before calling `fit`
6. Made miscellaneous linting clean-ups to those same files (using `black`, which is consistent with #456 ) 

### Sample Code and Plot

There are a lot of examples, so I've added them to [this notebook](https://github.com/rebeccabilbro/yellowbrick/blob/check-is-fitted/examples/rebeccabilbro/check_is_fitted.ipynb) to make it easier to see the fitted vs. unfitted examples.

#### Code/Docstring Changes
- [x] Add `check_fitted` utility that determines whether or not to check if a model is fitted.
- [x] Update affected docstrings to explain estimator param (can be fitted or unfitted)
- [x] Add `is_fitted` param to affected visualizers

#### Tests
- [x] Add test coverage for the new `is_fitted` and `check_fitted` helper methods (check with a bunch of different sklearn estimators just to be safe!)
- [ ] Add a test to each of the affected visualizers that calls fit on a MockEstimator and makes sure it's `fit` doesn't gets called when `is_fitted=='auto'` or `is_fitted=='True'` but does get called if `is_fitted=='False'`

#### Docs
- [ ] Add an explanation & example to the FAQ to show how YB works with fitted/unfitted estimators

#### Questions
- [x] Check that this also works with the quick methods? Or leave this as a todo for #600 ?
- [x] Figure out what to do (if anything) with multi-model visualizers like `Elbow`, `RFE`, `ValidationCurve` &mdash; is `is_fitted` even a real use-case in these cases?

### CHECKLIST


rebeccabilbro:
@bbengfort thank you so much for taking an initial look at this; I realize that it's already touching a lot of files and is likely to expand out to more, so before I make any additional changes, I wanted to talk through some of the points and questions you raised!

> Probably the most important thing to conclude this PR is a fairly comprehensive set of tests just for our confidence and anxiety levels. 

I agree &mdash; just spitballing here, but are you thinking the tests verify that if a visualizer gets passed an estimator that is already fitted, the estimator's internal params aren't changed? 

Also just want to clarify that you're envisioning adding tests to all affected visualizers? Which (I think) include:
- `ClassificationScoreVisualizer`
    - `ClassPredictionError`
    - `ClassificationReport`
    - `ConfusionMatrix`
    - `PRCurve`
    - `ROCAUC`
- `Alphas`
- `ResidualsPlot`
- `ICDM`
- `SilhouetteVisualizer`
- `FeatureImportances`

> Great job on the discovery to determine where estimator fit calls were in the code! I'll admit you found a few places that I had forgotten wrapped an underlying estimator! I didn't do a review of my own because this seemed so thorough but didn't know if you were hoping if I'd do a second look to verify that we got everything, let me know if you'd like me to do that. Just out of curiosity how did you perform the search? 

Thanks 😅 since our library has gotten pretty big and complex, it would be great to get some feedback on my search algorithm to ensure that I haven't missed anything obvious or any edge cases. If you have any thoughts or suggestions, it was:
1. Search for any visualizer that had the substring `self.estimator.fit`: this wasn't perfect, since it also returned things like `Elbow` and `DecisionBoundaries`, which have the multimodel logic, and it didn't return `ResidualsPlot`. 
2. Search for any visualizer that had the substring `, self).fit(X`: this caught some of the other cases (but also had a lot of new false positives). 

So... it was kind of 🎃 

> Also, I just wanted to confirm that things like Manifold or TSNE were purposefully excluded since they already have their own fit management methods.

Yep! 

> I think that we should put some note in the docs about not fitting already fitted estimators - we'll want to be very clear to our users about what is going on because this might mask expected behavior. I think that at the very least we should update the docstrings of as many visualizers as we can about the behavior of fit and the estimator. 

Just want to make sure I'm clear on what you're suggesting; you propose that we add a sentence to the docstrings of each of the above visualizers to the effect of:

```
    Parameters
    ----------
    model : scikit-learn style estimator
        Can be a fitted or an unfitted model; if an unfitted model is passed in, `fit` 
        will be called by the Visualizer. 
```

> We might also want to add an FAQ entry.

Great idea! I'll add this to the TODOs

> As for the quick methods and the multi-model visualizers, I don't think any changes need to be made. The quick methods call fit of their visualizers, which does the check for isfitted - so I think that's already taken care of. And I think you're right when you say that multi-model visualizers are a completely different API.

👍 

> One other thing I'd like to discuss is adding a parameter to the visualizers: check_fitted=True to allow the user to have some control of this behavior. The is fitted behavior then becomes:
    if not (self.check_fitted and is_fitted(self.estimator)):
    self.estimator.fit(X, y) 
> I know that logic is painful but I think I've got the truth table correct:
    check_fitted is True:
    - is_fitted is True: don't fit
    - is_fitted is False: fit
    check_fitted is False:
    - is_fitted is True: fit
    - is_fitted is False: fit
> I know this is a lot more work and means editing super-classes, init params, and docstrings - but I think it would also make me the most confident given how unsure we are about how is_fitted works. What do you think?

I think I'm picking up what you're laying down. And we assume folks'll want `check_fitted==False` by default?

bbengfort:
A suggestion to allow both automatic check and user control (I'm not necessarily held to any of these names):

```python
def check_fitted(estimator, is_fitted_by='auto', **kwargs):
    """
    Returns True if the model is fitted. If ``is_fitted_by`` is ``'auto'`` then the 
    check uses the automatic ``is_fitted`` method to determine if a ``NotFitted``
    error is raised. If ``is_fitted_by`` is ``True`` or ``False``, we assume that
    the user has supplied the information about whether or not the model is fit.

    Parameters
    -----------
    estimator : sklearn.Estimator
        The model to check if it is fitted or not

    is_fitted_by : bool or str, default: 'auto'
        If bool, that value is returned, otherwise ``is_fitted`` is used to check
        for an exception

    kwargs : dict
        Other optional parameters specific to the ``is_fitted_by`` mechanism.

    Returns
    --------
    is_fitted : bool
        Whether or not the model is already fitted
    """
    if isinstance(is_fitted_by, str) and is_fitted_by.lower() == 'auto':
        return is_fitted(estimator)
    return bool(is_fitted_by)
```

We could then extend this method in the future if needed, for example, we could add `is_fitted_by='coefs` that does a check to see if `hasattr(estimator, `coefs_`)`, which might help with our model checking. 

Then visualizers would do the following:

```
class MyVisualizer(ModelVisualizer):
    """
    Parameters
    -----------
    estimator : sklearn.Estimator
        My visualizer wraps an Estimator in order to produce the visualization. If the
        estimator is not fitted, it is fit when the visualizer is fitted unless otherwise 
        specified by ``is_fitted``.

    is_fitted : bool or str, default='auto'
        Specify if the wrapped estimator is already fitted. If False, the estimator will
        be fit when the visualizer is fit, otherwise, the estimator will not be modified.
        If 'auto' (default), Yellowbrick will check if the estimator is fitted before fitting
        it again.
    """

    def __init__(self, estimator, ax=None, is_fitted='auto'):
        super(MyVisualizer, self).__init__(estimator, ax=ax, is_fitted=is_fitted)

    def fit(self, X, y=None):
        if not check_fitted(self.estimator, is_fitted_by=self.is_fitted):
            self.estimator.fit(X, y)
```

rebeccabilbro:
Oh, I really like this! I like that the visualizer attribute is called `is_fitted` and is set to `auto` by default (which feels very sklearn), but allows the user to override if they want (this should be particularly helpful with sklearn-style estimators that come from other 3rd party dependencies and might have slightly different fit behaviors), and I like that the add `check_fitted` util allows us to keep the checking that happens inside the actually visualizers to a single line!

PR #926, opened 2019-07-17
Add PCA plot directives

nickpowersys:
This relates to #687.

Each of four code directives is converted to plot directives:
- PCA, 2D
- PCA, 3D
- PCA biplot, 2D
- PCA biplot, 3D


nickpowersys:
1) After looking more closely, I understand that the changes I made to pca.py are not the standard way to format parameters, and it results in a lack of consistency with the rest of the docs. 
2) The reason that I did this in the first place is was due to a behavior stemming from recent Numpydocs and Sphinx versions. The versions I used to test locally with Windows were Sphinx 2.1.2 and numpydocs 0.9.1
3) There may be a more scalable way to handle this involving using a different version of numpydocs or Sphinx to build the docs, assuming the Numpydocs/Sphinx issue is not fixed before the 1.0 release. 

The investigation of the cause starts focusing on Numpydocs and Sphinx and a connection to HTML here:
[Comment 1](https://github.com/numpy/numpydoc/issues/215#issuecomment-487673561)
[Comment 2](https://github.com/numpy/numpydoc/issues/215#issuecomment-491534198)

nickpowersys:
The cause for the change is that HTML5 is used by default since Sphinx 2.0.


https://github.com/sphinx-doc/sphinx/issues/6472

Kautumn06:
@rebeccabilbro &mdash; those are great points about both the Manifold and JointPlot visualizers! My only suggestion is that we should update the code block examples to use our new datasets module. I'm happy to do this right now and it should be a very quick fix. Just let me know what you think!

rebeccabilbro:
That would be great @Kautumn06! Might be worth it to take a quick look at #921 (which also modifies a fair number of docs), just in case to avoid any merge conflicts, but there might not actually be any overlap in modified files (sorry, on my phone right now or else I’d take a closer look to see for sure!). Thank you!

lwgray:
@rebeccabilbro I agree with you.  We can close it out but i think we should
make sure the examples use the latest dataset module.

Kautumn06:
@rebeccabilbro &mdash; good idea about #921! Fortunately, I looked it over and it doesn't modify the `manifold.rst` or `jointplot.rst` files, so there wouldn't be any conflict there.

Issue #677, opened 2018-12-28
Extend ColorMap with target type detection and color resolution


bbengfort:
**Describe the solution you'd like**

The [`yellowbrick.style.colors.ColorMap`](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/style/colors.py#L124) object is one of the most underutilized objects in our code base and it solves many of the color issues we have such as #325 #327 #244 #556 #70 #73 #399. Now that #588 and #73 are solved in #676, I propose that we extend this object with target type detection and color resolution so that it can be used more meaningfully. The new implementation should do the following:

1. Operate as a function or dictionary that returns a color for a value.
2. Should be able to return both categorical and continuous color scales.
3. Should be able to accept a palette or cmap name, or a list of colors and resolve them.
4. Should be able to detect the target type and determine continuous or discrete colors

**Is your feature request related to a problem? Please describe.**

Our color handling is inconsistent at best. A number of visualizers such as RadViz, ParallelCoordinates, PCA, TSNE, etc. only allow for discrete/categorical colors when they could benefit from continuous colors for regression problems. The `resolve_colors` function only works for discrete values, the use of palette names and cmap names is also inconsistent. By creating a single color determination object across all of Yellowbrick, we'll have an easier time managing colorization in our plots.

**Examples**

Categorical colors with color maps/names:

```python
colors = ColorMap(colors='Set1', names=['a', 'b', 'c'])
colors['a'] # 'red'
colors[0]  # 'red'

colors = ColorMap(colors='Blues', n_colors=2)
colors['a']
colors[0]
colors['b'] # raises KeyError
```

Continuous colors

```python
colors = ColorMap(colors='Jet', minv=0.0, maxv=12.0)
colors[0.3]
colors[1.2]
```

Learning and transforming colors like an estimator

```python
colors = ColorMap(colors='RdBuGn').fit(y)
colors.target_type_ # 'continuous' 
colors.transform(y) # list of colors returned 

colors = ColorMap().fit(y)
colors.target_type_ # 'discrete' 
colors.transform(y) # ['r', 'b', 'g', 'r', 'g', 'r', ...]
```

Naba7:
Can I be assigned this issue? Please help me proceed with this.

pdamodaran:
[...]

Issue #556, opened 2018-08-10
Convert ScatterVisualizer to a mixin


rebeccabilbro:
**Describe the solution you'd like**
Convert `ScatterVisualizer`, currently in contrib, into a YB mixin.

**Is your feature request related to a problem? Please describe.**
Besides `ScatterVisualizer` many YB visualizers implement some kind of scatterplot: 

- [RadViz](https://github.com/DistrictDataLabs/yellowbrick/blob/bf5bd3f297ee6fdb589e457ecb0f954f085a5bf8/yellowbrick/features/radviz.py#L222)
- [Dispersion](https://github.com/DistrictDataLabs/yellowbrick/blob/25f1b9a6f3bdf1fa1a4cb88412b3ee095488d3ff/yellowbrick/text/dispersion.py#L106)
- [Manifold](https://github.com/DistrictDataLabs/yellowbrick/blob/75b0f6aa245f17844c1701383df0e02bc37875f4/yellowbrick/features/manifold.py#L388)
- [JointPlotVisualizer](https://github.com/DistrictDataLabs/yellowbrick/blob/9e8f76805132215aa77ee5a0a1ff20ca7034faf9/yellowbrick/features/jointplot.py#L247)
- [PCADecomposition](https://github.com/DistrictDataLabs/yellowbrick/blob/4d0483edd1468855df714f77bca1a0a93f01cbce/yellowbrick/features/pca.py#L157)
- [ResidualsPlot](https://github.com/DistrictDataLabs/yellowbrick/blob/718fc5b6f66a89527ee05f12de6dbe298b919e7f/yellowbrick/regressor/residuals.py#L173)
- [PredictionError](https://github.com/DistrictDataLabs/yellowbrick/blob/718fc5b6f66a89527ee05f12de6dbe298b919e7f/yellowbrick/regressor/residuals.py#L173)
- [TSNE](https://github.com/DistrictDataLabs/yellowbrick/blob/5ae6e320615ea32e6d1c201bdc9b551175c1e42b/yellowbrick/text/tsne.py#L336)

And it's likely that future visualizers may also require some form of scatterplot. In order to implement these in a more uniform and predictable way, it would be convenient to convert the current `ScatterVisualizer`, which was recently moved to contrib, into a YB mixin that could both serve as a base for the above visualizers and support future development.

See discussion in #475

bbengfort:
See also #676 and #677 which both would influence this mixin.

bbengfort:
[...]

Issue #476, opened 2018-06-14
Enhance PCA Decomposition


bbengfort:
We should enhance the `PCADecomposition` visualizer to provide many of the features the `Manifold` visualizer provides, including things like:

- [ ] Color points by class with a legend (See #458)
- [ ] Color points by heatmap for continuous y and add a colorbar 
- [ ] Add alpha parameter (see #475)
- [ ] Add random state to pass to PCA 
- [ ] Allow user to pass in a PCA transformer/pipeline 
- [ ] Update tests with better random data sets (more points; see manifold tests) 
- [ ] Include explained variance/noise variance (or explained variance ratio) in chart  
- [ ] Enhance biplots documentation

See also #455 as another enhancement that might not be related to this enhancement.

rohit-ganapathy:
Hey! i'm interested in tackling this.

bbengfort:
@rohit-ganapathy - that would be great, feel free to open a PR when you're ready for us to take a look.

Naba7:
Can I start working on this,even if @rohit-ganapathy is assigned?

rebeccabilbro:
Hello @Naba7 &mdash; as we explained last week in response to your questions on #738 and #677, we do not "assign" issues or reserve issues for contributors. Anyone is welcome to submit a PR for a feature or bugfix they work on. 

However, given that you already have one PR open that still needs to be completed (#755), have started working on #615, and are new to working on Yellowbrick and still getting to know our API, we would *really* appreciate if you would focus on getting those first PRs across the finish line before starting anything new. 

We appreciate your enthusiasm about contributing to Yellowbrick. One of the most important lessons to learn is that open source is a marathon, not a sprint, so we hope you can be patient and enjoy the journey &mdash; we promise Yellowbrick isn't going away!

Naba7:
It's so exciting and fun. I want to know and learn things quick. So,I am
asking questions to get assigned everywhere. Sorry I have noted this now.

bbengfort:
@naresh-bachwani has this issue been fixed by your work this summer?

naresh-bachwani:
@bbengfort I think that explained variance charts are left! But that will be covered in decomposition, right?

bbengfort:
@naresh-bachwani `ExplainedVariance` is separate to this issue. Would you mind ticking the checkboxes above based on your work?

BrandonKMLee:
[...]

PR #927, opened 2019-07-17
Extend DataVisualizer and Update DataVisualizer Subclasses

### LINKED ISSUE #923: Remove duplicate code from sub classes of DataVisualizer.
`DataVisualizer` now handles target_type_identification(implemented in #893). The visualizer which extends `DataVisualizer` have been handling these functionalities internally. Now we can update these visualizers to remove the duplicate codes.

For e.g., ParallelCoordinates(which handles only discrete target for now) calculates the classes_ internally by using `self.classes_ = [str(label) for label in np.unique(y)]`. Also, the features are manually extracted. These both functionality have been implemented in #893. Apart from these `self._colors` calculated in some visualizer is also dealt with in #893. 

Following classes extends DataVisualizer:

1.ParallelCoordinates
2. RadViz
3. contrib/missing/MissingDataVisualizer.
4. contrib/ScatterVisualizer

bbengfort:
This PR is primarily to fix #923 - which I thought would be a simple and short update to get @naresh-bachwani moving forward on his `ProjectionVisualizer` project. Nothing is simple, however, and the cleanup caused me to have to extend functionality and one thing led to another so here we are. 

Please see my inline comments in the PR for more detail, but broadly I have made the following changes:

1. I have extended the `DataVisualizer` with better color handling capability
2. I have updated `ParallelCoordinates` and `RadViz` to make use of this functionality
3. I have cleaned up some other areas in the code that were affected by the changes.
4. Fixed technical debt created by #73 

These changes pave the way to make ParallelCoordinates and RadViz more general data visualizers for different target types (radviz for regressors as discussed in #73!), and could be the start for solutions such as those requested in #677, #476, #556

### CHECKLIST


tests/test_contrib/test_scatter.py
```diff
@@ -113,7 +113,7 @@ def test_scatter_xy_changes_to_features(self):
         Assert that x,y with no features will not raise scatterviz error
         """
         visualizer = ScatterViz(x='one', y='two')
-        assert visualizer.features_ == ['one', 'two']
+        assert visualizer.features == ['one', 'two']
```
bbengfort: This test ensures that the "old functionality" is maintained; see the note in `scatter.py` for more detail.

tests/test_features/test_base.py
```diff
@@ -18,129 +18,444 @@
...
     request.cls.continuous = Dataset(X, y)
 
 
 ##########################################################################
 ## FeatureVisualizer Base Tests
 ##########################################################################
 
-class TestFeatureVisualizerBase(VisualTestCase):
+@pytest.mark.usefixtures("discrete")
+class TestFeatureVisualizer(object):
```
bbengfort: I extended the tests for the `FeatureVisualizer` to make the whole subsection of the class hierarchy more robust.

tests/test_features/test_base.py
```diff
@@ -18,129 +18,444 @@
...
+        assert viz.transform.called_once_with(X, y)
+        assert viz.poof.called_once_with(outpath="a.png", clear_figure=True)
+
+
+##########################################################################
+## MultiFeatureVisualizer Tests
+##########################################################################
+
+@pytest.mark.usefixtures('discrete')
+class TestMultiFeatureVisualizer(object):
```
bbengfort: MultiFeatureVisualizer was previously untested -- tests added.

tests/test_features/test_base.py
```diff
@@ -18,129 +18,444 @@
...
+        assert len(oz.features_) == X.shape[1]
+
 
 ##########################################################################
 ## DataVisualizer Tests
 ##########################################################################
 
 @pytest.mark.usefixtures("discrete", "continuous")
-class TestDataVisualizerBase(VisualTestCase):
+class TestDataVisualizer(object):
```
bbengfort: DataVisualizer now does so much work that I believed it was critical for it to be fully tested. I would welcome an analysis of the coverage of the tests here since a number of downstream visualizers will rely on this class including:

- PCA
- Manifold
- ParallelCoordinates
- RadViz

In particular, I'm worried about my coverage of user parameters (e.g. color/colormap, etc.) - though the defaults are well tested.

tests/test_features/test_pca.py
```diff
@@ -176,7 +176,7 @@ def test_biplot_3d(self):
         Test the PCADecomposition 3D biplot (proj_features).
         """
         params = {
-            "features": "ABCDEFGHIKLM",
+            "features": list("ABCDEFGHIKLM"),
```
bbengfort: When `self.features_` was changed to a `np.array` you can no longer treat a string like a list.

tests/test_features/test_projection.py
```diff
@@ -121,7 +122,7 @@ def test_discrete_plot(self):
         classes = ["a", "b", "c", "d", "e"]
         visualizer = MockVisualizer(projection=2, colormap="plasma", classes=classes)
         X_prime = visualizer.fit_transform(X, y)
-        assert(visualizer.classes_ == classes)
+        npt.assert_array_equal(visualizer.classes_, classes)
```
bbengfort: Side effect of changing `classes_` to a np array.

yellowbrick/contrib/missing/base.py
```diff
@@ -51,6 +53,12 @@ def fit(self, X, y=None, **kwargs):
         self : instance
             Returns the instance of the transformer/visualizer
         """
+        # Do not call super here - the data visualizer has been refactored
+        # to provide increased functionality that is not yet compatible with
+        # the current implementation. This mimicks the previous functionality.
+        # TODO: Refactor MissingDataVisualizer to make use of new features.
+        self.features_ = self.features
```
bbengfort: Part of #923 was to ensure that no duplicate work occurred in subclasses; however, the missing data visualizers require specialized functionality. It is possible to update them to use the new functionality but would require a bit more research. Both this and scatter I simply updated to behave similarly to the old functionality.

yellowbrick/features/pcoords.py
```diff
@@ -333,34 +346,15 @@ def fit(self, X, y=None, **kwargs):
...
-            self.classes_ = [str(label) for label in np.unique(y)]
-
-        # Create the color mapping for each class
-        # TODO: Allow both colormap, listed colors, and palette definition
-        # TODO: Make this an independent function or property for override!
-        color_values = resolve_colors(
-            n_colors=len(self.classes_), colormap=self.colormap, colors=self.color
-        )
-        self._colors = dict(zip(self.classes_, color_values))
-
```
bbengfort: A lot of duplicate code was removed and functionality unified in the super class - making things easier to test and change!

yellowbrick/features/pcoords.py
```diff
@@ -210,6 +212,17 @@ class ParallelCoordinates(DataVisualizer):
...
+        The names of the features discovered or used in the visualizer that
+        can be used as an index to access or modify data in X. If a user passes
+        feature names in, those features are used. Otherwise the columns of a
+        DataFrame are used or just simply the indices of the data array.
+
+    classes_ : ndarray, shape (n_classes,)
+        The class labels that define the discrete values in the target. Only
+        available if the target type is discrete. This is guaranteed to be
+        strings even if the classes are a different type.
+
```
bbengfort: I'm a bit worried about consistency in the docstring of ParallelCoordinates, RadViz, and DataVisualizer ... then eventually in PCA and Manifold as well. Any advice would be welcome!

yellowbrick/features/pcoords.py
```diff
@@ -425,17 +418,10 @@ def draw_instances(self, X, y, **kwargs):
         for idx in range(len(X)):
             Xi = X[idx]
             yi = y[idx]
-
-            # TODO: generalize this duplicated code into a single function
-            if isinstance(yi, str):
-                label = yi
-            else:
-                # TODO: what happens if yi is not in classes?!
```
bbengfort: Problem solved!

yellowbrick/features/projection.py
```diff
@@ -61,7 +61,7 @@ class ProjectionVisualizer(DataVisualizer):
         If classes is None and a y value is passed to fit then the classes
         are selected from the target vector.
 
-    color : list or tuple of colors, default: None
+    colors : list or tuple of colors, default: None
```
bbengfort: Right, so I changed this because `colors` makes way more sense than `color` in the new context, particularly because it really must be a list. This meant a lot of changes in a lot of places - please be alert for ways I may have broken things just by adding an s! Unfortunately, VSCode wasn't a whole lot of help here ...

yellowbrick/features/projection.py
```diff
@@ -92,18 +92,24 @@ class ProjectionVisualizer(DataVisualizer):
...
-    def __init__(self, ax=None, features=None, classes=None, color=None,
+    def __init__(self, ax=None, features=None, classes=None, colors=None,
              colormap=None, target_type="auto", projection=2, alpha=0.75,
              colorbar=True, **kwargs):
 
         # TODO: add this to resolve_colors
-        if color==None and colormap==None:
-            colormap=palettes.DEFAULT_SEQUENCE
+        if colors is None and colormap is None:
+            colormap = palettes.DEFAULT_SEQUENCE
```
bbengfort: Note that I did look into moving this to `resolve_colors` -- right now if both are None, `resolve_colors` just returns the default color cycle. We do still need to do something better than this though.

yellowbrick/features/__init__.py
```diff
@@ -26,3 +26,4 @@
 from .importances import FeatureImportances, feature_importances
 from .rfecv import RFECV, rfecv
 from .manifold import Manifold, manifold_embedding
+from .base import TargetType
```
bbengfort: Thought it would be important to have this at the top level, I also imported it in `yellowbrick/target/__init__.py` so it's available from both packages.

yellowbrick/features/base.py
```diff
@@ -20,110 +20,184 @@
...
+        feature names in, those features are used. Otherwise the columns of a
+        DataFrame are used or just simply the indices of the data array.
     """
 
     def __init__(self, ax=None, features=None, **kwargs):
         super(MultiFeatureVisualizer, self).__init__(ax=ax, **kwargs)
 
         # Data Parameters
-        self.features_ = features
+        self.features = features
```
bbengfort: This change is important - we're saving the hyperparameter independent of the learned parameter so users can compare what is passed into the class and what is "learned" particularly if one is a different type from the other. The learned parameter is always a numpy array. 

Note there is similar new behavior for `classes_` as well.

yellowbrick/features/base.py
```diff
@@ -20,110 +20,184 @@
...
+            # TODO: allow the user specified features to filter the dataset
+            if len(self.features) != n_columns:
+                raise YellowbrickValueError((
+                    "number of supplied feature names does not match the number "
+                    "of columns in the training data."
+                ))
 
-        # Handle the feature names if they're None.
-        if self.features_ is None:
+            self.features_ = np.array(self.features)
```
bbengfort: Features is now always a numpy array.

yellowbrick/features/base.py
```diff
@@ -234,49 +338,66 @@ def fit(self, X, y=None, **kwargs):
...
+                self.classes_ = np.asarray([str(c) for c in labels])
+
+            # Create a map of class labels to colors
+            color_values = resolve_colors(
+                n_colors=len(self.classes_),
+                colormap=self.colormap,
+                colors=self.colors
+            )
             self._colors = dict(zip(self.classes_, color_values))
+            self._label_encoder = dict(zip(labels, self.classes_))
```
bbengfort: I've added this to make it easier to pass class labels and have encoded target values. There were a number of bugs associated with this (but maybe in the classifier visualizers) but this fixes it here as well.

yellowbrick/features/base.py
```diff
@@ -307,3 +428,57 @@ def _determine_target_color_type(self, y):
...
+
+    def get_target_color_type(self):
+        """
+        Returns the computed target color type if fitted or specified by the user.
+        """
+        if self._target_color_type is None:
+            raise NotFitted("unknown target color type on unfitted visualizer")
+        return self._target_color_type
+
+    def get_colors(self, y):
```
bbengfort: This is the primary new functionality - because the DataVisualizer determines the target color type I thought it was also important for it to be able to compute the colors based on the target. This will also reduce a lot of duplicated functionality in the subclasses.

yellowbrick/features/base.py
```diff
@@ -20,110 +20,184 @@
...
+        Returns
+        -------
+        self : MultiFeatureVisualizer
+            Returns the visualizer/transformer for use in Pipelines and chaining.
         """
+        n_columns = X.shape[1]
+
+        if self.features is not None:
+            # Use the user-specified features with some checking
+            # TODO: allow the user specified features to filter the dataset
```
bbengfort: See #175

yellowbrick/features/base.py
```diff
@@ -293,17 +403,67 @@ def _determine_target_color_type(self, y):
         if y is None:
             self._target_color_type = TargetType.SINGLE
         elif self.target_type == TargetType.AUTO:
-            # NOTE: See #73 for a generalization to use when implemented
```
bbengfort: Getting rid of even more technical debt!

yellowbrick/utils/target.py
```diff
@@ -18,19 +18,46 @@
...
+class TargetType(Enum):
+    """Constants for defining target colors by input type"""
+
+    AUTO = 'auto'
+    SINGLE = 'single'
+    DISCRETE = 'discrete'
+    CONTINUOUS = 'continuous'
+    UNKNOWN = "unknown"
+
+    def __eq__(self, other):
```
bbengfort: Required so you can do `TargetType.AUTO == 'auto'`

tests/test_features/test_pca.py
rebeccabilbro: :+1:

tests/test_features/test_projection.py
rebeccabilbro: :+1:

yellowbrick/contrib/missing/base.py
rebeccabilbro: :+1:

yellowbrick/features/base.py
```diff
@@ -20,165 +20,244 @@
...
 
-    Accepts as input a DataFrame or Numpy array.
+    Parameters
+    ----------
+    ax : matplotlib.Axes, default: None
+        The axis to plot the figure on. If None is passed in the current axes
+        will be used (or generated if required).
+
+    kwargs : dict
+        Any additional keyword arguments to pass to the base Visualizer.
```
rebeccabilbro: Thank you very much for enriching this docstring!

yellowbrick/features/base.py
rebeccabilbro: Thanks for clarifying this in the docstring by noting both the parameter `features` and the attribute `features_`!

yellowbrick/features/base.py
rebeccabilbro: Nice addition!

yellowbrick/features/base.py
```diff
@@ -234,49 +325,66 @@ def fit(self, X, y=None, **kwargs):
...
+            )
             self._colors = dict(zip(self.classes_, color_values))
+            self._label_encoder = dict(zip(labels, self.classes_))
 
         # Compute target range if colors are continuous
         elif self._target_color_type == TargetType.CONTINUOUS:
             y = np.asarray(y)
             self.range_ = (y.min(), y.max())
 
+            # TODO: allow for Yellowbrick palettes here as well
```
rebeccabilbro: Perhaps we could use the [updated `resolve_colors` utility](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/style/colors.py#L57-L151)?

yellowbrick/features/base.py
rebeccabilbro: 🙂

yellowbrick/features/pcoords.py
rebeccabilbro: Good question... perhaps we should open up a follow-on task. @Kautumn06 &mdash; thoughts? You're always so good at helping with consistency in our communications across similar visualizers!

yellowbrick/features/pcoords.py
```diff
@@ -174,15 +177,14 @@ class ParallelCoordinates(DataVisualizer):
     shuffle : boolean, default: True
         specifies whether sample is drawn randomly
 
-    color : list or tuple, default: None
-        optional list or tuple of colors to colorize lines
-        Use either color to colorize the lines on a per class basis or
-        colormap to color them on a continuous scale.
+    colors : list or tuple, default: None
```
rebeccabilbro: Noticing a few places here where we seem to be deprecating the old `color` param for `colors` &mdash; which I'm at peace with (possibly no one but me and my Visual Analytics students ever used these 🤣) but maybe just something we should note as backwards incompatible in our endless release notes doc...

yellowbrick/features/pcoords.py
```diff
@@ -333,34 +346,15 @@ def fit(self, X, y=None, **kwargs):
         self : instance
             Returns the instance of the transformer/visualizer
         """
+        # Determine the features, classes, and colors
+        super(ParallelCoordinates, self).fit(X, y)
```
rebeccabilbro: See my other comment about how/whether we could better document our use of calling `super(Visualizer, self).fit(X,y)` inside `fit` for different visualizers &mdash; not sure what would make most sense but I know that this would help me reason better about our class hierarchy as a contributor!

yellowbrick/features/pcoords.py
rebeccabilbro: So slick!

yellowbrick/features/pcoords.py
rebeccabilbro: 🙂

yellowbrick/features/projection.py
rebeccabilbro: Thanks for noting this! It was also something I was noticing since I make a lot of use of the colorizing params during my Visual Analytics class and various talks. I'm totally at peace with the change in spite of the backwards incompatibility here since I agree that the naming makes more sense (the inconsistency in how we name our colorizing params across &mdash; and even within &mdash; visualizers is often a headache)! Perhaps we can just make a good, clear note about how we've changed the color params here (and in, for instance, #917) in our release notes, and  in our [FAQ](https://www.scikit-yb.org/en/develop/faq.html#how-can-i-change-the-color-of-a-yellowbrick-plot) &mdash; thoughts?

yellowbrick/features/radviz.py
```diff
@@ -140,18 +146,27 @@ class RadialVisualizer(DataVisualizer):
...
+    features_ : ndarray, shape (n_features,)
+        The names of the features discovered or used in the visualizer that
+        can be used as an index to access or modify data in X. If a user passes
+        feature names in, those features are used. Otherwise the columns of a
+        DataFrame are used or just simply the indices of the data array.
+
+    classes_ : ndarray, shape (n_classes,)
+        The class labels that define the discrete values in the target. Only
+        available if the target type is discrete. This is guaranteed to be
+        strings even if the classes are a different type.
```
rebeccabilbro: Thanks for enriching the docstring here!

rebeccabilbro:
Have just been reading through old issues &mdash; wondering if this (or perhaps one of the upcoming Manifold PRs that you and @naresh-bachwani are working on) will also close #437?

yellowbrick/contrib/scatter.py
```diff
@@ -215,6 +215,12 @@ def fit(self, X, y=None, **kwargs):
         """
         _, ncols = X.shape
 
+        # NOTE: Do not call super for this class, it conflicts with the fit.
```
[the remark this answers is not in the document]
bbengfort: I can take a look at those classes to see why they do not call super. I feel that if a subclass does not call super then that means that it might be poorly subclassed - e.g. it is using some functionality of the superclass and not other functionality. As we move forward if this is the case we should consider:

1. Breaking the superclass into two classes or a mixin and extending the inheritance tree
2. Extensively commenting why we don't call super (as in the note above)

Otherwise:

All of our visualizers should call super. 

Not calling super is a code smell and it can definitely break our code if we add functionality to the superclass but that is not propagated down to all of our visualizers! And you're right, it gets even harder the larger Yellowbrick gets and the farther we are from the original code.

yellowbrick/features/projection.py
bbengfort: Absolutely, I'll happily update the FAQ!

yellowbrick/contrib/scatter.py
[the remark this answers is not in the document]
bbengfort: Ah I see - I was mistaken; all of the visualizers you listed do call super, but they don't say why they call super either at the top or at the bottom of the method. I agree we can start doing a better job of that and when I'm working somewhere where I notice that isn't specifically documented, I will do so!

docs/faq.rst
```diff
@@ -51,79 +51,49 @@ You can change the ``title`` of a plot by passing in the desired title as a stri
...
 .. code:: python
 
-    from yellowbrick.features import ParallelCoordinates
+    from yellowbrick.features.manifold import Manifold
 
-    # Instantiate the visualizer
-    visualizer = ParallelCoordinates(
-        classes=classes, features=features, colormap="PRGn"
+    visualizer = Manifold(
+        manifold="isomap", target="continuous", colormap="YlOrRd"
```
bbengfort: @naresh-bachwani let's make sure this is true in #930!

bbengfort:
@rebeccabilbro as requested, I've updated the FAQ.

rebeccabilbro:
Thank you @bbengfort! The updated FAQ looks terrific &mdash; I just pushed up a tiny typo correction, so will get this merged in as soon as the tests finish running.

PR #951, opened 2019-08-16
Remove old baseline images from tests

rebeccabilbro:
This is a little PR that just removes a bunch of old test images from the `baseline_images` directory. These are images that were artifacts of test functions that have since been renamed or moved.


yellowbrick/__init__.py
```diff
@@ -2,6 +2,7 @@
 # A suite of visual analysis and diagnostic tools for machine learning.
 #
 # Author:   Rebecca Bilbro
+# Author:   Benjamin Bengfort
```
bbengfort: :-) Thanks!

Issue #960, opened 2019-08-28
Make quick methods primetime


bbengfort:
After the visualizer audit we believe that the quick methods are ready to be made prime time. 

Quick methods must:

1. Have a function signature identical to the Visualizer constructor with additional data for fit/score
2. Call `finalize()` and not `poof()`/`show()`
3. Return the fitted visualizer instance
4. Some visualizers must also insure that they call transform/fit_transform/score 

For each visualizer found in the list below we must:

- [x] refactor the quick method to the new API 
- [x] document the quick method 
- [x] ensure there is an associated test or tests for the quick method

See #600 for more details on this issue

- [ ] class_prediction_error
- [ ] classification_report
- [ ] confusion_matrix
- [ ] precision_recall_curve
- [x] roc_auc
- [ ] discrimination_threshold
- [ ] alphas
- [ ] cooks_distance
- [ ] prediction_error
- [ ] residuals_plot
- [ ] kelbow_visualizer
- [ ] intercluster_distance
- [x] silhouette_visualizer
- [ ] balanced_binning_reference
- [ ] class_balance
- [ ] feature_correlation
- [x] cv_scores
- [ ] learning_curve
- [ ] validation_curve
- [ ] explained_variance_visualizer
- [ ] feature_importances
- [ ] rank1d
- [ ] rank2d
- [x] rfecv
- [ ] joint_plot
- [ ] mainfold_embedding
- [x] pca_decomposition
- [x] parallel_coordinates
- [x] radviz
- [ ] dispersion
- [ ] freqdist
- [x] postag
- [x] tsne
- [ ] umap
- [ ] gridsearch_color_plot

pdamodaran:
The following is a link to the assignments:

https://docs.google.com/spreadsheets/d/1DX8RuRCVMiJrFd1YWguor57iaAmj5503BLJDX8KGjn8/edit?usp=sharing

bbengfort:
This was completed in v1.1

PR #975, opened 2019-10-05
Change sil viz to use show, add to params

mgarod:
Addresses Silhouette Visualizer for https://github.com/DistrictDataLabs/yellowbrick/issues/960

1. Add `show` to the parameters of the quick method
2. Make the visualizer in the quick method use show instead of poof

- [x] I didn't modify the unittest because it appeared to be sufficient as it stood. Here is the unit test that already existed before beginning my assignment https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_cluster/test_silhouette.py#L177-L189
- [x] Is there supposed to be additional documentation added other than within the module itself.

No picture provided because there has been no change to any output. Tests still pass.


yellowbrick/cluster/silhouette.py
```diff
@@ -285,26 +285,31 @@ def silhouette_visualizer(
...
     X : array-like of shape (n, m)
         A matrix or data frame with n instances and m features
 
     y : array-like of shape (n,), optional
         A vector or series representing the target for each instance
 
+    ax : matplotlib Axes, default: None
+        The axis to plot the figure on. If None is passed in the current axes
+        will be used (or generated if required).
+
```
bbengfort: Thank you for re-arranging the docstring params to match the function signature!

mgarod:
Thanks for the comprehensive review!

- RE `random_state` and `legend`: I traced them and could not determine where they actually get used. They kept getting passed in kwargs and died. I removed them from the test.

- In the test I set `show=False` which removed the warnings from test report.

- Added `silhouette_visualizer` to the api reference.

- Simply for a starting point, I added to the silhouette.rst. I put a short description of the quick method, and a copy-paste of the previous nfl example. After building the docs I realize that it produces exact same picture again, so I set `show=False`. Should I create an entirely new example and use new data to show off the quick method? Maybe I don't need an example at all, and should just show the one-liner of the quick method?

pdamodaran:
Thanks @mgarod for addressing the comments @bbengfort made to this PR.  I like that you included a quickmethod example in the documentation.  I think you can keep the example that you included in the documentation but take out `show=False`.  I think it will be good to demonstrate that the same visualization can be rendered by implementing the quick method.  If you can include that information to the documentation, that will be great.

tests/test_cluster/test_silhouette.py
```diff
@@ -183,7 +183,7 @@ def test_quick_method(self):
         )
 
         model = MiniBatchKMeans(3, random_state=343)
-        oz = silhouette_visualizer(model, X, random_state=93, legend=False)
+        oz = silhouette_visualizer(model, X, show=False)
```
bbengfort: :+1:

pdamodaran: @bbengfort - could you explain what you mean by `Models` and `Workflow` in your example above?

rebeccabilbro: Hey there @pdamodaran! Hope things have been well! Thought I'd jump in here since I couldn't resist 🙂 

`Models`: The type of machine learning that the `Visualizer` is designed for (e.g. Classification, Regression, Clustering &mdash; akin to the "default task" categorization on the [UCI Machine Learning Repository](https://archive.ics.uci.edu/ml/datasets.php)). For example, what kind of machine learning should you expect to do with a `SilhouetteVisualizer`?

`Workflow`: Where does this `Visualizer` fit into the [machine learning workflow](https://www.scikit-yb.org/en/develop/about.html#model-selection) &mdash; is it used for feature analysis, model evaluation, or hyperparameter tuning? For instance, at what stage of the workflow would a `ClassificationReport` be used?

pdamodaran: Thanks @rebeccabilbro for the explanation - it makes sense to me now.

@bbengfort - The template you suggested above looks good to me.

mgarod:
Added the "at-a-glance" table to the top of the documentation page with links to the api (though the link I had to create for the quick method doesn't exist yet).

Created a new example with the credit dataset for the quick method documentation.

PR #977, opened 2019-10-08
Changed rocauc quick method to use show, add image comparison test

mchestnut91:
Addresses ROCAUC Visualizer for #960   

1. Added `X_train`, `y_train`, `X_test`, `y_test` parameters to the quick method function signature
2. Added a `show` parameter to function signature
3. Made the quick method use `show` instead of `poof`
3. Added new parameters to the quick method docstring
4. Added an image comparison unit test, as well as a baseline image to compare the test against

### Sample Code and Plot

from yellowbrick.classifier.rocauc import roc_auc
from yellowbrick.datasets import load_occupancy
from sklearn.linear_model import LogisticRegression as lr

X, y = load_occupancy(return_dataset=True).to_numpy()
X_train, X_test, y_train, y_test = tts(X,y)
model = lr()
roc_auc(model, X_train, y_train, X_test=X_test, y_test=y_test)

image

### TODOs and questions

Still to do:

- [ ] Add quick method information to ROCAUC documentation


mchestnut91:
Hey @pdamodaran, 

Thank you for taking the time to review this PR! I have a few questions about your comments regarding the `train_test_split` and its implementation in the quick method. Currently I have the quick method function signature set up to require the user to supply only the `model`, `X_train` and `y_train` parameters. They also have the option to pass in `X_test` and `y_test`, but their default values are `None`. In the sample code that I provided, I split the data up and passed that into the quick method, but I could skip that step, simply plug `X` and `y` in, and it would still run by using `X` and `y` for both `fit()` and `score()`.

Are you suggesting that I should simply change the names of the `X_train` and `y_train` parameters to just `X` and `y`, or should I remove the optional `X_test` and `y_test` parameters? If I do that, does this mean that I should perform the `train_test_split` within the quick method itself, or would there simply not be a split for the quick method?

Thank you again for reviewing!

mchestnut91:
Hi @pdamodaran, what I am confused about is if I remove the `X_test` and `y_test` parameters,  then how would the user do their own `train_test_split` and pass test arguments into the quick method? 

I modeled my approach off of [this gist](https://gist.github.com/bbengfort/5d5b211c1cf49f7dac147654e75e5d24). I think that setting it up this way allows users to pass in optional `X_train` and `y_train` parameters, while still allowing for a simple `X` and `y` input if the user does not wish to `train_test_split`. Removing the `X_test` and `y_test` parameters seems like it would make doing a `train_test_split` impossible, but I could be missing something?

Thank you again for your feedback!

tests/test_classifier/test_rocauc.py
```diff
@@ -213,10 +212,11 @@ def test_rocauc_quickmethod(self):
         Test the ROCAUC quick method
         """
         X, y = load_occupancy(return_dataset=True).to_numpy()
-        model = DecisionTreeClassifier()
+        model = LogisticRegression()
```
bbengfort: At @pdamodaran's request I'm taking a quick look a this PR and I noticed this change - was just curious why it was required?

pdamodaran:
@mchestnut91 - I discussed the PR with @bbengfort and we came to agreement that the approach you took is correct.  If you can address the other comments Ben and I made, we should be good to go to get this PR merged.  Thanks again for your work on this!

tests/test_classifier/test_rocauc.py
mchestnut91: @bbengfort - This change was not required, as the quick method works with all classifiers. I changed it to be consistent with most of the other tests in the module, but I can change it back if it causes confusion!

mchestnut91:
@pdamodaran, Thank you for clarifying! 

In my most recent commit I did the following:

- Set `show = False` in the test to remove the warnings from the test report.

- Added `roc_auc` to the api reference.

- Added to rocauc.rst. Updates include a short description of the quick method, an example using the credit dataset, and an attempt at the table that @bbengfort posted yesterday.

Thank you both for your guidance thus far, looking forward to hearing your feedback.

docs/api/classifier/rocauc.rst
```diff
@@ -9,6 +9,13 @@ The Receiver Operating Characteristic (ROC) is a measure of a classifier's predi
 
 This leads to another metric, area under the curve (AUC), which is a computation of the relationship between false positives and true positives. The higher the AUC, the better the model generally is. However, it is also important to inspect the "steepness" of the curve, as this describes the maximization of the true positive rate while minimizing the false positive rate.
 
+=================   =================
+Visualizer           `ROCAUC <https://www.scikit-yb.org/en/latest/api/classifier/rocauc.html#yellowbrick.classifier.rocauc.ROCAUC>`_
+Quick Method         `roc_auc() <https://www.scikit-yb.org/en/latest/api/classifier/rocauc.html#yellowbrick.classifier.rocauc.roc_auc>`_
+Models               Classification
+Workflow             Model evaluation
+=================   =================
+
```
bbengfort: Nice! Thank you for adding this!

tests/test_classifier/test_rocauc.py
bbengfort: Ok no worries - just wanted to check; usually when there is a change like this it was because the original model wasn't working for what seemed like a minor reason to the contributor but actually revealed a small bug in our code that we needed to fix!

tests/test_classifier/test_rocauc.py
```diff
@@ -213,10 +212,11 @@ def test_rocauc_quickmethod(self):
...
 
-        # TODO: image comparison of the quick method
-        roc_auc(model, X, y)
+        #compare the images
+        visualizer = roc_auc(model, X, y)
+        self.assert_images_similar(visualizer)
```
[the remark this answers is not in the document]
bbengfort: Thanks for making this update!

pdamodaran:
Looks like we are all set to merge this PR in.  Thanks again for all of your work on this!

PR #987, opened 2019-10-30
KElbow quick method API update

mgarod:
1. Rearrange quick method function parameters to match docstring
1. Reformat docstring
1. Add show kwarg to quick method
1. Remove unnecessary args (random_state, legend) from test
1. Update docs with link, example of quick method, and add to automodule

- [x] I'm not a data science expert so the example I created for the kelbow quickmethod may be nonsense. Can someone please review the example?
- [x] I did basically nothing regarding the test since the quick method uses `xfail` due to some non-determinism in the output graph. Other tests seem to have have the same problem and fix. Is this fine?
- [ ] In the docs I included what I expect to be the hyperlink to the quick method (`kelbow_visualizer() <https://www.scikit-yb.org/en/latest/api/cluster/elbow.html#yellowbrick.cluster.elbow.kelbow_visualizer>`_), though I believe I have no way to test this until it actually gets merged and hosted. Is there a correct way to handle this?


naresh-bachwani:
Hey there @mgarod,
Great work in this PR! Although the final review would be done by @wagner2010, I think that I can answer some of your queries.

1. The example you created is great and has no such issues with it. But, I think that we can use a somewhat simple dataset(like ``nfl``) to get rid of those label encoders and make the example simpler.

2.  All you needed to do here was to add `show=False` and remove unused labels. `xfail` is used to tackle different output(due to timing lines). Due to this tests are supposed to fail and hence the use of `xfail`.

3. I have not worked with documentation a lot. I think that the @wagner2010 or other team members would be in a good position to answer this!

mgarod:
Thank you @naresh-bachwani I changed the example to use NFL instead. I think I just didn't want to use the same dataset that I had been using in the PR prior to this one.

PR #988, opened 2019-10-30
Parsing raw text with PosTagVisualizer 

ekwiecinska:
This PR introduces a new feature discussed in issue #771 which allows the user to optionally parse their raw text documents directly using the `PosTagVisualiser`.

1. `PosTagVisualiser` now accepts a `parse` and `tagger` argument, which allow the user to specify if they require parsing of raw text (their documents are untagged) and also which library and tagset to use
2. Unit tests for both `spacy` and `nltk` variants and also for the `word` and `wordpunct` taggers in `nltk`.


bbengfort:
Hi @ekwiecinska thank you so much for the PR - looks like a great start! I've been traveling so I haven't had time to review, but I'll try to get to it as soon as I can. Please bump if you haven't heard from me in a week or two!

ekwiecinska:
Hey @bbengfort that's fine, no worries. Thanks, I'm looking forwards to the review!

docs/api/text/postag.rst
```diff
@@ -5,10 +5,10 @@ PosTag Visualization
...
 
-The ``PosTagVisualizer`` is intended to support grammar-based feature extraction techniques for machine learning workflows that require natural language processing. The visualizer reads in a corpus that has already been sentence- and word-segmented, and tagged, creating a bar chart to visualize the relative proportions of different parts-of-speech in a corpus.
+The ``PosTagVisualizer`` is intended to support grammar-based feature extraction techniques for machine learning workflows that require natural language processing. The visualizer can either read in a corpus that has already been sentence- and word-segmented, and tagged, or perform this tagging automatically by specifying the parser to use (nltk or spacy). The viisualiser creates a bar chart to visualize the relative proportions of different parts-of-speech in a corpus.
```
bbengfort: minor typo: "viisualiser" ⟶ "visualizer"

docs/api/text/postag.rst
```diff
@@ -223,6 +223,58 @@ Libraries like SpaCy use tags from the Universal Dependencies (UD) framework. Th
...
+Or using `wordpunct`.
+
+.. code:: python
+
+    untagged_speech = u'Whose woods these are I think I know'
+
+    # Create the visualizer, fit, score, and show it
+    viz = PosTagVisualizer(parser='nltk', tagger='wordpunct')
+    viz.fit(untagged_speech)
+    viz.show()
```
bbengfort: Well written and clear updates to the documentation, thank you!

yellowbrick/text/postag.py
```diff
@@ -140,14 +141,16 @@ class PosTagVisualizer(TextVisualizer):
...
-        stack=False,
-        **kwargs
+            self,
+            ax=None,
+            tagset="penn_treebank",
+            colormap=None,
+            colors=None,
+            frequency=False,
+            stack=False,
+            parser=None,
```
bbengfort: Would you mind adding the argument documentation to the docstring under parameters?

yellowbrick/text/postag.py
```diff
@@ -167,21 +170,46 @@ def __init__(
         self.colormap = colormap
         self.colors = colors
         self.stack = stack
+        self.parser = parser
+        self.tagger = tagger
```
bbengfort: It appears that the parser value is validated on `__init__` using the setter, however, the tagger is not validated until `fit()`. It would be my preference if all arguments are validated at construction time rather than downstream because if the visualizer is in a pipeline, there might be a lot of computation that fails before getting to the validation exception. 

Since the `tagger` argument is only used for `nltk`, could I propose the following three options for the `parser` argument: `["nltk", "nltk_wordpunct", "spacy"]` or potentially `["word", "wordpunct", "spacy"]`. This would make validation simpler, earlier, and reduce the number of parameters which I think could be valuable.

yellowbrick/text/postag.py
```diff
@@ -167,21 +170,46 @@ def __init__(
...
+
+    @property
+    def parser(self):
+        return self.__parser
```
bbengfort: Do you think that there is a particular need for name mangling here?

yellowbrick/text/postag.py
```diff
@@ -167,21 +170,46 @@ def __init__(
...
+    @parser.setter
+    def parser(self, parser):
+        accepted_parsers = ['nltk', 'spacy']
+        if not parser:
+            self.__parser = None
+        elif parser in accepted_parsers:
+            try:
+                importlib.import_module(parser)
+            except ModuleNotFoundError:
+                raise ModuleNotFoundError("Can't find module '{}' in this environment.".format(parser))
```
bbengfort: Great use of `importlib` here! I do like this better than our try/except/None pattern that we've used in other places.

yellowbrick/text/postag.py
```diff
@@ -431,17 +522,16 @@ def show(self, outpath=None, **kwargs):
...
-    **kwargs
+        X,
+        y=None,
+        ax=None,
+        tagset="penn_treebank",
+        colormap=None,
+        colors=None,
+        frequency=False,
+        stack=False,
+        **kwargs
```
bbengfort: Please make sure to add the extra arguments to the quick method, and make sure that they're passed to the visualizer.

ekwiecinska:
Hi @bbengfort, sorry for the delay in my response. Life gets in the way! Thanks a lot for the review and comments. I'll review these and make changes this weekend.

bbengfort:
I totally understand, thank you!

docs/api/text/postag.rst
ekwiecinska: Thanks!

yellowbrick/text/postag.py
ekwiecinska: Sure, resolved in 666909c

docs/api/text/postag.rst
ekwiecinska: Good spot, resolved in 666909c

yellowbrick/text/postag.py
ekwiecinska: Thanks! My main motivation was pep8 compatibility to avoid having the imports below the top of the file.

yellowbrick/text/postag.py
ekwiecinska: Ah thanks I missed this. Resolved in d802d1f

yellowbrick/text/postag.py
ekwiecinska: This is a very good point. However the `self.tagger` argument is used for `spacy` as shown on L299-305

```
spacy = importlib.import_module('spacy')
try:
    if not self.tagger or self.tagger == "en_core_web_sm":
    nlp = spacy.load("en_core_web_sm")
else:
    nlp = spacy.load(self.tagger)
except OSError:
    raise OSError(f"Spacy model '{self.tagger}' has not been downloaded into this environment.")
```
I think that reducing the number of arguments is definitely the most valuable in terms of accessibility of the class. In light of this the solution that I propose is to merge the `tagger` and `parser` arguments into a single argument `parser`, where arguments take the form `"parser__tagger"`. For example `"nltk__wordpunct"`, or `spacy__en_core_web_sm`. The double underscore is for easy string separation if taggers contain single underscores e.g in `spacy`. Let me know what you think of this @bbengfort !

yellowbrick/text/postag.py
ekwiecinska: Hm, yes it's probably not necessary here. Resolved in dfd20f9 by changing to standard private variable property naming convention `_` rather than `__`.

yellowbrick/text/postag.py
ekwiecinska: I have implemented a solution merging `parser` and `tagger` into the `parser` argument (e.g `nltk_wordpunct` as described above in commit edfdbcf . Looking forward to your thoughts!

rebeccabilbro:
Hi  @ekwiecinska &mdash; merging some CI fixes so that @bbengfort can go ahead and take a final look at this in a bit!

ekwiecinska:
Thanks @rebeccabilbro that's great, I can squash the develop merges if needed to tonight too.

ekwiecinska:
Great, thanks @rebeccabilbro !

yellowbrick/text/postag.py
bbengfort: I very much like the idea of merging the parser and tagger into a single argument - it simplifies things considerably and allows multiple levels of validation in the setter like you've added! Very nice work!

ekwiecinska:
Thanks for the review and comments @bbengfort, it was a nice issue to work on. Hopefully I'll work on more in the future 🎉.  Great, looking forward to v1.1!

bbengfort:
@ekwiecinska looking forward to your next PR!

PR #997, opened 2019-11-21
Tsne quick method

mchestnut91:
Addresses the TSNEVisualizer for #960

1. Added a `show` parameter to the quick method function signature, as well as to the docstring
2. Added an example usage of the `tsne` quick method to the docstring
3. Added an image comparison unit test
4. Added a baseline image for the image comparison test
5.  Added the `tsne` quick method to the API reference
6. Added a summary table to the top of `tsne.rst`
7. Added an example usage of the `tsne` quick method in `tsne.rst`

### Sample Code and Plot

from yellowbrick.text.tsne import tsne
from sklearn.feature_extraction.text import TfidfVectorizer
from yellowbrick.datasets import load_hobbies

corpus = load_hobbies()
tfidf = TfidfVectorizer()

X = tfidf.fit_transform(corpus.data)
y = corpus.target

tsne(X, y)

image

### TODOs and questions
I noticed that a few of the other TSNE tests are failing due to an Image Comparison Failure. I believe that this is caused by the fact that the TSNE visualization does not always appear identically when rendering. The points are always clustered the same way, but they are sometimes located in different spots on the screen. To get my quick method image comparison test to work, I increased the `tol` parameter to 50. I was hesitant to touch the other tests, but I would be happy to make this change to get the failing tests to pass if you would like!


yellowbrick/text/tsne.py
```diff
@@ -117,12 +134,15 @@ def tsne(
     # Fit and transform the visualizer (calls draw)
     visualizer.fit(X, y, **kwargs)
     visualizer.transform(X)
-    visualizer.finalize()
+
```
pdamodaran: Can you change the parameter `classes` in the quick method to `labels`.  This will ensure the docstring of the quick method matches that of the visualizer.

mchestnut91:
Hi @pdamodaran - thank you for taking the time to review this PR and for the helpful feedback! I believe that I addressed all of the issues that you identified - let me know if you see anything else that should be changed.

Thanks again!

mchestnut91:
Hi @pdamodaran. I noticed that my commit failed the Travis build. I looked into this and found that when I check out commit `0f29687`, which was done to merge the develop branch into my feature branch, that I am unable to complete the tests, as It stops running on `test_pcoords.py` every time. The tests run properly for me if I rollback this commit, but I cannot get them to run as is. I can look into this further when I have more time later this week, but wanted to check with you to see if you know what I could be doing wrong? Thank you again for your help!

PR #1000, opened 2019-12-05
Fix breaking tests and update baseline images

rebeccabilbro:
This PR repairs a number of tests that began failing after recent updates to our dependencies including scikit-learn, numpy, and matplotlib.

1. Updated baseline images and array checks in classifier tests that used `LogisticRegression`, which apparently had a number of bugs in it's scoring methodology that have been corrected in the [latest sklearn](https://scikit-learn.org/stable/whats_new/v0.20.html). _Note: also had to set the `solver` to `liblinear` in one of the `FeatureImportances` tests since without it I was getting really high RMSEs on Travis_
2. Added small similarity tolerances for a few image comparison tests that failed after baseline image updates. 
3. In the Discrimination Threshold visualizer, switched from `RadiusNeighborsClassifier` (which was recently refactored in sklearn and has a `predict_proba` attr now, so was failing to fail our `is_probabilistic` check) to a dummy deterministic "classifier".
4. Updated Elbow array checks to use more degrees of precision for the array comparison (these were failing before).
5. In CV scores tests, updated to 5-fold cross validation, which is the new sklearn default. Also changed to `StratifiedKFold` from `ShuffleSplit`, which is also the new default in sklearn.
6. Updated `ProjectionVisualizer` to `set_array` on the `ScalarMappable` since not having that was triggering at Matplotlib `TypeError` ("You must first set_array for mappable").
7. Updated `PCoords` and `RFECV` tests to speed them up a bit.

### CHECKLIST

- [x] _Have you run the unit tests using `pytest`?_ 😁 


tests/test_model_selection/test_importances.py
```diff
@@ -230,7 +230,9 @@ def test_multi_coefs_stacked(self):
         """
         X, y = load_iris(True)
 
-        viz = FeatureImportances(LogisticRegression(random_state=222), stack=True)
+        viz = FeatureImportances(
+            LogisticRegression(solver="liblinear", random_state=222), stack=True
+        )
```
rebeccabilbro: This is failing on Travis/Linux with huge RMSEs (like 89 😬), but fine on my local Mac and in Appveyor CI; currently manually investigating setting more of the params here on the LogisticRegression to see if I can resolve it that way...

tests/test_features/test_pcoords.py
```diff
@@ -173,8 +173,12 @@ def test_parallel_coordinates_quickmethod(self):
...
         X, y = load_occupancy(return_dataset=True).to_numpy()
 
-        #compare the images
-        visualizer = parallel_coordinates(X, y, show=False)
+        # Use only the first 100 samples so the test will run faster
+        X_t = X[:100]
+        y_t = y[:100]
+
+        # Compare the images
+        visualizer = parallel_coordinates(X_t, y_t, show=False)
```
rebeccabilbro: This is a bit 🎃 &mdash; would probably be more elegant to use the `sample` param of the quick method?

tests/test_model_selection/test_importances.py
rebeccabilbro: Looks like specifying `liblinear` worked, which seems a bit weird to me but 🤷‍♀ &mdash; I might need to dig in a bit deeper to the sklearn changelog to figure out what exactly they've changed in LR

pdamodaran:
Hi @rebeccabilbro - thanks so much for all the changes you made to update the tests to fix all the errors caused by changes in the dependencies.  This was a lot of work!  I was able to run the tests successfully on my machine.  Is there anything else I should check before approving this PR?

tests/test_features/test_pcoords.py
pdamodaran: I like this idea!  I tested this out and it works.  While I did this, I also noticed that the quick method was not passing the `shuffle` and `random_state` parameters to the visualizer.  I have made these changes on my branch.  Let me know if you would like me to push these changes.

rebeccabilbro: Good catch @pdamodaran! I'm happy to have you include those fixes with this PR or for you to open up a separate one &mdash; I suppose it might depend on how you and the other maintainers are coordinating Quick Method updates this semester, so I'll leave it to you to decide what's best!

pdamodaran: Thanks @rebeccabilbro!  I pushed my changes to this branch since the pcoords quick method updates had already been merged into develop.

yellowbrick/features/projection.py
```diff
@@ -286,6 +286,8 @@ def finalize(self):
             if self.colorbar:
                 if self.projection == 3:
                     sm = plt.cm.ScalarMappable(cmap=self._colors, norm=self._norm)
+                    # Avoid MPL TypeError: "You must first set_array for mappable"
+                    sm.set_array([])
```
bbengfort: I've confirmed that the scalar values are correct for the concrete dataset, e.g. that the range of the labels in the color bar matches with this call to `set_array` and without. I've only been able to get the `TypeError` once, so it's not reproducible for me generally. 

I think we should move forward with this PR but make an issue to dive into this deeper. Creating colorbars without data seems to be a debated feature in matplotlib:

https://github.com/matplotlib/matplotlib/issues/3644

It might be as simple that we need to set the range in the `ScalarMappable` and keep the minimum and maximum of the data around so that we don't have to keep the data. It also might be as simple as us being able to use `self.fig` instead of `plt`. 

Either way - this PR is to important to block as it is blocking all of our other PRs.

pdamodaran: Thanks again @rebeccabilbro for all of your hard work on this and taking care of this so quickly and @bbengfort for adding your review.  I will go ahead and merge this in.

PR #1015, opened 2020-02-06
precision_recall_curve quick method updates

bbengfort:
Refactors the `precision_recall_curve` quick method to the new quick method signature and verifies that it works correctly. Tasks:

- [x] implement the new train/test split mechanism

Related to #960


yellowbrick/classifier/prcurve.py
```diff
@@ -434,14 +434,10 @@ def _get_y_scores(self, X):
...
     model,
-    X,
-    y,
+    X_train,
+    y_train,
     X_test=None,
     y_test=None,
-    train_size=0.8,
-    random_state=None,
-    shuffle=True,
```
bbengfort: Removed, RE: new train test split mechanism

tests/requirements.txt
```diff
@@ -15,6 +15,6 @@ coverage>=4.5.2
 # Optional Testing Dependencies
 nltk>=3.2
 # spacy>=2.0.18
-pandas>=0.20
+pandas>=0.20,<1
```
bbengfort: Thanks for the intermediate fix on this! Sounds like we'll take a closer look at this in #1016

tests/test_classifier/test_prcurve.py
```diff
@@ -353,9 +352,10 @@ def test_quick_method(self):
             iso_f1_curves=True,
             ap_score=False,
             random_state=2,
+            show=False
```
bbengfort: I think we can remove  the `random_state` param now that it's been removed from the quick method

PR #1019, opened 2020-02-07
Make residuals_plot quick method primetime

bbengfort:
Refactors the residuals_plot quick method to the new quick method signature and verifies that it works correctly. Tasks:

- [x] implement the new train/test split mechanism

Related to #960


tests/test_regressor/test_residuals.py
```diff
@@ -343,11 +343,50 @@ def test_residuals_quick_method(self):
...
+        """
+        msg = "both X_test and y_test are required if one is specified"
+        with pytest.raises(YellowbrickValueError, match=msg):
+            residuals_plot(
+                Lasso(),
+                self.data.X.train,
+                self.data.y.train,
+                self.data.X.test,
+            )
+
```
rebeccabilbro: Thank you for adding these tests to ensure the edge cases of a user passing in only training data (which should work as normal), and a user passing in incomplete training and test splits (which should not)!

yellowbrick/regressor/residuals.py
```diff
@@ -790,15 +799,24 @@ def residuals_plot(
...
-    visualizer.finalize()
+    # Score the visualizer
+    if X_test is not None and y_test is not None:
+        viz.score(X_test, y_test)
+    elif X_test is not None or y_test is not None:
+        raise YellowbrickValueError(
+            "both X_test and y_test are required if one is specified"
+        )
+    else:
+        viz.score(X_train, y_train)
```
rebeccabilbro: Thank you for adding in this logic to protect against edge cases &mdash; this might be a useful snippet for others to use as a reference for other `ScoreVisualizer` quick methods!

PR #1024, opened 2020-02-07
ClassBalance quick method

rebeccabilbro:
Refactors the `class_balance` quick method to the new quick method signature and verifies that it works correctly (towards #960)

# TODOs specific to this PR
- [x] run tests in CI and watch something fail 😢 

# General YB PR checklist


docs/api/target/class_balance.rst
```diff
@@ -7,6 +7,14 @@ One of the biggest challenges for classification models is an imbalance of class
 
 There are several techniques for dealing with class imbalance such as stratified sampling, down sampling the majority class, weighting, etc. But before these actions can be taken, it is important to understand what the class balance is in the training data. The ``ClassBalance`` visualizer supports this by creating a bar chart of the *support* for each class, that is the frequency of the classes' representation in the dataset.
 
+=================   ==============================
+Visualizer           :class:`~yellowbrick.target.class_balance.ClassBalance`
+Quick Method         :func:`~yellowbrick.target.class_balance.class_balance`
+Models               Classification
+Workflow             Feature analysis, Target analysis, Model selection
```
bbengfort: Nice! I think we're pretty together a very good taxonomy of ML workflow that might be significant on its own.

tests/test_target/test_class_balance.py
```diff
@@ -231,9 +231,11 @@ def test_numpy_occupancy_compare(self):
 
     def test_quick_method(self):
         """
-        Test the quick method with
+        Test the quick method producing a valid visualization
```
bbengfort: Hehe, thanks for nipping that cliff hanger in the bud!

yellowbrick/target/class_balance.py
```diff
@@ -325,7 +337,11 @@ def class_balance(
 
     # Fit and transform the visualizer (calls draw)
     visualizer.fit(y_train, y_test)
-    visualizer.finalize()
```
bbengfort: What happens in the case when `y_test`is `None`? Does `ClassBalance` gracefully ignore it?

rebeccabilbro: Yes; I took another look at the tests and the one that we have for the quick method tests this case explicitly:
```python
    def test_quick_method(self):
        """
        Test the quick method producing a valid visualization
        """
        dataset = make_fixture(binary=False, split=False)

        viz = class_balance(dataset.y)

        assert isinstance(viz, ClassBalance)
        self.assert_images_similar(viz, tol=0.5)
```
So I added another quick method test for the other case, where there *are* train and test splits, just to be safe! Here's how it looks with splits:

test_quick_method_with_splits

and without:

test_quick_method

rebeccabilbro:
> Looking at AppVeyor, it looks like the test failed because it timed out after an hour - and it did timeout working on a class balance test. I'm hopeful we can simply restart that one job to get it to pass (or, rather, when we merge develop to update this PR that the tests will run and pass). If it doesn't pass and we can identify which test it's getting hung on, we mind as well simply mark that as skipped -- there are no changes in this PR that should produce that behavior!

Yeah unfortunately this behavior was not resolved @bbengfort ; after rerunning, it hangs on the first quick method test on Appveyor/Miniconda indefinitely until it times out. I have no idea since I don't think anything on our end changed on that test. I've marked that one test skipped for now, which hopefully resolves things in the short term (CI is running now), but we may want to revisit...

rebeccabilbro:
Weirdl

tests/test_target/test_class_balance.py
```diff
@@ -213,6 +213,9 @@ def test_pandas_occupancy_compare(self):
         # oz.finalize()
         self.assert_images_similar(oz, tol=0.5)  # w/o tol fails with RMS 0.433
 
+    @pytest.mark.xfail(
+        IS_WINDOWS_OR_CONDA,
+        reason="test stopped working on Miniconda, hangs indefinitely. See #1024")
```
rebeccabilbro: @bbengfort weirdly it seems like this is the test that is causing the problems. No idea why that would be

PR #1035, opened 2020-02-09
Update Classification Report quick method

Kautumn06:
As part of Issue #960
- Added `show` parameter to the `classification_report` quick method
- Updated quick method test
- Updated the docs to include an example using the quick method


tests/test_classifier/test_classification_report.py
```diff
@@ -189,6 +189,7 @@ def test_numpy_integration(self):
             "f1": {"unoccupied": 0.9800031994880819, "occupied": 0.9366447034972124},
         }
 
+    @pytest.mark.xfail(sys.platform == "win32", reason="images not close on windows")
```
rebeccabilbro: Usually we only add an `xfail` or `skip-if` decorator if the test has already run through CI and we have seen the specific failure case (e.g. it didn't work on a Windows virtual machine on Appveyor, or it didn't work on a Linus virtual machine using Miniconda on Travis). For now let's remove this decorator for now and let the test run as written.

tests/test_classifier/test_classification_report.py
```diff
@@ -203,11 +204,18 @@ def test_quick_method(self):
...
         _, ax = plt.subplots()
         model = DecisionTreeClassifier(random_state=19)
-        classification_report(model, X, y, ax=ax, random_state=42)
+        visualizer = classification_report(
+            model, X_train, y_train, X_test, y_test, ax=ax, show=False
+        )
 
-        self.assert_images_similar(ax=ax, tol=25.0)
+        assert isinstance(visualizer, ClassificationReport)
+        self.assert_images_similar(visualizer, tol=5.0)
```
rebeccabilbro: Can you also please remove the `tol=5.0` here until we know that we need a tolerance level to pass the image comparison test?

docs/api/classifier/classification_report.rst
```diff
@@ -5,6 +5,13 @@ Classification Report
 
 The classification report visualizer displays the precision, recall, F1, and support scores for the model. In order to support easier interpretation and problem detection, the report integrates numerical scores with a color-coded heatmap. All heatmaps are in the range ``(0.0, 1.0)`` to facilitate easy comparison of classification models across different classification reports.
 
+=================   =================
+Visualizer           :class:`~yellowbrick.classifier.ClassificationReport`
+Quick Method         :func:`~yellowbrick.classifier.classification_report`
+Models               Classification
+Workflow             Model evaluation
+=================   =================
```
rebeccabilbro: Thank you for adding this summary to the docs! It looked like the links weren't working so I doublechecked; looks like this should be instead:
```
=================   =================
Visualizer           :class:`~yellowbrick.classifier.classification_report.ClassificationReport`
Quick Method         :func:`~yellowbrick.classifier.classification_report.classification_report`
Models               Classification
Workflow             Model evaluation
=================   =================
```

docs/api/classifier/classification_report.rst
```diff
@@ -55,10 +62,44 @@ The metrics are defined in terms of true and false positives, and true and false
...
+    # Create the training and test data
+    tscv = TimeSeriesSplit()
+    for train_index, test_index in tscv.split(X):
+        X_train, X_test = X.iloc[train_index], X.iloc[test_index]
+        y_train, y_test = y.iloc[train_index], y.iloc[test_index]
+
+    # Instantiate the visualizer
+    visualizer = classification_report(
+        GaussianNB(), X_train, y_train, X_test, y_test, classes=classes, support=True
+    )
```
rebeccabilbro: This is a great way of showing that all of the original parameters of the class work with the quick method too &mdash; looks like the image turns out identical!

Kautumn06:
Hi @rebeccabilbro, thank you for your thoughtful notes! I've fixed the links in the docs and updated the tests with your suggestions. However, if there's anything else you need, please just let me know!

yellowbrick/classifier/classification_report.py
```diff
@@ -385,15 +396,24 @@ def classification_report(
         **kwargs
     )
 
-    # Create the train and test splits
-    X_train, X_test, y_train, y_test = train_test_split(
-        X, y, test_size=test_size, random_state=random_state
-    )
-
     # Fit and transform the visualizer (calls draw)
```
rebeccabilbro: @Kautumn06 taking a look at the test failures on Travis &mdash; looks like now that we've removed the splitting step from the quick method, we need to delete the unused import from the top of this file (`yellowbrick/classifier/classification_report.py`). Would you mind removing the line that reads `from sklearn.model_selection import train_test_split`?

tests/test_classifier/test_classification_report.py
```diff
@@ -203,11 +203,18 @@ def test_quick_method(self):
...
         _, ax = plt.subplots()
         model = DecisionTreeClassifier(random_state=19)
-        classification_report(model, X, y, ax=ax, random_state=42)
+        visualizer = classification_report(
+            model, X_train, y_train, X_test, y_test, ax=ax, show=False
+        )
 
-        self.assert_images_similar(ax=ax, tol=25.0)
+        assert isinstance(visualizer, ClassificationReport)
+        self.assert_images_similar(visualizer)
```
rebeccabilbro: @Kautumn06 took a look at the image comparison failure on Travis and it looks like we need to add an image comparison tolerance to line 217. Since the Travis Miniconda tests are failing with `yellowbrick.exceptions.ImageComparisonFailure: images not close (RMS 11.995)`, I propose that we try updating this line to be: 

```python
self.assert_images_similar(visualizer, tol=12)
```

... and we'll see if that works! 🤞 🙏 😅

Kautumn06: No problem @rebeccabilbro! I pushed the changes so fingers crossed! However, if Travis is still unhappy, just let me know what else you'd like me to try!

PR #954, opened 2019-08-22
[WIP] Added features to Decomposition.py

naresh-bachwani:
This PR fixes #316. I added the following features :

1. Added a parameter for showing cumulative explained variance. 
2. Added a parameter which takes percentage of variance needed in the components and shows corresponding principal components.

### Sample Code and Plot

```python
from yellowbrick.features.decomposition import ExplainedVariance

from sklearn.datasets import load_digits
digits = load_digits()
X = digits.data

viz = ExplainedVariance(cumulative=True, cutoff=85)
viz.fit(X)
viz.transform(X)
viz.poof("Varaince.png")
```

abc

Still to do:
- [ ] Add more features.
- [ ] Add tests.
- [ ] Add documentation

More Todos:

- [x] _Is the commit message formatted correctly?_
- [ ] _Have you noted the new functionality/bugfix in the release notes of the next release?_

- [x] _Included a sample plot to visually illustrate your changes?_
- [ ] _Do all of your functions and methods have docstrings?_
- [ ] _Have you added/updated unit tests where appropriate?_
- [ ] _Have you updated the baseline images if necessary?_
- [ ] _Have you run the unit tests using `pytest`?_
- [ ] _Is your code style correct (are you using PEP8, pyflakes)?_
- [ ] _Have you documented your new feature/functionality in the docs?_

- [ ] _Have you built the docs using `make html`?_

PR #1037, opened 2020-02-11
[WIP] Explained Variance

### LINKED ISSUE #316: Finish decomposition explained variance visualizer
Thanks to @georgerichardson we have an explained variance visualizer for PCA decomposition. This needs to be wrapped up as follows:

- [ ] expressiveness for other explained variances 
- [ ] labeling and key values features
- [ ] tests 
- [ ] documentation 

See #238

bbengfort:
This PR fixes #316 and extends #954 to get it wrapped up. 

1. Added documentation and tests
2. Refactored the quick method according to the quick method sprint

### Sample Code and Plot

_If you are adding or modifying a visualizer, PLEASE include a sample plot here along with the code you used to generate it._

### TODOs and questions

Still to do:

- [ ] Integrate #954 
- [ ] Update visualizer
- [ ] Write documentation
- [ ] Add better tests

### CHECKLIST

### DIFF
--- a/docs/api/features/explained_variance.rst
+++ b/docs/api/features/explained_variance.rst
@@ -1,0 +1,54 @@
+.. -*- mode: rst -*-
+
+Explained Variance
+==================
+
+=================   =================
+Visualizer           :class:`~yellowbrick.features.explained_variance.ExplainedVariance`
+Quick Method         :func:`~yellowbrick.features.explained_variance.explained_variance`
+Models               Decomposition
+Workflow             Feature Engineering
+=================   =================
+
+.. plot::
+    :context: close-figs
+    :alt: Explained variance quick method on the credit dataset
+
+    from yellowbrick.datasets import load_credit
+    from yellowbrick.features import ExplainedVariance
+
+    # Specify the features of interest and the target
+    X, _ = load_credit()
+
+    # Instantiate the visualizer, fit and transform the data
+    oz = ExplainedVariance()
+    oz.fit_transform(X)
+    oz.show()
+
+
+Quick Method
+------------
+
+The same functionality above can be achieved with the associated quick method ``explained_variance``. This method will build the ``ExplainedVariance`` visualizer with the associated arguments, fit it, then (optionally) immediately show it.
+
+.. plot::
+    :context: close-figs
+    :alt: Explained variance quick method on the concrete dataset
+
+    from yellowbrick.datasets import load_concrete
+    from yellowbrick.features import explained_variance
+
+    # Specify the features of interest and the target
+    X, _ = load_concrete()
+
+    # Determine the optimal number of components
+    oz = explained_variance(X)
+
+
+API Reference
+-------------
+
+.. automodule:: yellowbrick.features.explained_variance
+    :members: ExplainedVariance, explained_variance
+    :undoc-members:
+    :show-inheritance:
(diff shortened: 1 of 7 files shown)


bbengfort:
Resources:

- https://jakevdp.github.io/PythonDataScienceHandbook/05.09-principal-component-analysis.html
- https://towardsdatascience.com/an-approach-to-choosing-the-number-of-components-in-a-principal-component-analysis-pca-3b9f3d6e73fe
- https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html
- https://matplotlib.org/3.1.1/gallery/lines_bars_and_markers/fill_between_demo.html

bbengfort:
@naresh-bachwani prototype version:

Explained Variance Prototype

naresh-bachwani:
Hey there @bbengfort,
The above plot looks quite good! Few features that we can add:
1. We have a parameter `cutoff` which takes in the threshold value and returns the number of component corresponding to it.  We can also have a reverse scenario where we can pass a parameter `n_features` which returns the explained variance by that number of features. Although, I am not sure if we have a use case corresponding to this.
2. Also, In the above example, we do not get the exact number of features(it is somewhere between 20 and 30). It would be better if we can provide ticks corresponding to the cutoff line.

BrandonKMLee:
Questions: Is 85% variance some kind of magical number similar to p=0.05, or "rule of thumb" correlation coefficients or effect size?
|  | Correlation | Cohen's d | Angular | p-value | Odds Ratio |
|---|---|---|---|---|---|
| Very Strong | > 0.8 | > 2 | < 15° | < 0.001 | > 10 |
| Strong | > 0.6 | > 1.2 | < 30° | < 0.005 | > 3 |
| Moderate | > 0.4 | > 0.8 | <45° | < 0.01 | > 1.5 |
| Weak | > 0.2 | > 0.5 | < 60° | < 0.05 | > 1.2 |
| Very Weak | > 0.1 | >0.2 | < 75° | < 0.10 | > 1.0 |
| Negligible | > 0 | > 0 | = 90° | < 1 | = 1 |

Reference: https://archive.fo/Bi6yF https://archive.fo/ourlZ https://archive.fo/hVKdt https://archive.fo/jQOX8 https://archive.fo/Xh9xk

bbengfort:
@BrandonKMLee - yes, exactly right, it's just a rule of thumb about how many components are needed to have a "very strong" correlation or "enough explained variance" similar to the table you provided or a p value of 0.05.

gregparkes:
@bbengfort Thanks for the issue mention.

Looking at the above plot (which is pretty good) I would mention a few things.

1. I'm not convinced about plotting both explained var and cumulative variance. From my experience I have only ever plotted cumulative explained variance as generally it's easier to visually gauge the rate at which you approach 1. A steep curve is good, flat curve is bad. It also opens up things like area-under-curve.
2. Do we want this as a percentage (0-100) instead of (0-1)? People like percentages, or failing this we want to reference the fact that the value is an explained variance **ratio**/proportion/relative EV and not just "explained variance". These values have been normalized to 1 and not just the scaled singular values. 
3. I'm happy with a default of 85% choice for cumulative variance assuming this parameter can be changed

BrandonKMLee:
Some alternative ideas:
- Visuals for K1 or Kaiser criterion https://en.wikipedia.org/wiki/Exploratory_factor_analysis#Kaiser_criterion
- The "broken stick" method https://blogs.sas.com/content/iml/2017/08/02/retain-principal-components.html

PR #1050, opened 2020-03-15
Fix Travis Errors Related to switching method for making new directories in the Image Comparison Class

lwgray:
Fix Travis Errors Related to:

1.  Implement new way to create directories in Image Comparision Class
    We were getting Travis errors attributed to the fact that the current matplotlib.cbook.mkdir function was deprecated in MLP 3.0 To address this, the MPL function was changed to os.makedirs
2. Secondarily, we were getting an error stating that dataframes do not have a method "as_matrix". This is because "as_matrix()" has been deprecated.

This PR fixes #issue_number _(If you are fixing a bug)_ which reported a bug that caused a problem to occur when users...
Partially fixes #916 
_(or if you are introducing a new feature)_ which requested a feature to allow the user to...

I have made the following changes:

1. changed matplotlib.cbook.mkdir() to os.makedirs()
2. changed "df.as_matrix()" to "df.values" in contrib's scatter and boundaries visualizers
3.

### Sample Code and Plot

_If you are adding or modifying a visualizer, PLEASE include a sample plot here along with the code you used to generate it._

### TODOs and questions
What do I need to do to resolve the 2 image comparison failures?

1. Get rid of grid
2.  Increase Tolerance

Still to do:

- [ ] 
- [ ]
- [ ]

Questions for the @DistrictDataLabs/team-oz-maintainers:

- [ ]
- [ ]

### CHECKLIST

- [ ] _Is the commit message formatted correctly?_
- [ ] _Have you noted the new functionality/bugfix in the release notes of the next release?_

- [ ] _Included a sample plot to visually illustrate your changes?_
- [ ] _Do all of your functions and methods have docstrings?_
- [ ] _Have you added/updated unit tests where appropriate?_
- [ ] _Have you updated the baseline images if necessary?_
- [ ] _Have you run the unit tests using `pytest`?_
- [ ] _Is your code style correct (are you using PEP8, pyflakes)?_
- [ ] _Have you documented your new feature/functionality in the docs?_

- [ ] _Have you built the docs using `make html`?_

PR #1053, opened 2020-03-29
Update baseline images following breaking changes in mpl 3.2.1

rebeccabilbro:
This PR updates the baseline images for 65 image comparison tests that began failing after the [most recent release of matplotlib](https://matplotlib.org/3.2.1/users/whats_new.html). It also includes the excellent bugfixes that @lwgray discovered in #1050, which removed our `cbook` dependency for directory creation and changed `as_matrix()` to `values` in several tests.

# TODO
- [x] see if these new images, generated on my Macbook Pro, will be tolerated by Travis Miniconda and Appveyor.
- [x] update our `tests/requirements.txt` to require `matplotlib>=3.2.1`? 

# Questions
- Are we ok with me also updating our `tests/requirements.txt` to require `matplotlib>=3.2.1`? The changes made to the baseline images in this PR will cause some tests to fail if run locally with the older versions of MPL currently allowed in `tests/requirements.txt` 

# Done


lwgray:
Updates to the MPL api in 3.2.1 has also lead to behavioral change that might account for test failures.  The api changes can be found here: [api changes]( https://matplotlib.org/3.2.1/api/api_changes.html)

lwgray:
@rebeccabilbro I don't think we should have >= 3.2.1. If I understand correctly, it might be easier to peg the test to a specific version of matplotlib.  This would allow deliberate upgrading of matplotlib to the latest release thus more intentional wrangling of the inevitable test failures.  We would just need reminders.

bbengfort:
To comment on the question of requiring tests to have `matplotlib>=3.2.1` - I'm certainly ok with this, and really I think we may be reaching a point (especially given the acceleration of matplotlib development) that we deprecate support for < matplotlib 3.1 even in the main library. We do have some tests that are ignored if there isn't a specific version of matplotlib, but if we increase our matplotlib dependency then we can simply remove those tests. 

In terms of pegging the dependency to `matplotlib==` in tests, I certainly understand the motivation and I could be convinced to do this; but at first glance, I'm uncomfortable with it. The issue is that CI is supposed to catch these extremely dramatic changes in the latest versions of our dependencies that might impact our users; for example if the matplotlib API changes in a way that causes our multi-axes figures to render incorrectly or cause some misinterpretation of the visual diagnostic tool.  In that sense, the CI forcing us to re-evaluate our dependencies is exactly what is supposed to be happening, especially since new installs will install the latest version. 

On the other hand, if we don't have the bandwidth to deal with these changes and actively support breaking changes like this … then maybe we should fix the dependencies? Honestly, if we go this route, I'd want to consider going even further and consider closing all active development on YB and making the repo read-only? While that seems extreme, it's worth discussing because problems like this will come up pretty routinely even with the deps fixed, and I'm worried about having upset users because it doesn't work exactly the way the expect.

bbengfort:
@rebeccabilbro I'll review this PR tomorrow when I get a chance if that's ok!

lwgray:
@bbengfort I vote that we don't pause development.  It isn't that big of a chore to fix Travis test failures.  In fact, It is probably a necessary exercise for all contributors to participate in.  

My previous support for pegging to a specific version was because:I thought it would be easier for everyone else and didn't fully understand the importance of CI/CD and didn't take into account how it would affect the end-user.  With that being said, I reneg my support for pegging to a specific version

tests/base.py
```diff
@@ -287,9 +287,8 @@ def _image_path(self, root):
         imgdir = os.path.join(root, self.test_module_path)
 
         # Create directory if it doesn't exist
-        # TODO: remove dependency on mpl.cbook
         if not os.path.exists(imgdir):
-            mpl.cbook.mkdirs(imgdir)
+            os.makedirs(imgdir, mode=0o777, exist_ok=True)
```
bbengfort, line 291: This feels much better to me anyway - thank you for making this update!

tests/test_contrib/test_classifier/test_boundaries.py
```diff
@@ -390,7 +390,7 @@ def test_quick_method(self):
         data = datasets.load_iris()
         feature_names = [name.replace(" ", "_") for name in data.feature_names]
         df = pd.DataFrame(data.data, columns=feature_names)
-        X = df[["sepal_length_(cm)", "sepal_width_(cm)"]].as_matrix()
+        X = df[["sepal_length_(cm)", "sepal_width_(cm)"]].values
```
bbengfort, line 393: Great catch!

rebeccabilbro: all credit to @lwgray on this one!

PR #1059, opened 2020-04-12
Add support for sample_weight in KElbowVisualizer

### LINKED ISSUE #1057: Support for sample_weight parameter in KElbowVisualizer
KMeans and MiniBatchKMeans both have an optional `sample_weight` parameter in their `fit` methods. We should be able to pass this parameter to KElbowVisualizer.

Currently, supplying sample weights to `KElbowVisualizer` has no effect on the result.

**Examples**
N/A

Express50:
This PR closes #1057 by passing `kwargs` when calling `estimator.fit` in KElbowVisualizer.

### Sample Code and Plot

N/A

### CHECKLIST

- [ ] _Have you built the docs using `make html`?_


rebeccabilbro:
Hi @Express50 and thank you so much for identifying this bug in #1057 and submitting this PR to address it; what a nice fix! I think this is good to go in terms of merging in, though I did have one question for you, which is below. Let me know your thoughts.

Each of our `Visualizers` has a corresponding quick method, which is a one-liner designed to fit and render the plot in a single line. Our quick method for `KElbowVisualizer` is [here](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/cluster/elbow.py#L414-L506). As you'll see, it also takes in parameters including **kwargs, though it currently passes those kwargs only to init &mdash; what this means is that your fix won't manifest when users use the one-line version of this `Visualizer`. Do you think this is a feature that should also work in that case? If so, we might add something simple to the quick method, such as a little logic to detect any KMeans- or MiniBatch- specific kwargs and pop them off so they can be passed into `fit`.

Express50:
Thanks @rebeccabilbro!

I'm wondering if it wouldn't be better to simply pass the kwargs into the `fit` call inside the quick method? Looking at the other visualizers there's some inconsistency with regards to where those kwargs should go:

- In [confusion_matrix.py](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/classifier/confusion_matrix.py#L473) they are passed to both the `init` and `fit` methods
- In [validation_curve.py](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/model_selection/validation_curve.py#L416) they are passed to the `show` method only

Perhaps the kwargs should just be passed to all the methods that can support them?

rebeccabilbro:
Thanks for the quick response @Express50! 

Yes, there are definitely some variations in our API across the different visualizers (it's surprisingly tough to keep things consistent over time and contributors! 😅), including with how the quick methods are implemented, and the examples you've identified are totally valid (in fact, I'm pretty sure the ValidationCurves quick method has a bug... 🐞). 

I think our main goal in this case is stability &mdash; so that no matter how things change on the scikit-learn side (e.g. if they introduce a new kwarg for KMeans, or even some new centroidal clustering algorithm!), our quick method will reliably and correctly consume the sklearn-specific kwargs on `fit` (as opposed to the Yellowbrick kwargs like title name or plot size). For that reason, I propose we do something akin to what we've done [here in the TSNE Visualizer](https://github.com/DistrictDataLabs/yellowbrick/blob/c725a7c2e96fa19f70724938d239e35e09948b32/yellowbrick/text/tsne.py#L238-L241).

Express50:
Yeah that looks like a good approach. I will work on this.

Express50:
@rebeccabilbro, taking another look at the TSNE implementation, there is a slight difference to the current case, since the `sample_weight` parameter is not actually a parameter of the (MiniBatch-)KMeans model, but a parameter of the `fit()` method. Hence, we wouldn't be able to pull out this parameter using `get_params()`.

One possible way to overcome this, is to pull out `fit()` parameters using [the signature](https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object). 
```python
In [8]: inspect.signature(km.fit).parameters
Out[8]:
mappingproxy({'X': <Parameter "X">,
              'y': <Parameter "y=None">,
              'sample_weight': <Parameter "sample_weight=None">})
```
This might be a bit overkill as I'm not sure how likely scikit are to change the `fit()` method signatures?

Let me know what you think.

rebeccabilbro:
I think that's an excellent solution! And `inspect` works for Python >=3.4, right? Very clever!

Express50:
I added the logic for pulling out `fit`-related kwargs. 

Not quite sure about the tests... Since we're not consuming any `init` args in KElbowViz, should I instead test that the scikit `fit` method does not consume yellowbrick named params (size, title, color)?

yellowbrick/utils/helpers.py
```diff
@@ -185,6 +186,28 @@ def is_monotonic(a, increasing=True):
...
+    parameters : list
+        A list of keyword-only parameter names for method.
+    """
+    signature = inspect.signature(method)
+    parameters = [p for p in signature.parameters.values()
+                    if p.name != 'self' and p.kind != p.VAR_KEYWORD]
+    
+    return sorted([p.name for p in parameters])
+
+
```
rebeccabilbro, line 210: Thank you so much for making this a utility so that we'll be able to leverage it for other Visualizers! This looks great; I think the only thing we'll want to add is some test coverage for this function to  `tests/test_utils/test_helpers.py`, e.g. things like: 

```python
    @pytest.mark.parametrize(
        "estimator",
        [SVC, SVR, Ridge, KMeans, RidgeCV, GaussianNB, MiniBatchKMeans, LinearRegression],
        ids=["SVC", "SVR", "Ridge", "KMeans", "RidgeCV", "GaussianNB", "MiniBatchKMeans", "LinearRegression"],
    )
    def test_get_param_names(self, estimator):
        """
        Assert we successfully extract the parameters from sklearn estimators
        """
        assert "sample_weight" in get_param_names(estimator)
```

Perhaps we might also be slightly safer off if we wrap the call to `inspect.signature` in a try/except, in case the thing passed into the function isn't a method (e.g. if the method were renamed or mistyped).

However, these go beyond the scope of what you were initially going for, which you have absolutely achieved &mdash; so let me know if you'd prefer me to add these and push them to your branch!

yellowbrick/cluster/elbow.py
```diff
@@ -487,6 +487,15 @@ def kelbow_visualizer(
...
     """
+    klass = type(model)
+
+    # figure out which kwargs correspond to fit method
+    fit_params = get_param_names(klass.fit)
+
+    fit_kwargs = {
+        key: kwargs.pop(key) for key in fit_params if key in kwargs
+    }
+
```
rebeccabilbro, line 498: This looks great; just what I had in mind!

yellowbrick/cluster/elbow.py
```diff
@@ -496,7 +505,7 @@ def kelbow_visualizer(
         locate_elbow=locate_elbow,
         **kwargs
     )
-    oz.fit(X, y)
+    oz.fit(X, y, **fit_kwargs)
```
rebeccabilbro, line 508: I think the best way to test this would be to add two tests to `tests/test_cluster/test_elbow.py`, one to make sure we're using the kwargs properly now, and one to make sure that the quick method correctly sorts the kwargs, e.g.:

```python

def test_sample_weights(self):
    """
    Test that passing in sample weights correctly influences the clusterer's fit
    """
    X, y = load_nfl()

    visualizer = KElbowVisualizer(
        MiniBatchKMeans(random_state=37), k=5, timings=False, locate_elbow=False
    )
    visualizer.fit(X, sample_weight=y)  # is this right?
    visualizer.finalize()
    expected = np.array(
        [
            # fill in ...
        ]
    )
    assert_array_almost_equal(visualizer.k_scores_, expected)


def test_quick_method_params(self):
    """
    Test the quick method correctly consumes the user-provided parameters
    """
    X, y = load_nfl()
    custom_title = "My custom title"
    model = KMeans(3, random_state=13)
    oz = kelbow_visualizer(model, X, sample_weight=y, title=custom_title)
    assert oz.title == custom_title
```

Express50: Thanks for this! The quick method and helper tests look pretty straightforward, and I can work on them.

Regarding sample weights test: I was thinking about something like this, but I'm not sure how to craft the data in a way to make the test pass. In this case the sample_weight should be a vector describing the importance of each observation when computing centroids. My original idea was to craft a dataset with, say K=5 clusters, but tweak the sample_weights in a way that we get a different number of clusters according to the elbow. Not sure what is the best approach. Let me know what you think!

rebeccabilbro: That sounds great!

yellowbrick/utils/helpers.py
```diff
@@ -185,6 +186,32 @@ def is_monotonic(a, increasing=True):
...
+    Returns
+    -------
+    parameters : list
+        A list of keyword-only parameter names for method.
+    """
+    try:
+        signature = inspect.signature(method)
+    except (ValueError, TypeError) as e:
+        raise e
```
rebeccabilbro, line 207: Thank you for adding this!

tests/test_utils/test_helpers.py
```diff
@@ -173,6 +173,24 @@ def test_check_fitted(self):
...
+        Assert we successfully extract the parameters from sklearn estimators
+        """
+        assert "sample_weight" in get_param_names(estimator.fit)
+
+    def test_get_param_names_type(self):
+        """
+        Assert a type error is raised when passing a non-method
+        """
+        with pytest.raises(TypeError):
+            get_param_names('test')
```
rebeccabilbro: Thank you for adding test coverage for the new util!

tests/test_cluster/test_elbow.py
```diff
@@ -400,6 +395,32 @@ def test_timings(self):
...
+            KMeans(random_state=seed), k=(2, 12), timings=False
+        )
+        visualizer.fit(X)
+        assert visualizer.elbow_value_ == 5
+
+        # weights should push elbow down to 4
+        weights = np.concatenate([np.ones(5) * 0.0001, np.ones(120)])
+
+        visualizer.fit(X, sample_weight=weights)
+        assert visualizer.elbow_value_ == 4
```
rebeccabilbro, line 422: Excellent test &mdash; also a good, clear example of why the sample weights can have such a big impact on the clusters! 

_Note: Other than autoformatting, I've removed the calls to `finalize` here, since that just adds the finishing touches to the visualizer, like title, etc, so we don't need it in this case & it makes the tests a teensy bit faster to run if we can avoid it!_

Express50:
Thank you for your very helpful feedback along the way @rebeccabilbro! It was great to work on this!

PR #1092, opened 2020-08-05
Include license file in source distribution

melonhead901:
It would be helpful to include the license file in the source distribution


bbengfort:
@melonhead901 thank you for your interest in Yellowbrick. I agree that the license file should be included in the source distribution, and in fact, it already is by virtue of line 3 in the Manifest. I ran the following command:

```
$ python setup.py build sdist
```

And verified that LICENSE.txt is included in `dist/yellowbrick-1.1.tar.gz`. 

While your change certainly makes the inclusion of the LICENSE.txt file more explicit, and I'm fine with making the update for that virtue alone, I'm concerned that you found a source distribution without the LICENSE.txt file. Could you go into more detail about how you discovered this issue?

rebeccabilbro:
@melonhead901 could you let us know more detail about how you discovered a source distribution of Yellowbrick without the LICENSE.txt file?

melonhead901:
Hi - there wasn't a distribution without a license file. I just noticed while conda forge packaging (linked PR) that it wasn't included explicitly in the manifest (so there's an extra step to specify the license file in conda).

bbengfort:
@melonhead901 ok, thanks for making it easier to include Yellowbrick with conda forge!

PR #1098, opened 2020-09-18
Non-sklearn dependent estimator check

Yard1:
The `is_estimator` will return False if the passed object is not an instance of sklearn `BaseEstimator`. This causes issues with eg. https://github.com/rapidsai/cuml, which implements a sklearn API but has their own base class. The proposed change is to check if the object has two critical estimator methods - `fit` and `predict`.

### CHECKLIST

### DIFF
--- a/yellowbrick/utils/types.py
+++ b/yellowbrick/utils/types.py
@@ -20,8 +20,6 @@ Detection utilities for Scikit-Learn and Numpy types for flexibility
 import inspect
 import numpy as np
 
-from sklearn.base import BaseEstimator
-
 
 ##########################################################################
 ## Model Type checking utilities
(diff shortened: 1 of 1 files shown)


bbengfort:
See also #1099

bbengfort:
@Yard1 Thank you so much for your contribution and for your interest in using Yellowbrick with cuML - we're very excited about this possibility! While the change is seemingly simple, as you can see from the CI tests - it has a lot of pervasive consequences. Put simply, Yellowbrick relies on scikit-learn for more than just its method interface, but also relies on fitted attributes and hyperparameters. 

We've put a lot of thought into how to facilitate using cuML and other machine learning libraries more easily; the result is in #1103. Although we haven't completely tested cuML with the new wrapper - I think it should probably work. Would you be willing to explore if cuML works with Yellowbrick and look over the other PR?

Yard1:
@bbengfort Hah, apologies, I made this PR and forgot to check the CI when it was done, and later got caught up in many other things. Yes, I can see the issue now. I am happy to hear you have been working on a more comprehensive solution for this problem. I'll be happy to test it and report back, provided I can fit it into my schedule.

bbengfort:
@Yard1 I appreciate it! Thanks again for using Yellowbrick!

PR #1102, opened 2020-10-02
Add param to limit Feature Importances to top_n

mgarod:
WIP

This PR addresses a bullet point from issue #322 

1. Add parameter to FeatureImportances visualizer for `top_n` as well as documentation for it
1. Add validation to ensure that `N < |features|`
1. The FeatureImportances visualizer now trims its data to only the top_n or bottom_n, and discarding the rest

Still to do:

- [ ] Add tests
- [ ] Complete checklist below

- [ ] Is there place for `top_n` when `stacked=True`? I think no, so `top_n` is is effectively ignored when `stacked=True`.
- [ ] Are there any other combination of parameters that could conflict with top_n?

**Top N example:**
top_n

**Bottom N example:**
bottom_n

### CHECKLIST

- [ ] _Is the commit message formatted correctly?_

- [ ] _Have you built the docs using `make html`?_


bbengfort (commented):
@mgarod thank you so much for this PR and the excellent PR message and thoughtful way you're contributing in October! It looks like a good start so far - and I've made a few comments in the code. To answer your questions:

1. When `stacked=True` there is a `topn` context - namely the magnitude of the whole bar, e.g. you'll have to sum the absolute values of the feature importance and take the topn biggest (said another way, the top n most (or least) influential features across all classes) 
2. There is another parameter that will affect topn - `relative=False` More on this below. 

One of the challenges with `topn` is that you're actually looking for the biggest _absolute_ values of feature importance, particularly when it comes to coefficients (this doesn't really apply to probabilities or decision trees). E.g. what coefficients are having the most influence on their respective variables, either negatively or positively correlated. When `relative=True` all values are normalized so as to be positive, but when `relative=False`, some of your `topn` values may be negative. Hence, the slicing mechanism really needs to use argmax. This is kind of an edge case, but one that is critical to understanding what the relationship of a model to its features is. 

I do like your use of negative `topn` to give the smallest values, but I would interpret this to be "what are the n least informative features", e.g. to remove them. 

Sorry if this makes the issue a bit more complex than you were expecting, but I'm looking forward to seeing what kind of solution you come up with! 

Also a quick note on hacktoberfest - there is a possibility that we've opted out from it given the huge amounts of spam we've received in the past. If you're serious about participating for a t-shirt, let me know so we can try to figure it out.

yellowbrick/model_selection/importances.py
```diff
@@ -96,6 +96,10 @@ class FeatureImportances(ModelVisualizer):
         Keyword arguments that are passed to the base class and may influence
         the visualization as defined in other Visualizers.
 
+    top_n : int, default=None
+        Display only the top N results with a positive integer, or the bottom N
+        results with a negative integer. If None or 0, all results are shown.
```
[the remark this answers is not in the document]
mgarod: Will do!

yellowbrick/model_selection/importances.py
```diff
@@ -218,13 +224,24 @@ def fit(self, X, y=None, **kwargs):
...
             sort_idx = np.argsort(np.mean(self.feature_importances_, 0))
             self.features_ = self.features_[sort_idx]
             self.feature_importances_ = self.feature_importances_[:, sort_idx]
         else:
             sort_idx = np.argsort(self.feature_importances_)
+            if self.top_n:  # Keep only the top or bottom n examples
+                if self.top_n > 0:
+                    sort_idx = sort_idx[-self.top_n:]
+                else:
+                    sort_idx = sort_idx[:-self.top_n]
```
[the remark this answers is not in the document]
mgarod: I totally agree with the worry of confusion. I had hoped that the idea of using a negative number to grab from the back instead of the front would be simple since we see it all the time in python slicing. I'll ponder on some better nomenclature, and hope that the solution is not to have a separate `bottom_n` parameter entirely.

Speaking of nomenclature, would you prefer `top_n` or `topn`?

mgarod:
> Also a quick note on hacktoberfest - there is a possibility that we've opted out from it given the huge amounts of spam we've received in the past. If you're serious about participating for a t-shirt, let me know so we can try to figure it out.
I wasn't concerned with the shirt anyway! I'm just using hacktoberfest as a symbolic reminder to contribute.

mgarod:
Changed `top_n` to compute via absolute value.

In the example in documenation where `relative=False`, there is a feature with a very "large" negative value.
example in documentation

With this change, `top_n=4` and `top_n=-4` would be:
top4
top-4

In the bottom N example, the x-axis scale auto adjusts without knowing the top example's coefficient, so it looks deceiving.

Still working on the case where `stack=True`.

yellowbrick/model_selection/importances.py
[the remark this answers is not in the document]
bbengfort: Slicing is a very good point and also an excellent analogy. If we can't come up with a better name, I think a negative `top_n` is preferable to also having a `bottom_n` parameter. 

As for names, I do prefer `topn` slightly over `top_n` - however, I think this is a good place for you to make your mark on Yellowbrick, so please feel free to select your preference (let me know, and I'll be careful to use that arg from here on out).

bbengfort:
@mgarod good progress! Please let me know when you're ready for another review - until then I'll let you carry on.

yellowbrick/model_selection/importances.py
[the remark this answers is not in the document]
mgarod: I default to snake case, but will alter to `topn` to be consistent with `xlabel` also in `FeatureImportances`

mgarod:
Finished logic for `topn` when `stacked=True`. Added tests for image comparison. Added function to further tailor the plot title to include `Top` or `Bottom` when necessary.

Original stacked FI plot (`topn=None`)
iris_stacked

`topn=3`
iris_stacked_top3

`topn=-3` Note that features were selected by least magnitude (smallest by absolute value), then resorted by true value. If the FI were not resorted, the output plot would be features 3-0-1 instead of the nicer 3-1-0.
iris_stacked_bottom3

Will work on case when `relative=False` next.

mgarod:
I did not find the case where `relative=False` needs to be explicitly handled. Since importances are selected based on their magnitude/absolute value, it doesn't matter whether the numbers are real numbers (-inf, inf) or percentages [0,1].

Working on writing up something in the documentation for `topn`.

docs/api/model_selection/importances.rst
```diff
@@ -107,10 +107,54 @@ Taking the mean of the importances may be undesirable for several reasons. For e
     X, y = data.data, data.target
 
     model = LogisticRegression(multi_class="auto", solver="liblinear")
-    viz = FeatureImportances(model, stack=True, relative=False)
+    viz = FeatureImportances(model, stack=True, relative=False, topn=3)
```
bbengfort, line 110: What was your motivation to add this parameter to the stacked chart docs?

docs/api/model_selection/importances.rst
```diff
@@ -107,10 +107,54 @@ Taking the mean of the importances may be undesirable for several reasons. For e
...
     viz.fit(X, y)
     viz.show()
 
+Top and Bottom Feature Importances
+----------------------------------
+
+It may be more illuminating to the feature engineering process to identify the most or least informative features. To view only the N most informative features in a graph, use ``topn`` with a postive integer. To view the N least important, use a negative integer.
```
bbengfort, line 117: It would be great if we could mention the slicing metaphor that you proposed earlier. Perhaps:

```
To view only the N most informative features, specify the ``topn`` argument to the visualizer. Similar to slicing a ranked list by their importance, if ``topn`` is a postive integer, then the most highly ranked features are used, if ``topn`` is a negative integer, then the lowest ranked features are displayed instead. 
```

docs/api/model_selection/importances.rst
```diff
@@ -107,10 +107,54 @@ Taking the mean of the importances may be undesirable for several reasons. For e
...
+
+    # Title case the feature for better display and create the visualizer
+    labels = list(map(lambda s: s.title(), dataset.meta['features']))
+    viz = FeatureImportances(Lasso(), labels=labels, relative=False, topn=3)
+
+    # Fit and show the feature importances
+    viz.fit(X, y)
+    viz.show()
+
+When repeating the previous concrete dataset graph using ``topn=3``, we see that the three most informative features are splast, cement, and water. Although water has a negative coefficent, it is nonetheless informative to reveal the negative correlation associated with the feature.
```
bbengfort, line 139: Suggested edit:

Using `topn=3`, we can identify the three most informative features in the concrete dataset as `splast`, `cement`, and `water`. This approach to visualization may assist with _factor analysis_ - the study of how variables contribute to an overall model. Note that although `water` has a negative coefficient, it is the magnitude (absolute value) of the feature that matters since we are closely inspecting the negative correlation of water with strength. An alternative, `topn=-3` would reveal the least informative features in the model. This approach is useful to model tuning similar to [RFECV](https://www.scikit-yb.org/en/latest/api/model_selection/rfecv.html), but instead of automatically removing features, would allow you to identify the lowest-ranked features as they change in different model instantiations. In either case, if you have many features, using `topn` can significantly increase the visual and analytical capacity of your analysis. 


Note on the `RFECV` link, please use:

```
:doc:`RFECV <api/model_selection/rfecv>`
```

As the API link in rst, but you may have to ensure the path is relative if the docs do not build (the path given above is from the root).

bbengfort (approved):
@mgarod this is looking really great - thank you for such a complete and well thought out PR! I very much approve of the `np.argsort` and reduction methods you implemented. I did have to think a while about directly modifying `features_` and `feature_importances_` instead of masking them, but I do think this was the right approach. I'm really excited for this feature to be part of Yellowbrick!

Documentation and tests are included - are there any items remaining?

docs/api/model_selection/importances.rst
mgarod: That's a mistake from the editing process! Will eliminate.

docs/api/model_selection/importances.rst
mgarod: Good suggestion. Will integrate.

docs/api/model_selection/importances.rst
mgarod: Thank you for the domain knowledge!

mgarod:
@bbengfort Documentation and tests were the last items. Does there need to be a mention of this in a release notes draft? If so where is that?

Do you think the tests I added are superfluous? They are simple "draw a graph and compare it to an image" 4 times based on the combination of `stacked=True/False` and `topn=+/-`. It just feels messy, even though it may be considered good coverage?

bbengfort:
@mgarod I don't think the tests are superfluous, they capture all dimensions of the different parameters, which will protect us from regression. If you'd like, my advice would be to vary the tests a bit in other ways. For example, instead of using `iris` for all the tests, use `make_classification` and `make_regression` to set a specific number of features by importance - then use two different models; one that has `feature_importances_`, and one that has `coefs_` (the `coefs_` one would likely be the stacked version). You could then test specifically the underlying arrays that were filtered, which would give you more test surface area beyond just the image comparison - which to be honest is a little unstable with CI (you've already got decently high RMSE tolerances). 

However, let's just leave that as lessons learned for the next time. I'm going to go ahead and merge this in as it is now. If you'd like to update the tests we can do that in another PR, or you can update it when you get a chance to do another YB feature.

Thank you so much for contributing! I think this feature is going to be very useful moving forward, and you implemented it very well. I really appreciate your thoughtfulness and completeness with the documentation and the tests. This was an excellent PR, and in my opinion, worthy of several t-shirts!

PR #1144, opened 2021-01-15
Release candidate Hotfix 1.2.1

bbengfort:
**NOTE**: I do not intend to merge this PR into master, I wanted to open the PR to run our CI checks. We should close this PR when reviewed without merging, then merge the release into both `master` and `develop` with a tag in our normal release process. 

On December 22, 2020, scikit-learn released version 0.24 which deprecated the external use of scikit-learn's internal utilities such as `safe_indexing`. Unfortunately, Yellowbrick depends on a few of these utilities and must refactor our internal code base to port this functionality or work around it. To ensure that Yellowbrick continues to work when installed via `pip`, we have temporarily changed our scikit-learn dependency to be less than 0.24. We will update our dependencies on the v1.3 release when we have made the associated fixes.


rebeccabilbro (approved):
Thank you so much @bbengfort for creating this hotfix as an interim way of addressing the large influx of issues and stackoverflow posts we've been getting from users of Yellowbrick who are struggling with these new API changes in scikit-learn. 

Next we'll get to work on a fix that addresses the private/public API changes so that users will be able to make use of versions both before and after sklearn 0.24 (and possibly any additional changes coming in sklearn 1.0 that aren't yet announced), but this will make sure users have the best possible experience until we can get all those changes sorted out.

rebeccabilbro:
Note - the CI errors we're seeing appear to be relate to the scipy 1.6 issue I mentioned the other day. https://github.com/scikit-learn/scikit-learn/issues/19111

PR #1171, opened 2021-04-01
Fixes #1163 The  inability to place FeatureCorrelation Visualizations into subplots.

### LINKED ISSUE #1163: ax param ignored in feature_correlation quick method
**Describe the bug**
When using matplotlib subplots create 2 plots of feature correlations side by side, the ax parameter seems to be ignored  and two images are created.

image1
image2

**To Reproduce**

```python
import numpy as np
from sklearn import datasets
from yellowbrick.target.feature_correlation import feature_correlation

#Load the diabetes dataset
data = datasets.load_iris()
X, y = data['data'], data['target']

features = np.array(data['feature_names'])

_, axes = plt.subplots(ncols=2, figsize=(18, 6))

feature_correlation(X, y, labels=features, method='pearson', ax=axes[0])
feature_correlation(X, y, labels=features, method='mutual_info-classification', ax=axes[1])

plt.tight_layout()
```

**Expected behavior**
The expected behaviour would be the plots side by side in a single image.

 - OS: macOS
 - Python Version: Python 3.7.10 (Google Colab)
 - Yellowbrick Version: 1.3.post1

**Additional context**
I can confirm it happens also locally with `Yellowbrick 1.3.post1` running using command line and using jupyter notebooks

lwgray:
When using ax param in the FeatureCorrelation visualizer to specify an axes location within a matplotlib subplot, it would not plot in the correct position.  However, it would create an entirely new plot

To fix this, I changed the ax in the Super init so it was not explicitly setting it to None which was preventing axes from being specified

Before:
image

After:
image

Fixes #1163

### DIFF
--- a/yellowbrick/target/feature_correlation.py
+++ b/yellowbrick/target/feature_correlation.py
@@ -123,7 +123,7 @@ class FeatureCorrelation(TargetVisualizer):
         color=None,
         **kwargs
     ):
-        super(FeatureCorrelation, self).__init__(ax=None, **kwargs)
+        super(FeatureCorrelation, self).__init__(ax, **kwargs)
 
         self.correlation_labels = CORRELATION_LABELS
         self.correlation_methods = CORRELATION_METHODS


yellowbrick/target/feature_correlation.py
```diff
@@ -123,7 +123,7 @@ def __init__(
         color=None,
         **kwargs
     ):
-        super(FeatureCorrelation, self).__init__(ax=None, **kwargs)
+        super(FeatureCorrelation, self).__init__(ax, **kwargs)
```
bbengfort, line 126: Good catch; must have been a copy and paste error!

bbengfort (approved):
Thank you so much for digging into this - I know a one-line fix doesn't seem satisfying, but I know how much work you had to do to dive into the code, set up the test case, and trace the problem to its source. These bugs are the most frustrating but your efforts have made Yellowbrick a lot better! 

I'm rerunning the CI tests - I'm not sure why they failed since it wouldn't give the logs; I'm assuming its transient but I'll keep an eye on it so we can merge this and fix the bug.

Issue #1166, opened 2021-03-29
Argument to remove colorbar in classification plots


nateGeorge:
**Describe the solution you'd like**
Can we add an argument to remove the colorbar for the classification plots, like sklearn has plot_confusion_matrix() with the colorbar argument? This would be for at least the confusion matrix and classification reports, maybe more.

**Is your feature request related to a problem? Please describe.**
I consider the colorbar chartjunk since we already have the raw counts on the plot.

nateGeorge:
Currently this can be done with a slightly hacky method:

```python
from yellowbrick.classifier import ClassificationReport

f, ax = plt.subplots(1, 1)
viz = ClassificationReport(lr_model,
                           support=True,
                           classes=['no default', 'default'],
                           cmap='Blues')
viz.score(x_test, y_test)
f.delaxes(f.axes[0])
viz.show()
```

bbengfort:
Hi @nateGeorge thank you for using Yellowbrick; we'll certainly take your request under consideration and are more than open to PRs or other contributions. I would have suggested the `delaxes` mechanism in the meantime - thank you for adding it here in case anyone else has a similar question!

PR #1174, opened 2021-04-03
WIP: Give option to Remove colorbar from classification_report

lwgray:
This PR addresses #1166 a feature request to remove the colorbar from classification report.

I made the following changes:
1. I added `colorbar` as an argument
2. I added self.color_bar = color_bar in `__init__`
3. I set an if/else state to control the flow of the building of the colorbar based on if the argument was true

### CHECKLIST


tests/test_classifier/test_classification_report.py
```diff
@@ -312,3 +312,21 @@ def test_with_fitted(self):
...
+        viz.fit(self.binary.X.train, self.binary.y.train)
+        viz.score(self.binary.X.test, self.binary.y.test)
+
+        self.assert_images_similar(viz, tol=40)
+
+        assert viz.scores_ == {
+            "precision": {0: approx(0.7446808), 1: approx(0.8490566)},
+            "recall": {0: approx(0.8139534), 1: approx(0.7894736)},
+            "f1": {0: approx(0.7777777), 1: approx(0.8181818)},
+        }
```
bbengfort, line 332: I recommend that we remove this assertion since it has nothing to do with the colorbar test. The more we simplify our tests, the easier it is to maintain them in the long run!

bbengfort (changes requested):
@lwgray looks good, thank you so much for taking on the feature request so quickly. As always small changes build Yellowbrick into a much more robust library! Thank you for being so thorough and adding a test as well. 

Just one quick change required and one question. First the change - please also add the `color_bar` argument to the quick method and quick method docstring. 

Second, a question, is there a reason you used `color_bar` as the arg instead of `colorbar`?

tests/test_classifier/test_classification_report.py
lwgray: i removed it...:thumbsup:

lwgray:
@bbengfort there was no rhyme or reason with the underscore in color_bar, therefore I converted them all to `colorbar`  👍

lwgray:
@bbengfort  
Ready to :shipit:

bbengfort (approved):
@lwgray thank you for making those changes!

bbengfort:
> @bbengfort there was no rhyme or reason with the underscore in color_bar, therefore I converted them all to `colorbar` 👍

Ah ok - I thought it might have been a name conflict with the axes or something like that. I think `colorbar` makes the most sense since it mirrors the matplotlib command; thank you for making the change!

PR #1176, opened 2021-04-15
Updates to Matplotlib 3.4.1 and Numpy 1.20.2

pdamodaran:
This PR updates the test requirements to the latest versions of Matplotlib and Numpy

1. Updated Matplotlib to 3.4.1 and Numpy to 1.20.2
2. Updated baseline images for tests that failed due to minor differences in the images caused by the change to the latest Matplotlib version


bbengfort (approved):
@pdamodaran thank you so much for keeping on top of our dependency changes and updating our tests to reflect the latest and greatest matplotlib and numpy! I reviewed all of the baseline images and they look good. Also, thank you for cleaning up the baseline images where tests were being skipped!

PR #1178, opened 2021-05-17
Fix broken url to pytest fixtures

emarkou:

1. Removed broken URL 
2. replaced it with a working one

### CHECKLIST


rebeccabilbro (approved):
Great catch @emarkou and thank you so much for the fix!

PR #1184, opened 2021-07-06
Allow Numpy 1.20+

akx:
This follows up on #1176 to also allow Numpy 1.20+ for end users of the library.

The end user problem is that Pip's new dependency resolver can get very ornery when an user attempts to install a new Numpy alongside Yellowbrick; if one is unlucky, it'll spend a long time trying to find a constellation of (various data science) packages that would accept Numpy < 1.20 to satisfy the pin in Yellowbrick's requirements.

This places the onus of properly locking dependency versions (with e.g. `pip-tools` or `poetry`) on the end user, as it should be.

1. Removed the `<1.20` numpy pin from the `requirements.txt` files that had been untouched by #1176.

I've elided editing the changelog here, since it looks like there have been several other changes since 1.3.post1 anyway, and those need to be edited in too...

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,7 +2,7 @@
 matplotlib>=3.3
 scipy>=1.6.0
 scikit-learn>=0.24
-numpy>=1.16.0,<1.20
+numpy>=1.16.0
 cycler>=0.10.0
 
 # Documentation Dependencies
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 matplotlib>=2.0.2,!=3.0.0
 scipy>=1.0.0
 scikit-learn>=0.20
-numpy>=1.16.0,<1.20
+numpy>=1.16.0
 cycler>=0.10.0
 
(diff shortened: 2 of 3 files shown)


bbengfort:
Hi @akx thanks for the PR - I can't remember why we had numpy pinned this way; there was some issue that was causing the YB tests to fail. I've bumped the YB tests up to numpy 1.21 to see if the issue has been resolved; I'll run those tests locally then check if I can update your branch with the changes; if the tests work, we can merge this PR. Thanks for contributing to Yellowbrick!

bbengfort (approved):
@akx thanks for keeping an eye on the dependencies and submitting the PR!

akx:
@bbengfort Anytime! This actually came up with a @valohai customer who was having trouble with a requirements file pip's new dependency resolver was taking forever to solve :)

hosungs:
Hello, I'm wondering when this change will be released and available at the [pypi repo](https://pypi.org/project/yellowbrick/#history)? Should I file an issue for that?

@craigcitro FYI

PR #1187, opened 2021-07-10
Housekeeping

### LINKED ISSUE #707: Manage test warnings: 'Agg backend'

### LINKED ISSUE #1113: Housekeeping
- [x] rename master branch to main
- [x] create release issue template

@rebeccabilbro

### LINKED ISSUE #1127: Add .gitattributes to control line ending whitespace
Add a `.gitattributes` file to set per-repository defaults for line endings. This is in response to @lwgray's issue with whitespace changes when he moved to a windows machine. Depending on how `git` is configured it may checkout files with a `\r\n` (Windows) line ending rather than a `\n` line ending as on OS X and Linux. Setting a `.gitattributes` file will ensure that no matter our contributor's global git settings, that the files will be checked out correctly, preventing whitespace only commits. 

See: https://docs.github.com/en/free-pro-team@latest/github/using-git/configuring-git-to-handle-line-endings

Proposed `.gitattributes` file:

```
# Set the default behavior, in case contributors don't have core.autocrlf set.
* text=auto

# Baseline images are binary =and should not be modified
*.png binary
*.jpg binary
*.pdf binary
```

@lwgray is working on the solution on his machine which will allow us to figure out more clearly what is happening.

bbengfort:
This PR fixes #1113 by handling some housekeeping issues. We've changed our master branch to the main branch and created a release issue template (though I don't think it is used automatically by GitHub). We've also checked the new pip resolver and all of our dependencies are compatible. This PR also updates our CI to test Python 3.8 and 3.9 to ensure we're testing on the latest version of Python.

This PR also addresses some warnings that were generated by the tests and fixes #707 

This PR also adds a .gitattributes to fix #1127

Other housekeeping items:

- Apparently Sphinx `add_stylesheet` [is deprecated](https://github.com/sphinx-doc/sphinx/issues/7747) so updated to use `add_css_file` instead. 
- The cross-validation doc images were not building, so fixed the code by adding `shuffle=True` (scikit-learn deprecation)
- Adds a Sphinx build to GitHub ACtions for ensuring that the docs build correctly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,7 @@ jobs:
           conda config --set always_yes yes --set changeps1 no
           conda update -n base conda --yes
           conda env create -f tests/requirements.txt -n yellowbrick python=${{ matrix.python-version }}
+          conda activate yellowbrick
           python -m nltk.downloader popular
 
       - name: Run Unit Tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,3 +104,27 @@ jobs:
           env_vars: OS,PYTHON
           name: codecov-umbrella
           fail_ci_if_error: false
+
+  # Test building docs using Sphinx since this also runs a lot of Yellowbrick code
+  test_docs:
+    runs-on: ubuntu-latest
+    name: Build Docs
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+
+      - name: Install Dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install -r docs/requirements.txt
+
+      - name: Run Sphinx
+        uses: ammaraskar/sphinx-action@master
+        with:
+          docs-folder: "docs/"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,104 +3,103 @@
 
 name: Yellowbrick CI
 
-on: 
-  # Trigger on push to the primary branches only
+on:
+  # Trigger on push to the primary branches and all versioned tags
   push:
     branches:
       - develop
       - main
-      - master
+    tags:
+      - 'v*'
   # Trigger on pull request always (note the trailing colon)
   pull_request:
 
 jobs:
-  test_conda:
-    name: Conda (${{ matrix.python-version }}, ${{ matrix.os }})
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: ["ubuntu-latest"]
-        python-version: ["3.7", "3.8"]
-    steps:
-      - uses: actions/checkout@v2
-      - uses: conda-incubator/setup-miniconda@v2
-        with:
-          auto-update-conda: true
-          python-version: ${{ matrix.python-version }}
-          channels: conda-forge,spyder-ide
-          channel-priority: flexible
-
-      - name: Conda info
-        shell: bash -l {0}
-        run: conda info
-      - name: Add conda to system path
-        shell: bash -l {0}
-        run: |
-          # $CONDA is an environment variable pointing to the root of the miniconda directory
-          echo $CONDA/bin >> $GITHUB_PATH
-      - name: Install dependencies
-        shell: bash -l {0}
-        run: |
-          conda config --set always_yes yes --set changeps1 no
-          conda update -n base conda --yes
-          conda env create -f tests/requirements.txt -n yellowbrick python=${{ matrix.python-version }}
-      - name: Test with pytest
-        shell: bash -l {0}
-        run: |
-          conda activate yellowbrick
-          python -m nltk.downloader popular
-          make test
-
-      - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v1
-        with:
-          env_vars: OS,PYTHON
-          name: codecov-umbrella
-          fail_ci_if_error: false
-
-
+  # Test Vanilla Python (using PyPI for dependencies) as a matrix.
   test_pypi:
+    runs-on: ${{ matrix.os }}
     name: PyPi (${{ matrix.python-version }}, ${{ matrix.os }})
     strategy:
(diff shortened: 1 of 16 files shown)


docs/api/model_selection/cross_validation.rst
```diff
@@ -41,7 +41,7 @@ In the following example, we show how to visualize cross-validated scores for a
     X, y = load_occupancy()
 
     # Create a cross-validation strategy
-    cv = StratifiedKFold(n_splits=12, random_state=42)
+    cv = StratifiedKFold(n_splits=12, shuffle=True, random_state=42)
```
bbengfort, line 44: The docs were raising an exception because a random state was supplied without the shuffle argument set to true. The result was that the images were not shown in the build docs. In fact, they're not shown in our [current docs](https://www.scikit-yb.org/en/latest/api/model_selection/cross_validation.html#classification). Adding `shuffle=True` prevented the exception from being raised.

.github/workflows/ci.yml
```diff
@@ -3,104 +3,128 @@
...
-    strategy:
-      fail-fast: false
-      matrix:
-        python-version: [3.7, 3.8]
-        os: [ubuntu-latest, macos-latest, windows-latest]
-
-    runs-on: ${{ matrix.os }}
-
+  # Test building docs using Sphinx since this also runs a lot of Yellowbrick code
+  test_docs:
```
bbengfort, line 109: Because of the docs build error that prevented the cross-validation visualizer to be shown in the docs I thought it might be good to have a GitHub action for testing the docs build. This test doesn't have to block a PR, but it will give us more visibility into the documentation.

.github/workflows/ci.yml
```diff
@@ -3,104 +3,128 @@
...
       - develop
       - main
-      - master
+    tags:
+      - 'v*'
   # Trigger on pull request always (note the trailing colon)
   pull_request:
 
 jobs:
+  # Test Vanilla Python (using PyPI for dependencies) as a matrix.
```
bbengfort, line 18: Most of the changes below are reordering to have the PyPI tests first then conda tests, as well as some whitespace and comments.

docs/conf.py
```diff
@@ -232,7 +232,7 @@
 
 
 def setup(app):
-    app.add_stylesheet("theme_overrides.css")
+    app.add_css_file("theme_overrides.css")
```
bbengfort, line 235: Apparently, the Sphinx API has changed and this update was required … glad I build the docs!

docs/contributing/developing_visualizers.rst
```diff
@@ -139,12 +139,14 @@ Visual tests are notoriously difficult to create --- how do you test a visualiza
             try:
                 visualizer = MyVisualizer()
                 assert visualizer.fit(X, y) is visualizer, "fit should return self"
-                visualizer.show()
+                visualizer.finalize()
```
bbengfort, line 142: Updates docs per #707

tests/test_classifier/test_prcurve.py
```diff
@@ -435,7 +435,10 @@ def test_quick_method_with_test_set(self):
         )
 
         viz = precision_recall_curve(
-            RandomForestClassifier(random_state=72), X_train, y_train, X_test, y_test
+            RandomForestClassifier(random_state=72),
+            X_train, y_train,
+            X_test, y_test,
+            show=False
```
bbengfort, line 441: Ensure that quick method doesn't call `plt.show()` to prevent warnings per #707

tests/test_cluster/test_elbow.py
```diff
@@ -432,7 +434,7 @@ def test_quick_method_params(self):
         custom_title = "My custom title"
         model = KMeans(3, random_state=13)
         oz = kelbow_visualizer(
-            model, X, sample_weight=np.ones(X.shape[0]), title=custom_title
+            model, X, sample_weight=np.ones(X.shape[0]), title=custom_title, show=False
```
bbengfort, line 437: Ensure that quick method doesn't call `plt.show()` to prevent warnings per #707

tests/test_features/test_radviz.py
```diff
@@ -138,7 +138,8 @@ def test_integrated_radviz_with_pandas(self):
 
         # Test the visualizer
         visualizer = RadViz()
-        visualizer.fit_transform_show(X, y)
+        visualizer.fit_transform(X, y)
+        visualizer.finalize()
```
bbengfort, line 142: Ensure that quick method doesn't call `plt.show()` to prevent warnings per #707

.github/workflows/ci.yml
rebeccabilbro: Great idea &mdash; the docs behavior is always a little mysterious since it has it's own slightly different requirements.txt; this will be a great feature to add to our CI!

rebeccabilbro (approved):
Thank you so much @bbengfort for addressing these issues; it's been a while since we checked in on some of these things (new sklearn deprecations, sphinx, etc), and I'm surprised but happy that there weren't more things that needed to be addressed! 

I really like the idea of having GH Actions build the docs since the differences in local and hosted builds have  often caught me off guard. I'm noticing that GH Actions is flagging two lines in the rst as causing exceptions &mdash; both are inside plot directive plotting routines (line 99 in [learning curve](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/docs/api/model_selection/learning_curve.rst#L99) and line 141 in [umap](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/docs/api/text/umap_vis.rst#L141)). I wonder if this is related to the dependencies? Curious, but agree this should not block merging.

rebeccabilbro (approved):

PR #1189, opened 2021-07-13
PrePredict Estimator

bbengfort:
A contrib estimator to wrap inferences made outside of the Yellowbrick workflow. This estimator can wrap predicted data for use in many visualizers (though not all, depending on if the visualizer needs access to learned attributes) or load data from disk or via a callable.

1. Implemented `PrePredict` a contrib estimator
2. Created tests for the `PrePredict` estimator
3. Wrote documentation for the `PrePredict` estimator

### Sample Code and Plot

Please see the documentation for sample code.

- [ ] I thought we already had some like this in Yellowbrick; @rebeccabilbro mentions there is a blog post, could we perhaps link that blog to the docs?

### CHECKLIST


bbengfort:
Tests passed *whew!

yellowbrick/contrib/prepredict.py
```diff
@@ -0,0 +1,104 @@
...
+        return self
+
+    def predict(self, X):
+        """
+        Predict returns the embedded data but does not perform any checks on the
+        validity of X (e.g. that it has the same shape as the internal data).
+        """
+        return self._load()
+
+    def score(self, X, y=None):
```
lwgray, line 73: Do you want to provide the full array of scores? or allow the score to be provided as an argument?
https://scikit-learn.org/stable/modules/model_evaluation.html

lwgray (commented):
@bbengfort , as always, great job on this PR.  It is definitely needed.  I only had a few items that I was hoping you could respond to.
1. Is there a particular reason you don't include workable code in the docs in order to generate a visualization
2. Some contrib visualizers have quick methods others don't... I think we can merge this without but we have to address this in the future
3. Do you think we need to expand the number of scoring metrics provided in Prepredict?

yellowbrick/contrib/prepredict.py
bbengfort: That's a good idea, we could have the scoring metric as an argument - the real reason this is necessary is for visualizer that embed the score in the visualization; for example the `PredictionError` visualizer puts the `R^2` score into the legend; therefore if we didn't have a score method, some visualizers wouldn't work. Perhaps we could open on a follow on story to expand this if necessary?

bbengfort:
@lwgray thanks for the review! It's been so long since I did this that I don't really remember all the details, but I'll try to answer:

1. I didn't include workable code since we didn't have a dataset that might be well suited to the documentation. In the tests, I used a generated data set, but I don't think that really helps with understanding how this works, since really this is just a tool that gets you to another Visualizer. 
2. Yeh, the `contrib` module is supposed to be for third party or not fully fledged out code; it's kind of an auxiliary so it doesn't necessarily follow the same rules that the other YB visualizers do. 
3.  As I mentioned, the `score` method was really just to get visualizers to work with this data; I could see the scoring method being a useful argument; I recommend that we open up a story to add that feature in the future.

lwgray (approved):
LGTM

Issue #1193, opened 2021-07-30
PredictionError plot forces min/max limits on the graph that are unhelpful if you have lots of tiny errors


ianozsvald:
**Describe the bug**

The default min/max behaviour in the `PredictionError` class is not helpful if you have many tiny errors - adding a +/- 1 boundary when you have only tiny errors gives you a plot like:

![image](https://user-images.githubusercontent.com/273210/127678342-02147621-c871-47a3-8745-8f1c83d5c0ed.png)

If I remove the limits in `regressor/prediction_error.py` in `draw` then I get:

![image](https://user-images.githubusercontent.com/273210/127679224-d9e34775-f2c2-4341-b89a-686dfb917f83.png)

I'm working on the Kaggle Optiver Volatility competition. My errors are in the range +/0 0.002 to 0.0000002.

The two relevant lines of code are here https://github.com/DistrictDataLabs/yellowbrick/blob/develop/yellowbrick/regressor/prediction_error.py#L207 and the comment in the code notes that the +/- 1 limits aren't perhaps the best choice.

I'm not sure when limits like this _would_ be a good idea, maybe the idea is to move a few large errors off of the boundary of the plot? Maybe a solution includes a test for the expected range of the errors so a smaller proportional modifier might be used for smaller errors?

**To Reproduce**

_None provided_ but I can make one if useful.

**Dataset**
Did you use a specific dataset to produce the bug? Where can we access it?

**Expected behavior**

I'd expect to _not_ have the additional +/- 1 min and max boundary limits on the plot.

**Desktop (please complete the following information):**
- Linux Mint
- Python 3.8
- Yellowbrick '1.3.post1'

rebeccabilbro:
Hello @ianozsvald and thanks for reaching out! We should all be so lucky to have such small errors in our regression models 😁  😅 

I really like the idea of allowing the error density influence the plot range! Perhaps we could introduce a user param called something like `zoom_in`, that focuses in on the main density of the error, even if it means ignoring some error points on the outer boundaries. I think the main challenge will be to keep the plot square, since most folks seem to find these plots easier to reason about when the identity and best fit lines are consistently comparable. 

This does seem like an excellent feature to add if anyone is looking for a good candidate for a contribution!

pkaf:
My 2 cents: 

a. Should not plot like this have synchronized x and y limits in a 1:1 aspect ratio axes?
[...]

PR #1208, opened 2022-01-19
BUG: Fixes axes limit for PredictionError plot #1193

pkaf:
This fixes Issue #1193. X and Y axis limits are synced and set to overall max/min values of (X, Y).

Screen Shot 2022-01-20 at 9 16 58 am


bbengfort:
@pkaf thank you again for contributing to Yellowbrick! Since this is a pretty simple change; I think this will work well - the tests are failing because the baseline images are now incorrect; I'm going to regenerate the images and spot-check a few of them. I'll let you know when I do so that you can also view the new baseline images. Note that this means I need write access to your fork - I'll let you know if you need to do anything extra.

bbengfort (commented):
@pkaf I re-ran the baseline images and everything looks ok; one test is still failing: [test_peplot_no_shared_limits](https://github.com/DistrictDataLabs/yellowbrick/blob/develop/tests/test_regressor/test_prediction_error.py#L166) - which asserts that the axes are different based on a user kwarg. This zoom method ensures the aspect ratio is square, should we remove the `shared_limits` kwarg and references to that kwarg in the code? Is it still valuable? Is there a way to change the test so that it will pass?

pkaf:
@bbengfort may be the _shared_limits_ is worth keeping with True as a default value. With following code modification inside the draw function?
```python
self.ax.set_xlim(min(min(y), min(y_pred)), max(max(y), max(y_pred)))
self.ax.set_ylim(self.ax.get_xlim())
 ```
 to 
 ```python
if self.shared_limits == False:
    self.ax.set_xlim(min(y), max(y))
    self.ax.set_ylim(min(y_pred), max(y_pred))
else:
    self.ax.set_xlim(min(min(y), min(y_pred)), max(max(y), max(y_pred)))
    self.ax.set_ylim(self.ax.get_xlim())
 ```
 This will make the current test in-place still relevant. Suggestion?

bbengfort:
@pkaf that works for me - looks like the tests are passing; I'll handle the merge conflicts and we'll get this merged! Thanks again for your patience.

bbengfort (approved):

Issue #1201, opened 2021-10-22
ClassificationReport axes not rendered properly when in a subplot


h-joshi:
**Describe the bug**
ClassificationReport not displaying classification outcomes on Y axis and measures (Precision, Recall, etc.) on X axis
![image](https://user-images.githubusercontent.com/1222657/138379261-dcdb9ad5-7b3d-4f18-847a-197c4db7d3cb.png)

**To Reproduce**

```python
from sklearn.ensemble import RandomForestClassifier
from sklearn.model_selection import train_test_split

import pandas as pd
import numpy as np

import matplotlib.pyplot as plt

from yellowbrick.classifier import ClassificationReport, ConfusionMatrix, ClassPredictionError, DiscriminationThreshold
from yellowbrick.datasets import load_spam


dataset = load_spam()


X_train, X_test, y_train, y_test = train_test_split(dataset[0], dataset[1], test_size=0.33, random_state=42)


def pref_report(estimator, title, X_train, y_train, X_test, y_test):
    plt.rcParams["font.family"] = "Arial"
    
    fig, ax = plt.subplots(nrows=2, ncols=2, figsize=(17, 12))
    fig.suptitle(title)

    visualizer1 = ClassificationReport(estimator(random_state=1), support=True, encoder={0:'Not spam', 1:'Spam'}, ax=ax[0,0])
    visualizer1.fit(X_train, y_train)
    visualizer1.score(X_test, y_test)
    
    visualizer2 = ConfusionMatrix(estimator(random_state=1), encoder={0:'Not spam', 1:'Spam'}, ax=ax[0,1])
    visualizer2.fit(X_train, y_train)
    visualizer2.score(X_test, y_test)
    
    for tick in ax[0,1].get_xticklabels():
        tick.set_rotation(30)
    
    visualizer3 = DiscriminationThreshold(estimator(random_state=1), encoder={0:'Not spam', 1:'Spam'}, ax=ax[1,0])
    visualizer3.fit(X_train, y_train)
    
    visualizer4 = ClassPredictionError(estimator(random_state=1), encoder={0:'Not spam', 1:'Spam'}, ax=ax[1,1])
    visualizer4.fit(X_train, y_train)
    visualizer4.score(X_test, y_test)
    
    for tick in ax[1,1].get_xticklabels():
        tick.set_rotation(30)
    
    fig.legend()


[...]

PR #1210, opened 2022-01-25
BUG: Adds missing X and Y axes labels in ClassificationReport

pkaf:
ClassificaitonReport had no citing of x- and y-ticklabels, which is now added. 

This PR addresses part of the #1201 issue.

1. Added y-axes labels based on class labels
2. Added x-axes labels based on SCORES_KEYS i.e., ['Precision', 'Recall', 'F1', 'Support'] 

Still to do:
- Pass visual/image tests (with some help)

Screen Shot 2022-01-25 at 4 21 42 pm


pkaf:
I can push new images into baseimages if I know where are self.binary coming from in https://github.com/pkaf/yellowbrick/blob/develop/tests/test_classifier/test_classification_report.py ?

bbengfort:
@pkaf `self.binary` is a binary classification dataset that's used as a fixture for testing - it's a bit of pytest magic to be sure since it's not imported, but it's coming from `conftest.py`.

The simplest way to regenerate the baseline images is as follows:

1. Remove the baseline images:

```
$ python -m tests.images --clear tests/test_classifier/test_classification_report.py
```

2. Rerun the tests (they will all fail since there are no baseline images):

```
$ pytest tests/test_classifier/test_classification_report.py
```

3. Sync the generated images from the tests with the baseline images:

```
$ python -m tests.images tests/test_classifier/test_classification_report.py
```

Hopefully, when you run the tests again they will pass! Take care to ensure you're only regenerating the images for the classification report tests; if you don't supply the path to the test to our helper module you could end up regenerating all the images. Different operating systems generate images differently, so it might break other tests if that happened.

yellowbrick/classifier/classification_report.py
```diff
@@ -188,6 +190,13 @@ def score(self, X, y):
         y_pred = self.predict(X)
         scores = precision_recall_fscore_support(y, y_pred)
 
+        # Decode the target with the label encoder and get human readable labels
```
lwgray, line 193: @pkaf Are lines 193 to 199 necessary?  Can you explain your logic here?  

@bbengfort Did I miss something here?

lwgray (changes requested):
@pkaf This is wonderful, Thank you so very much for submitting this PR.  We are sorry that it has taken this long to get you a review.  I need to request some changes and provide some clarification.
1. Add fontsize definition to docstrings of Visualizer and quick method
2. Add fontsize as a parameter to the quickmethod
3. Address my comments on Lines 193-199

lwgray (approved):

Issue #1086, opened 2020-07-14
PredictionError hard codes score legend to R2 even when other scoring functions are used


pablo14:
Hi! Thanks for this package! It's wonderful.

The problem is I see another metric in the legend, the one I use to optimize, and the legend still shows the R2, as you can see:

```
from sklearn.model_selection import GridSearchCV
from sklearn.ensemble import RandomForestRegressor 


model_rf = RandomForestRegressor()

params = {
    'n_estimators' : [100, 200, 500, 1000],
    'max_features': [2,3]
}

grid_rf = GridSearchCV(estimator = model_rf,
                        param_grid = params,
                        scoring = 'neg_mean_absolute_error',  # 
                        cv = 5, 
                        verbose = 1,
                        n_jobs = -1)

grid_rf.fit(x_train, y_train)

## Now the plot:
from yellowbrick.datasets import load_concrete
from yellowbrick.regressor import PredictionError

# ojo: poner best_estimator_
visualizer = PredictionError(grid_rf)

# copy-paste de la doc oficial: https://www.scikit-yb.org/en/latest/quickstart.html
visualizer.fit(x_train, y_train)  # Fit the training data to the visualizer
visualizer.score(x_test, y_test)  # Evaluate the model on the test data
visualizer.show()                 # Finalize and render the figure

```
Then I see:
![imagen](https://user-images.githubusercontent.com/9369685/87471116-329c0800-c5f4-11ea-95cc-9d8132e7ca3e.png)

As you can see, I see the neg mean average error with the R2 legend. 

However, if I pass the best fit:
`visualizer = PredictionError(grid_rf.best_estimator_)`

Then:

![imagen](https://user-images.githubusercontent.com/9369685/87471351-97eff900-c5f4-11ea-91d6-da19b58b9438.png)

The true R2 appear.

Thanks!

bbengfort:
@pablo14 thank you so much for using Yellowbrick; I'm glad to hear you're enjoying it! 

You're right - this is an unfortunate bug, since all `PredictionError` is doing is calling the `score()` method of the regressor. In most cases, this score is an `R2`, but you've clearly shown the case for using another scoring method! I really only see two options of fixing this:

1. Generically renaming the legend to be `score` - but I'm not a fan of this since it doesn't give a lot of information.
2. Figuring out some way to identify what the `scoring` function is on the estimator. 

If anyone has any suggestions on how to do 2 - let me know and I'd be happy to create a fix!

[...]

PR #1212, opened 2022-01-25
BUG: Corrects legend issues other than R2 in PredictionError

pkaf:
This PR addresses the issue raised at #1086

Unfortunately, generic .score() method in scikit-learn does not return the name of the scoring metrics used. The fix I have patched will return correct label if and when estimator has attribute providing the name of the scoring metric otherwise it will fall back to the default value i.e. R2


lwgray (approved):
These changes allow for the scoring name to be accurately displayed and default to R2 when scoring name is not provided.  I have tested this with a variety of metrics from sklearn.metrics.SCORERS.keys().  We may need tests to cover name changes.

PR #1221, opened 2022-02-25
Fixed #1209 `is_fitted` parameter not setting to given value

### LINKED ISSUE #1209: Argument to PredictionError is overwritten
**Describe the bug**
Argument `is_fitted` is overwritten when Instantiating a `PredictionError` or  `ResidualsPlot`.

**To Reproduce**

```python
from yellowbrick.regressor import PredictionError
from sklearn.linear_models import Lasso

visualizer = PredictionError(Lasso(), "test title", is_fitted=True)
print(visualizer.is_fitted)
```

**Expected behavior**
When `is_fitted` is passed in, I expect it to not be overwritten in the __init__ calls

 - OS: Ubuntu 20.04
 - Python Version 3.8.10
 - Yellowbrick Version :  1.3.post1

lwgray:
This PR fixes #1209 so that the is_fitted parameter is no longer being overwritten by the default setting.

1. I updated the super for both ResidualsPlot and PredictionError to include is_fitted as a parameter
2. I deleted the self assignment of `is_fitted` in both visualizers

### DIFF
--- a/yellowbrick/regressor/prediction_error.py
+++ b/yellowbrick/regressor/prediction_error.py
@@ -121,11 +121,13 @@ class PredictionError(RegressionScoreVisualizer):
         is_fitted="auto",
         **kwargs
     ):
-        # Whether or not to check if the model is already fitted
-        self.is_fitted = is_fitted
 
         # Initialize the visualizer
-        super(PredictionError, self).__init__(estimator, ax=ax, **kwargs)
+        super(PredictionError, self).__init__(
+            estimator,
+            is_fitted=is_fitted,
+            ax=ax,
+            **kwargs)
 
         # Visual arguments
         self.colors = {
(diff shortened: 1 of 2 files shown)


yellowbrick/regressor/prediction_error.py
```diff
@@ -121,11 +121,13 @@ def __init__(
...
-        # Whether or not to check if the model is already fitted
-        self.is_fitted = is_fitted
 
         # Initialize the visualizer
-        super(PredictionError, self).__init__(estimator, ax=ax, **kwargs)
+        super(PredictionError, self).__init__(
+            estimator,
+            is_fitted=is_fitted,
+            ax=ax,
+            **kwargs)
```
bbengfort, line 130: *whew, glad that was such an easy fix!

bbengfort (approved):
@lwgray looks good thank you for looking into this! I think the only thing extra we could do is add a test to make sure that the bug is solved and that we don't regress in the future; but I'll leave it up to you if you'd like to do that or not. Glad that it's sorted though, I was very confused by this when I first saw it!

PR #1234, opened 2022-03-27
Full example of Regression Model Visuals

uricod:
I was not able to find a example of using yellowbrick visuals for a regression models so I build one with custom dataset.


lwgray:
@uricod Thank you for using Yellowbrick and for submitting this example.  We do in fact have a notebook for regression https://github.com/DistrictDataLabs/yellowbrick/blob/1979a511d8d0a15339d0c3845f88660fad5945e0/examples/regression_visualizers.ipynb

Let me reach out to the team and see if we should include yours.

uricod:
@lwgray  - I looked at that example. It only has a portion of what you can do. Compare that to mine and I you will see what I mean. The classification examples are much better.

bbengfort:
@uricod thank you for adding your example regression notebook. Like @lwgray mentioned, if you'd like to update the `regression_visualizers.ipynb` notebook we'd be happy to take a look at those changes. In the meantime, we'll certainly merge your example notebook - but we'll rename the folder from regressionVisualizers to your GitHub username for proper attribution of the example. 

Thank you for contributing to Yellowbrick!

bbengfort (approved):

PR #1243, opened 2022-05-02
[WIP] Add WordCorrelationPlot visualizer

### LINKED ISSUE #148: WordCorrelationPlot
Implement a `WordCorrelationPlot` that enables users to visualize the correlations between a word and other words in the corpus, similar to what is available in [ggplot2](http://stackoverflow.com/questions/19549280/plot-highly-correlated-words-against-a-specific-word-of-interest).

pdeziel:
This PR will fix #148 which requested a `WordCorrelationPlot` visualizer to allow the user to visualize the correlations between words in the corpus. I came across this issue in the backlog and I can see myself using it so I'd be happy to work on it.

### Sample Code and Plot

TODO

### TODOs and questions

Still to do:

- [x] Implement the visualizer
- [x] Create the tests
- [x] Add the documentation

### CHECKLIST

- [ ] _Is the commit message formatted correctly?_


lwgray:
Thanks for opening a PR.  Please move forward with the development of this feature.  All the best

lwgray:
@pdeziel **bump**  Is there anything we can do to help with your PR?

pdeziel:
> @pdeziel **bump** Is there anything we can do to help with your PR?

Hi @lwgray, I think I have most of the functionality completed, although I still need to clean up the code comments, add documentation, etc.

There is one visual bug I was working on tracking down which I might need help with; the color mapping on the visualization doesn't seem correct (e.g., 0.29 in the below chart should be "bluer" than 0.16). This might have to do with `pcolormesh` expecting a value in the range [0,1] but the correlation value range is from [-1,1]. I would also ideally like to have the red and blue values swapped in the default color map (so something like `BuYlRd`) but that color map doesn't currently exist in YB.

correlation

lwgray:
@pdeziel Awesome work... Maybe you can internal convert it to be between 0 and 1 while still display -1 to 1.  Maybe @bbengfort or @rebeccabilbro can chime in on this.  I guess in the meantime you can work on the docs and tests.

yellowbrick/text/correlation.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def _construct_term_index(self):
+        """
+        Constructs a dictionary mapping terms to indices in the sparse doc-term matrix.
+        """
+        self.term_index_ = {}
+        for i, term in enumerate(self.vocab_):
+            if term in self.words:
+                if term in self.term_index_:
+                    raise ModelError("Duplicate term in vocabulary: {}".format(term))
```
lwgray, line 113: Awesome error handling here, Could you write a test to raise it?    Also, do you think that is a different type of error than a ModelError?  Just wondering ❓

bbengfort:
@pdeziel @lwgray I'm wondering if the colors bug is a mismatch of the labels and the colors; e.g. a matrix got inverted somewhere incorrectly. If you look at (Tatsumi Kimishima, Nintendo) = .29, c=yellow; and (woman, play) = 0.04 c=blue, these cells are inverse of each other on the matrix but seem to have the other cell's color. 

I'd look at the `wc_display` flip on [L174](https://github.com/DistrictDataLabs/yellowbrick/pull/1243/files#diff-55dbf262ca044c266963626badf213741bd847a183f99b1e45b857b555cd34d2R174) as a start, then check the top left and top right positions in `X` and `Y`.

yellowbrick/text/correlation.py
bbengfort: I agree @lwgray this might be a good place to introduce a subclass of `ModelError` specifically for text.

pdeziel: @lwgray @bbengfort I'm not sure I can trigger this error from the API, since sklearn _should_ not be producing a vocab with repeated words, so I would have to make the unit test more specific to `_construct_term_index`. This leads me to the question, should we assume that sklearn will behave this way? Instead of catching the error, we could convert the vocab to a `set` to ensure that the values are unique. This would protect us a bit but we might also want to know if the sklearn API changes. What do you guys think?

lwgray: @bbengfort Can you chime in?

bbengfort: The sklearn API does change quite a bit, so being defensive about it is usually a good idea. Just out of curiosity, the vocab comes from `vec.get_feature_names_out()` right? Why not use `vec. vocabulary_.keys()`? This variable is a dictionary, so would guarantee uniqueness and would prevent the need for an intermediate `set`. 

If triggering this error is too difficult for a test we should put a comment above the `raise` to indicate what we thinking is going on (e.g. unexpected duplicate vocab possibly due to sklearn API change).

pdeziel:
@lwgray @bbengfort I've fixed the color mapping issue and added the documentation, so I think is ready for review again. There seems to be one CI error, but it's in a different module so I haven't figured out how it's related to my changes yet.

One thing I think would be useful to add in the future is a parameter where the user would only have to supply one word, and the visualizer finds the top correlated words in the corpus.

lwgray (approved):
Great job on the docs, they are high quality.

PR #1260, opened 2022-06-20
Fixes minor typo in rank docs

davidgilbertson:
This is my first PR to Yellowbrick.

I've made this change through the GitHub UI, is that OK? Or do I need to actually checkout the repo and install everything and build the docs?

PR #1263, opened 2022-06-21
Fixes minor typo in rank docs

davidgilbertson:
Now targeting develop, same change as https://github.com/DistrictDataLabs/yellowbrick/pull/1260


bbengfort (approved):
Thanks for opening the PR again - and you're right; it's unfortunate that the link points to main. Read the Docs automatically adds that link to the file from the build; our "latest" docs point to our previous versioned release, which is in the `main` branch. 

I did a little Googling and it may be possible to change this link to point to develop, but would require a bit of configuration and testing on our side. @lwgray what do you think?

https://stackoverflow.com/questions/66227417/read-the-docs-theme-change-the-edit-on-github-link-using-sphinx

Don't worry about the commits - we're going to squash and merge into `develop` so it should all be sorted. Thanks again for using Yellowbrick!

lwgray:
@bbengfort I will look into this!

davidgilbertson:
Good stuff. By the way, you should add Yellowbrick to https://github.com/vinta/awesome-python#data-visualization. I'm just starting out and that was my first stop for discovering neat visualisation tools (I eventually found Yellowbrick via the Matplotlib docs).

bbengfort:
@davidgilbertson that's a good suggestion - would you be willing to open a PR for us there?

davidgilbertson:
Done. https://github.com/vinta/awesome-python/pull/2083

bbengfort:
Much appreciated!

PR #1293, opened 2023-01-13
Fix typo in kneed.py

w4bo:

1. fix a typo


bbengfort (approved):
Thank you for catching this typo and fixing it!

PR #1294, opened 2023-01-19
`SilhouetteVisualizer` add support for more estimators

### LINKED ISSUE #1182: Add more estimators to Silhouette Visualizer
At the moment Silhouette Visualizer can handle KMeans and MiniBatchKMeans. But as far as I know, the construction of silhouette plot isn't tied to centroid clustering in any way, but simply counts the score according to the specified metric. The only problem I can see here is that some algorithms, like OPTICS and DBSCAN, have a separate label for outliers.

Maybe I could help with this, however I'm new on GitHub, so I might need some help :)

stergion:
This PR closes #1182 .

1. It adds support for clustering estimators that do not implement a `predict()` method,
2. for estimators without attribute `n_clusters` infer clusters from labels,
3. uses the estimator's `metric` or `affinity` as the silhouette metric.

I decided not to implement special handling for estimators that produce outlier values, eg DBSCAN, 
as sklearn doesn't do neither in their examples.

I wasn't sure weather to use the estimator's metric attribute for the silhouette metric, 
or add a parameter `metric` in `SilhouetteVisualizer` constructor. I chose the first option because it didn't 
alter the class signature and seemed like the safer option. Although, I do believe the second option to be the better one.


bbengfort:
@stergion thank you so much for your interest in Yellowbrick and for opening this PR; we really appreciate all contributions to Yellowbrick. We'll find a reviewer for this PR as soon as possible so that we can include it in our next release.

yellowbrick/cluster/silhouette.py
```diff
@@ -135,18 +135,30 @@ def fit(self, X, y=None, **kwargs):
...
         # is a little different and I'm not sure how it's used.
 
-        if not check_fitted(self.estimator, is_fitted_by=self.is_fitted):
-            # Fit the wrapped estimator
-            self.estimator.fit(X, y, **kwargs)
+        # if estimator is fitted AND has attribute
+        if check_fitted(self.estimator, is_fitted_by=self.is_fitted) and hasattr(self.estimator, "predict"):
+            labels = self.estimator.predict(X)
+        else:  # if estimator is NOT fitted, OR estimator does NOT implement predict()
+            labels = self.estimator.fit_predict(X, y, **kwargs)
```
lwgray: Great 👍  way to cover fit_predict here.

yellowbrick/cluster/silhouette.py
```diff
@@ -135,18 +135,30 @@ def fit(self, X, y=None, **kwargs):
...
 
         # Get the properties of the dataset
         self.n_samples_ = X.shape[0]
-        self.n_clusters_ = self.estimator.n_clusters
+
+        if hasattr(self.estimator, "n_clusters"):
+            self.n_clusters_ = self.estimator.n_clusters
+        else:
+            self.n_clusters_ = len(set(labels)) - (1 if -1 in labels else 0)
```
lwgray: I wonder if there is a more expressive way to write this (specifically the 1 if -1 in labels else 0)

lwgray (commented):
This is a great PR.  I like how you simply added support for more estimators.  The only thing left is to write some tests.  Let me know if you need help with that.  After that I will approve.

yellowbrick/cluster/silhouette.py
stergion:
```suggestion
            unique_labels = set(labels)
            n_noise_clusters = 1 if -1 in unique_labels else 0
            self.n_clusters_ = len(unique_labels) - n_noise_clusters 
```

Is this any better?

stergion:
@lwgray Thanks for the comments and the review. Sorry, I took so long to reply.

If you could help me with the tests, it would be great, since I don't have any experience.

lwgray:
I went back and reviewed this again and realized that all but two clustering algorithm work immediately with this fix.  I am unsure why I get the ValueError with SpectralClustering.  FeatureAgglomeration gives back an AttributeError and it is because it doesn't have a fit_predict method.  See table below:

|                                                                     | Works | Error                                                                                                                                                                                                                                                                                                                                                                                                                                      |
| ------------------------------------------------------------------- | ----- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| <class 'sklearn.cluster._kmeans.KMeans'>                            | Yes   |                                                                                                                                                                                                                                                                                                                                                                                                                                            |
| <class 'sklearn.cluster._kmeans.MiniBatchKMeans'>                   | Yes   |                                                                                                                                                                                                                                                                                                                                                                                                                                            |
| <class 'sklearn.cluster._affinity_propagation.AffinityPropagation'> | Yes   |                                                                                                                                                                                                                                                                                                                                                                                                                                            |
| <class 'sklearn.cluster._mean_shift.MeanShift'>                     | Yes   |                                                                                                                                                                                                                                                                                                                                                                                                                                            |
| <class 'sklearn.cluster._spectral.SpectralClustering'>              | NO    | ValueError: Unknown metric rbf. Valid metrics are ['euclidean', 'l2', 'l1', 'manhattan', 'cityblock', 'braycurtis', 'canberra', 'chebyshev', 'correlation', 'cosine', 'dice', 'hamming', 'jaccard', 'kulsinski', 'mahalanobis', 'matching', 'minkowski', 'rogerstanimoto', 'russellrao', 'seuclidean', 'sokalmichener', 'sokalsneath', 'sqeuclidean', 'yule', 'wminkowski', 'nan_euclidean', 'haversine'], or 'precomputed', or a callable |
| <class 'sklearn.cluster._dbscan.DBSCAN'>                            | Yes   |                                                                                                                                                                                                                                                                                                                                                                                                                                            |
| <class 'sklearn.cluster._optics.OPTICS'>                            | Yes   |                                                                                                                                                                                                                                                                                                                                                                                                                                            |
| <class 'sklearn.cluster._agglomerative.AgglomerativeClustering'>    | Yes   |                                                                                                                                                                                                                                                                                                                                                                                                                                            |
| <class 'sklearn.cluster._birch.Birch'>                              | Yes   |                                                                                                                                                                                                                                                                                                                                                                                                                                            |
| <class 'sklearn.cluster._agglomerative.FeatureAgglomeration'>       | NO    | AttributeError: - Does not have a fit_predict method                                                                                                                                                                                                                                                                                                                                                                                       |

lwgray:
A couple of questions... 

1. What do we do about tests?  Do we write individual test for each clustering algorithm?
2. After tests have been dealt with, Can we write up issues for the two remaining clustering algorithms that aren't fixed by this PR?  I suggest we merge this PR after test added.

yellowbrick/cluster/silhouette.py
```diff
@@ -135,18 +135,30 @@ def fit(self, X, y=None, **kwargs):
...
+
+        if hasattr(self.estimator, "metric"):
+            metric = self.estimator.metric
+        elif hasattr(self.estimator, "affinity"):
+            metric = self.estimator.affinity
```
bbengfort: @lwgray this is where the error is occurring for SpectralClustering - `SpectralClustering` does have an attribute `affinity` which is used to compute the adjacency matrix between instances. For spectral clustering the attribute `affinity` is not a distance metric and defaults to "rbf" -- which is why that test is failing. 

@stergion what model prompted you to add this metric selector?

bbengfort (commented):
@lwgray I think it is simple enough to add a test for all the clusters using pytest parameterize -- we do that for a lot of tests, and I agree it should be part of this PR. 

We do need to fix SpectralClustering since the bug was introduced in this PR. FeatureAgglomoration is a transformer not a model -- so I think we can omit that from the tests.

yellowbrick/cluster/silhouette.py
stergion: it was for the AffinityPropagation, AgglomerativeClustering, FeatureAgglomeration.

Since sklearn version 1.2, `affinity` is deprecated in AgglomerativeClustering and FeatureAgglomeration 
and `metric` is used instead, like the other clustering algorithms.

AffinityPropagation was not updated, it still uses `affinity`. Although, AffinityPropagation uses the negative 
squared euclidean distance between points, when `affinity='euclidean'`

lwgray:
@stergion @bbengfort  Can you review my changes?

bbengfort (approved):
@lwgray thank you for making those changes! I made a few changes of my own to try to help with the CI tests and linting. Would you and @stergion please review those changes?

bbengfort:
@lwgray RE: the conda tests; in a separate PR we're going to have to update our Python versions for testing. The Conda 3.8 and 3.9 test failures can be ignored.

stergion (commented):
Looks good to me. Your changes also made the code cleaner and easier to understand.

bbengfort:
@stergion thank you so much for your contribution to Yellowbrick!

PR #1301, opened 2023-03-07
feature_datasaurus produce image without color usage warning message #1299

### LINKED ISSUE #1299: Matplotlib warning about color usage in Datasaurus
**Describe the bug**
the datasaurus function used to demonstrate Anscombe's quartet is isssuiing the following warning

WARNING:matplotlib.axes._axes:*c* argument looks like a single numeric RGB or RGBA sequence, which should be avoided as value-mapping will have precedence in case its length matches with *x* & *y*.  Please use the *color* keyword-argument or provide a 2D array with a single row if you intend to specify the same RGB or RGBA value for all points.

**To Reproduce**

```python
from yellowbrick import datasaurus
datasaurus()
```

**Dataset**

**Expected behavior**
Produce the image without warnings

 - OS: [e.g. macOS] Google Colab
 - Python Version [e.g. 2.7, 3.6, miniconda] 3.8.10
 - Yellowbrick Version [e.g. 0.7] 1.5

clmangham:
This PR fixes #1299 which reported a warning about color usage in Datasaurus

1. Replaced the parameter 'c' with 'color' in the matplotlib scatter function to produce image without warning

Sample Code

OLD: `ax.scatter(x, y, c=color)`
NEW: `ax.scatter(x, y, color=color)`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,4 +27,4 @@ cycler>=0.10.0
 # twine>=3.3
 
 ## Pre-Commit Requirements (uncomment to use pre-commit hooks)
-pre-commit>=2.20.0
+# pre-commit>=2.20.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,4 +27,4 @@ cycler>=0.10.0
 # twine>=3.3
 
 ## Pre-Commit Requirements (uncomment to use pre-commit hooks)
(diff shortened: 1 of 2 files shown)


lwgray (approved):
This is a great - simple fix.  Thanks for contributing @clmangham

Issue #1330, opened 2026-01-30
Deprecated distutils - incompatibility with Python 3.12


tohuuuuu:
**Describe the bug**
Trying to follow the [Quick Start](https://www.scikit-yb.org/en/latest/quickstart.html) example with Python 3.12, i ran into a ModuleNotFound error: No module named 'distutils'.

**To Reproduce**

Just run the first example from the "[Walktrough](https://www.scikit-yb.org/en/latest/quickstart.html#walkthrough)" section of the Quick Start guide. See Traceback below.

**Traceback**

```python
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], line 2
      1 import pandas as pd
----> 2 from yellowbrick.datasets import load_bikeshare
      4 X, y = load_bikeshare()
      5 print(X.head())

File ~/develop/git/yb-playground/.venv/lib/python3.12/site-packages/yellowbrick/__init__.py:31
     28 from .version import get_version, __version_info__
     30 # Import the style management functions
---> 31 from .style.rcmod import reset_defaults, reset_orig
     32 from .style.rcmod import set_aesthetic, set_style, set_palette
     33 from .style.palettes import color_palette, set_color_codes

File ~/develop/git/yb-playground/.venv/lib/python3.12/site-packages/yellowbrick/style/__init__.py:20
     12 """
     13 Manage the style and aesthetic of the yellowbrick library.
     14 """
     16 ##########################################################################
     17 ## Imports
     18 ##########################################################################
---> 20 from .colors import *
     21 from .palettes import *
     22 from .rcmod import *

File ~/develop/git/yb-playground/.venv/lib/python3.12/site-packages/yellowbrick/style/colors.py:33
     29 from yellowbrick.exceptions import YellowbrickValueError
     32 # Check to see if matplotlib is at least sorta up to date
---> 33 ``from distutils.version import LooseVersion``
     35 mpl_ge_150 = LooseVersion(mpl.__version__) >= "1.5.0"
     38 ##########################################################################
     39 ## Color Utilities
     40 ##########################################################################

ModuleNotFoundError: No module named 'distutils'
```

**Desktop (please complete the following information):**
 - OS: Linux
 - Python Version 3.12
 - Yellowbrick Version 1.5

Should be solved with #1323 , I assume.

PR #1328, opened 2025-06-26
Update Dependencies for Latest Version Tests

bbengfort:
This PR is an update to all of the test dependencies to ensure Yellowbrick works with the latest versions of it's primary dependencies.

This PR fixes #1327

I have made the following changes:

1. Updated the tests/requirements.txt
2. Updated all of the baseline images with the new versions
3. Fixed import errors and other issues in the code
4. Updated tests as necessary

### CHECKLIST

- [x] _Is the commit message formatted correctly?_
- [ ] _Have you noted the new functionality/bugfix in the release notes of the next release?_

- [ ] _Included a sample plot to visually illustrate your changes?_
- [ ] _Do all of your functions and methods have docstrings?_
- [x] _Have you added/updated unit tests where appropriate?_
- [x] _Have you updated the baseline images if necessary?_
- [x] _Have you run the unit tests using `pytest`?_
- [x] _Is your code style correct (are you using PEP8, pyflakes)?_
- [ ] _Have you documented your new feature/functionality in the docs?_

- [ ] _Have you built the docs using `make html` (must be run from `docs/`)?_

PR #1334, opened 2026-03-07
Fix issue #1330: Remove distutils dependency for Python 3.12+ compatibility

edschofield:
distutils was removed from the stdlib in Python 3.12. Yellowbrick only used distutils to check whether matplotlib >= 1.5.0, a version that is now ancient history, even though Yellowbrick depends on Matplotlib >= 2.0.2.

This patch removes the version checks and obsolete workarounds for older versions of matplotlib.

This is inspired by [this commit](https://github.com/DistrictDataLabs/yellowbrick/pull/1328/changes/ad375662d81d6e5afb325bee7e6e5ebecff41fee#diff-31175572045416fb3bb300935d3e7ece40ff519d7faeae3846d4a5b4e11131e5) but cleaner, omitting many unrelated modifications to test images.


williamgilpin:
@edschofield Thank you! I was hitting this same bug in my workflow, and installing from this PR fixed it for me. I very much appreciate it.

PR #1337, opened 2026-05-18
Fix type checking functions broken by sklearn 1.8

### LINKED ISSUE #1316: Conda install fails
It appears that districtdatalabs  is unreachable when appempting to install via conda

conda install -c districtdatalabs yellowbrick 

requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://conda.anaconda.org/districtdatalabs/linux-64/current_repodata.json

conda config --show channels
channels:
  - districtdatalabs
  - defaults
  - conda-forge
  - bioconda

krishnavamshi01143-lgtm:
While trying to use Yellowbrick with scikit-learn 1.8, I found that 
the library completely breaks — every classifier, regressor, and 
clusterer gets rejected as the wrong type.

The root cause is that sklearn 1.8 removed the `_estimator_type` 
attribute from its mixin classes. Yellowbrick's `is_classifier()`, 
`is_regressor()`, and `is_clusterer()` in `utils/types.py` were 
checking for exactly that attribute, so they now silently return 
False for every estimator.

The fix is straightforward — instead of reimplementing this check 
ourselves, just delegate to sklearn's own `is_classifier`, 
`is_regressor`, and `is_clusterer` from `sklearn.base`. They handle 
the detection correctly regardless of sklearn version.

Tested on Python 3.14 + sklearn 1.8.0 with DecisionTreeClassifier 
and LogisticRegression.

Fixes #1316

### DIFF
--- a/yellowbrick/utils/types.py
+++ b/yellowbrick/utils/types.py
@@ -20,7 +20,12 @@ Detection utilities for Scikit-Learn and Numpy types for flexibility
 import inspect
 import numpy as np
 
-from sklearn.base import BaseEstimator
+from sklearn.base import (
+    BaseEstimator,
+    is_classifier as _sklearn_is_classifier,
+    is_regressor as _sklearn_is_regressor,
+    is_clusterer as _sklearn_is_clusterer,
+)
 from yellowbrick.contrib.wrapper import ContribEstimator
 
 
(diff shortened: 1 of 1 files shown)


krishnavamshi01143-lgtm:
Hi I've fixed the nested is_clusterer function bug causing CI failures across all environments Would appreciate a review when you get a chance Thank you!

